PR #441, opened 2022-10-13
fix: Improve lock/unlock states

acezard:
- Delete data on logout
- Show splashcreen on wake up
- Show splashscreen on boot
- Show lockscreen on boot

### DIFF
--- a/src/screens/home/components/HomeView.js
+++ b/src/screens/home/components/HomeView.js
@@ -10,12 +10,17 @@ import { CozyProxyWebView } from '/components/webviews/CozyProxyWebView'
 import { consumeRouteParameter } from '/libs/functions/routeHelpers'
 import { resetUIState } from '/libs/intents/setFlagshipUI'
 import { useSession } from '/hooks/useSession'
+import { routes } from '/constants/routes'
+import { navigate } from '/libs/RootNavigation'
+import { getData, StorageKeys } from '/libs/localStore/storage'
 
 const unzoomHomeView = webviewRef => {
   webviewRef?.injectJavaScript(
     'window.dispatchEvent(new Event("closeApp"));true;'
   )
 }
+
+let hasRenderedOnce = false
 
 const HomeView = ({ route, navigation, setLauncherContext, setBarStyle }) => {
   const client = useClient()
@@ -108,6 +113,16 @@ const HomeView = ({ route, navigation, setLauncherContext, setBarStyle }) => {
       getHomeUri()
     }
   }, [uri, client, route, navigation, session])
+
+  useEffect(() => {
+    const lockRedirect = async () => {
+      const shouldRedirect = await getData(StorageKeys.AutoLockEnabled)
+      shouldRedirect && navigate(routes.lock)
+      hasRenderedOnce = true
+    }
+
+    !hasRenderedOnce && void lockRedirect()
+  }, [])
 
   const handleTrackWebviewInnerUri = webviewInneruri => {
     if (webviewInneruri !== trackedWebviewInnerUri) {
--- a/src/screens/lock/hooks/useLockScreen.ts
+++ b/src/screens/lock/hooks/useLockScreen.ts
@@ -11,6 +11,7 @@ import { navigate } from '/libs/RootNavigation'
 import { resetUIState, StatusBarStyle } from '/libs/intents/setFlagshipUI'
 import { routes } from '/constants/routes'
 import {
+  ensureLockScreenUi,
   getBiometryType,
   logout,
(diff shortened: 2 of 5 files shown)


src/screens/lock/hooks/useLockScreen.ts
```diff
@@ -51,6 +52,8 @@ export const useLockScreenProps = (route?: RouteProp): LockViewProps => {
     void getData(StorageKeys.BiometryActivated).then(async activated => {
       if (!activated) return
 
+      await hideSplashScreen()
```
Ldoppea, line 55: My understanding is that SplashScreen would never be hiden whe no biometry is activated (i.e. only password or pin code). Did I miss something?

src/screens/home/components/HomeView.js
```diff
@@ -109,6 +112,15 @@ const HomeView = ({ route, navigation, setLauncherContext, setBarStyle }) => {
...
   }, [uri, client, route, navigation, session])
 
+  useEffect(() => {
+    const lockRedirect = async () => {
+      const shouldRedirect = await getData(StorageKeys.AutoLockEnabled)
+      shouldRedirect && navigate(routes.lock)
+    }
+
+    void lockRedirect()
+  }, [])
```
Ldoppea, line 122: With this effect, we will have to be sure that the home cannot be re-mount during app's lifecycle. Hope we won't encounter this kind of side effect in the future :sweat_smile:

src/screens/lock/hooks/useLockScreen.ts
acezard: You're right, but it would be still hidden by cozy-home hideSplashScreen intent (I had an issue where fingerprint view showed up before the LockScreen components). I will nonetheless try the scenario.

src/screens/home/components/HomeView.js
acezard: This would be a terrible bug indeed, but I can try to find an out-of React storage (something in RAM, not asyncStorage) to use as a tag to failproof this so it doesn't ever fire twice

src/screens/lock/hooks/useLockScreen.ts
Ldoppea: Isn't the cozy-home intent called only on app's bootstrap? What about the scenario when the app's is resumed from background (the 1st time when `routes.lock` is not set yet)

src/screens/home/components/HomeView.js
Ldoppea: So you didn't find any way to call the home/lock simultaneous navigation from the bootstrap?

src/screens/lock/hooks/useLockScreen.ts
acezard: Yea I think I should add a hideSplashScreen() on first render just in case (having multiple hide/hiding if it's already hidden is not an issue)

src/screens/home/components/HomeView.js
acezard: That's what it's doing. As soon as Home get to this useEffect, it redirects to lockScreen. Some time later, the intent will hideSplashScreen while LockScreen is already rendering/rendered

Ldoppea: What I meant was "from the useAppBootstrap.js file". This would ensure that home's rerender cannot have side effect on this.

acezard: For now the only way to have simultaneous renders that I found working is doing it this way; navigating into the nested route then navigating to the modal ASAP => this is a trick but it will actually render the nested route and the dialog route at the same time

src/screens/home/components/HomeView.js
```diff
@@ -10,13 +10,18 @@ import { CozyProxyWebView } from '/components/webviews/CozyProxyWebView'
...
+import { navigate } from '/libs/RootNavigation'
+import { getData, StorageKeys } from '/libs/localStore/storage'
 
 const unzoomHomeView = webviewRef => {
   webviewRef?.injectJavaScript(
     'window.dispatchEvent(new Event("closeApp"));true;'
   )
 }
 
+let hasRenderedOnce = false
```
Ldoppea, line 23: This is used to mitigate the side effect of not being able to initiate the Home route and the Lock route simultaneously. It works for now but it may makes harder to understand/maintain the code later. Can we keep track of this and think about how to improve this later?

Ldoppea (approved):

src/screens/home/components/HomeView.js
acezard: Sure, it's not ideal

PR #442, opened 2022-10-14
fix: add style improvements

acezard:
Font size, font family, keyboard behaviour

### DIFF
--- a/src/screens/lock/LockScreen.tsx
+++ b/src/screens/lock/LockScreen.tsx
@@ -1,5 +1,11 @@
 import React from 'react'
 import RnMaskInput from 'react-native-mask-input'
+import {
+  Keyboard,
+  KeyboardAvoidingView,
+  Platform,
+  TouchableWithoutFeedback
+} from 'react-native'
 
 import { Button } from '/ui/Button'
 import { Container } from '/ui/Container'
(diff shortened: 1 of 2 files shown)


src/screens/lock/LockScreen.tsx
```diff
@@ -131,6 +134,15 @@ export const LockScreen = (props: LockScreenProps): JSX.Element => (
   <>
     <FlagshipBars />
 
-    <LockView {...useLockScreenProps(props.route?.params)} />
+    <TouchableWithoutFeedback
+      onPress={Keyboard.dismiss}
```
Ldoppea, line 138: This dismiss the keyboard, but does it remove the focus on text field too?

acezard: it does indeed

Crash--:
No safe area for the button?

Ldoppea:
@acezard I force pushed a small fix.

This import does not work (at least on iOS side) 
```js
import { TouchableWithoutFeedback } from 'react-native-gesture-handler'
```
The behavior is that the `<TouchableWithoutFeedback>` wrapper makes it impossible to click on any button of the screen (only the text input is still interactive)

But if instead we use
```js
import { TouchableWithoutFeedback } from 'react-native'
```

Then we have the correct behavior. It seems like `react-native-gesture-handler` is an old dependency that should be updated or maybe removed. I'll investigate on it.

Crash-- (approved):

PR #447, opened 2022-10-25
docs: JDK 11 is required

paultranvan:


Ldoppea (approved):
I'm ok to add this line, but it is part of the requirements of the RN Environnement setup (cf first line of this list). I'm not sure if it is better to have duplicate info, or to keep a single source of truth. Feel free to merge it if you still consider it useful.

paultranvan:
> I'm ok to add this line, but it is part of the requirements of the RN Environnement setup (cf first line of this list). I'm not sure if it is better to have duplicate info,

From this link, I found:

> Android Studio installs the latest Android SDK by default. Building a React Native app with native code, however, requires the Android 12 (S) SDK in particular.

So it does not seem that SDK 11 is part of the first link (however if we add it, it should be a sub-point of this first bullet, not the second one)

Ldoppea:
I'm talking about this section (this one is for linux but the ones for iOS and windows are similar):

image

paultranvan:
> But I realise that we only gives the link without explaining if the dev should install the Expo Go version or the React Native CLI. Also we don't precise the RN version which may be problematic as the documentation may be updated on future RN version and we may do the update only after a delay.

I slighly changed the commit to move the Java 11 requirement in troubleshooting and insisting on the need to correctly follow the CLI doc ;)

Crash-- (approved):

PR #449, opened 2022-10-26
fix: Prevent app crash when `setFlagshipUI.caller` is `undefined`

Ldoppea:
Based on TS typing, a method's `.caller` cannot be `undefined`

But in practice we encountered some crashes when trying to open cozy apps due to undefined `.caller` (it happens only in dev mode)

So we want to use an optional chaining operator here

Note that this operator was present before, but it has been removed in e6ecf3f2d11594d9cfb044e440b3d05aa81830f3 after the file was migrated to TS, probably due to the linter. So we want to add an `eslint-disable` instruction to prevent rolling back this fix

### DIFF
--- a/src/libs/intents/localMethods.ts
+++ b/src/libs/intents/localMethods.ts
@@ -72,7 +72,7 @@ export const internalMethods = {
     setFlagshipUI(
       intent,
       EnvService.nameIs(strings.environments.test)
-        ? internalMethods.setFlagshipUI.caller.name
+        ? internalMethods.setFlagshipUI.caller?.name // eslint-disable-line @typescript-eslint/no-unnecessary-condition
         : ''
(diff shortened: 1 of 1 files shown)


Ldoppea:
Also note that using `.caller` is not recommended. but I think we can keep it here as it is executed only in `dev` environment.

More info: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Function/caller

Crash-- (approved):

Crash--:
> But in practice we encountered some crashes when trying to open cozy apps due to undefined .caller (it happens only in dev mode)

I +1 this PR, but the real issue to address is why the caller is undefined. This should never happen and we should find out why it happens

PR #451, opened 2022-10-31
fix: Add Hermes to iOS linked libraries

Ldoppea:
When doing a fresh build from scratch, the app would crash on startup with the following error
`Library not loaded: @rpath/hermes.framework/hermes`

To fix this we need to add Hermes to project's linked libraries

More info: https://github.com/facebook/react-native/issues/34601#issuecomment-1239398489


Crash-- (approved):

Ldoppea:
I force-pushed some new config : https://github.com/cozy/cozy-flagship-app/compare/370cc56f243a678bc1f825ff8781c97a15e75cac..801b023e16c6e721f9564f6c462927658d165cf7

Previous edits were not enough to make the app work correctly. They only allowed to run the app on iOS simulators and using XCode debug, but the app would still crash on physical device or in simulator without XCode (re-run previous build).

To make it work on every scenario I add to add Hermes to Embedded libraries as stated in https://github.com/facebook/react-native/issues/34601#issuecomment-1243232921

Crash-- (approved):

PR #453, opened 2022-11-02
Fix: Biometry fixes for iOS

Ldoppea:
This PR handles multiple fixes, each one is described in its own commit, so feel free to check commit messages for more details.

### DIFF
--- a/src/ui/Container/index.tsx
+++ b/src/ui/Container/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { View, ViewProps } from 'react-native'
+import { SafeAreaView, View, ViewProps } from 'react-native'
 import { getNavbarHeight } from '/libs/dimensions'
 
 import { styles } from '/ui/Container/styles'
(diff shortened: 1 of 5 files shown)


src/ui/Container/index.tsx
```diff
@@ -15,6 +15,6 @@ export const Container = ({
     style={[styles.container, { paddingBottom: getNavbarHeight() + 16 }, style]}
     {...props}
   >
-    {children}
+    <SafeAreaView>{children}</SafeAreaView>
```
Crash--, line 18: nit: What do you think of creating a CozyView that wraps SafeArea view on iOS and nothing on android? Can it makes sense?

Crash-- (approved):

src/ui/Container/index.tsx
Ldoppea: I just did the test about double margins. The SafeAreaView is able to adapt to container's position

Here is a version with the container's height=100% (green = container, red = SafeArea)
image

Here is a version with the container's height=99%
image

Here is a version with the container padding bottom being less than the Safe area margin
image

The `déverrouiller` button won't move if the container's padding is moving but stays under the SafeArea limit.

The behavior is the same for the top margin.

So it seems like we can safely keep the SafeAreaView in all our view without risk of margin conflict.

PR #457, opened 2022-11-07
feat: Handle denied biometry access on iOS

Ldoppea:
On iOS with FaceID the biometry feature is guarded by OS level authorizations

If the user reject the authorisation process, then the application won't be able to use FaceID anymore until the user re-authorizes it from the OS settings

We want this information to be available from the settings cozy-app and to be able to redirect the user to the OS settings
so they can re-autorize the biometry

___

### Related PRs

- https://github.com/cozy/cozy-libs/pull/1870
- https://github.com/cozy/cozy-settings/pull/512
___

### TODO

- [x] Update cozy-intent and cozy-device-helper after https://github.com/cozy/cozy-libs/pull/1870 being merged

___

#### First access:
- Start the app for the first time (re-install it if needed) and login
- Open cozy-settings/lockscreen
- Check `Unlock with Face ID`
- On the OS dialog: reject the authorisation
Screenshot 2022-11-07 at 17 55 17

- Nothing should change on settings
- Try to check `Unlock with Face ID` again, a message dialog should appear (texts to be changed)
Screenshot 2022-11-07 at 17 57 52

- Click `Open settings`
- App's OS settings should be opened

#### During app's lifecycle:
- Open the Cozy app
- Open cozy-settings/lockscreen
- Check `Unlock with Face ID`
- Open iOS Settings -> Cozy Cloud
- Uncheck authorization for Face ID
Screenshot 2022-11-07 at 18 01 08

- Go back on the Cozy app (it should has restart)
- Open cozy-settings/lockscreen
- The `Unlock with Face ID` should appear unchecked
- Try to Check `Unlock with Face ID`, a message dialog should appear
- Go back on iOS Settings and authorize Face ID again
- Go back on Cozy app (it should has restart)
- Open cozy-settings/lockscreen
- The `Unlock with Face ID` should appear checked


src/libs/intents/localMethods.ts
```diff
@@ -67,6 +68,17 @@ const fetchSessionCodeWithClient = (
...
 }
 
+const openAppOSSettings = async (): Promise<null> => {
+  if (Platform.OS === 'android') {
+    throw new Error(
+      `openAppOSSettings shouldn't be called from Android as no authorization is needed for biometry`
+    )
+  }
+
+  await Linking.openURL('app-settings:')
```
Crash--, line 79: Oh nice <3

src/libs/intents/setBiometryState.ts
```diff
@@ -40,6 +42,11 @@ export const openSettingBiometry = async (): Promise<boolean> => {
 export const makeFlagshipMetadataInjection = async (): Promise<string> => {
   const rnBiometrics = new ReactNativeBiometrics()
 
+  const sensorData = await rnBiometrics.isSensorAvailable()
+  const biometryDenied = BIOMETRY_DENIED_BY_USER_IOS_ERROR.test(
+    sensorData.error ?? ''
+  )
```
Crash--, line 59: Why don't you use the `isBiometryDenied` you created in the next commit?

Ldoppea: At first I used the Metadata injection as it was not supposed to change during the app's lifecycle (when changing OS settings, the app restarts automatically)
But then I realized that on the first launch, the authorization would change this prop's state without restarting the app. So I implemented the new "on demand" method.

I hesitated to refactor the first commit and also use the `onDemand`. This would imply calling an async method on cozy-settings side to initialize the component state
image
Nothing difficult here. I'm wondering if this would introduce some flickering (but maybe it's a false assumption). I'll try it.

acezard (approved):

src/libs/intents/setBiometryState.ts
Crash--: Yep it was just about reusing the method. I thought at first that the code was exactly the same.

Crash-- (approved):

PR #458, opened 2022-11-08
fix: Decouple LockScreen bars from Flagship bars

acezard:
### Feature

Unlock screen sucess 

### Context

The unlock screen was tampering with the flagshipUi, creating unwanted side-effects on app re-open, eg on cozy-banks which was losing its blue bars

### Implementation

Decoupling the UI, lockScreen now has its own 'flagshipUI' event/view and can't create side effects out of its own bubble. It still use pure functions from setFlagshipUi that are now exported

### Testing

Manual only

### Thoughts

Code is verbose as is usual with setFlagshipUi but this topic can't be tackled in a bug fix


src/screens/lock/events/LockScreen.events.ts
```diff
@@ -0,0 +1,35 @@
...
+export const setLockScreenUI = (intent: FlagshipUI): Promise<null> => {
+  handleSideEffects(
+    Object.fromEntries(
+      Object.entries({
+        ...intent,
+        bottomTheme: formatTheme(intent.bottomTheme as ThemeInput),
+        topTheme: formatTheme(intent.topTheme as ThemeInput)
+      })
+        .filter(([, v]) => v)
+        .map(([k, v]) => [k, v?.trim()])
```
Ldoppea, line 30: Do we really need this complexity as we have control on the caller code?

src/screens/lock/view/LockScreenBars.tsx
```diff
@@ -28,12 +28,12 @@ export const FlagshipBars = ({
   } = UIState
 
   useEffect(() => {
-    flagshipUI.on('change', (state: NormalisedFlagshipUI) => {
+    lockScreenUi.on('lockscreen:change', (state: NormalisedFlagshipUI) => {
       setUIState({ ...UIState, ...state })
```
Ldoppea, line 32: By using `setUIState(oldState => ({...oldState, ...state}))` you can remove the hook dependency to UIState and so reduce the amount of call to `.removeAllListeners()`/`.on()`

Ldoppea (approved):

src/screens/lock/events/LockScreen.events.ts
acezard: In my opinion, we do not necessarily have control over the caller code, that's why I added this initially. Many apps aren't typed, and who knows what is coming through during runtime? I don't feel safe if I don't sanitize the input

src/screens/lock/view/LockScreenBars.tsx
acezard: not sure here, where do we get the `oldState`?

Ldoppea: this is provided by the useState API: https://reactjs.org/docs/hooks-reference.html#usestate (search for `prevState`)

`setState` can get a value, or a function callback that takes the previous state in argument

src/screens/lock/events/LockScreen.events.ts
acezard: Yes, that is true, I copy pasted from flagshipUI. I guess we can remove it since the scope is fairly limited for now

src/screens/lock/view/LockScreenBars.tsx
acezard: yea good catch

PR #461, opened 2022-11-10
Fix/total energies connector

KillianCourvoisier:

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/connectors/totalenergies/src/index.js
+++ b/connectors/totalenergies/src/index.js
@@ -266,7 +266,7 @@ class TemplateContentScript extends ContentScript {
     const startDate = new Date(year, month, day)
     const href = contractElement.querySelectorAll('.ml-std')[1].children[1].getAttribute('href')
     const fileurl = `https://www.totalenergies.fr${href}`
-    const filename = `${day}-${month}-${year}_TotalEnergie_Contrat_${offerName.replaceAll(' ', '-')}.pdf`
+    const filename = `${year}-${month}-${day}_TotalEnergie_Contrat_${offerName.replaceAll(' ', '-')}.pdf`
     const contract = [{
       filename,
       fileurl,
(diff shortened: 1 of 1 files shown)


connectors/totalenergies/src/index.js
```diff
@@ -258,15 +258,15 @@ class TemplateContentScript extends ContentScript {
...
+    const rawStartDate = contractElement.querySelector('p[class="font-700"]').innerHTML
     const splittedStartDate = rawStartDate.split('/')
     const day = splittedStartDate[0]
     const month = splittedStartDate[1]
     const year = splittedStartDate[2]
     const startDate = new Date(year, month, day)
     const href = contractElement.querySelectorAll('.ml-std')[1].children[1].getAttribute('href')
     const fileurl = `https://www.totalenergies.fr${href}`
-    const filename = `${day}-${month}-${year}_TotalEnergie_Contrat_${offerName.replaceAll(' ', '-')}.pdf`
+    const filename = `${year}-${month}-${day}_TotalEnergie_Contrat_${offerName.replaceAll(' ', '-')}.pdf`
```
paultranvan, line 269: We should probably have a `normalizeFileName` function shared accross konnectors to avoid reinventing the wheel and remain consistent ;)

connectors/totalenergies/src/index.js
```diff
@@ -258,15 +258,15 @@ class TemplateContentScript extends ContentScript {
     this.log('getContract starts')
     const contractElement = document.querySelector('.cadre2')
     const offerName = contractElement.querySelector('h2').innerHTML
-    const rawStartDate = contractElement.querySelector('p[class="txt-gras"]').innerHTML
+    const rawStartDate = contractElement.querySelector('p[class="font-700"]').innerHTML
```
paultranvan, line 261: That looks quite fragile, but I guess this is inevitable with client-side konnectors :man_shrugging:

paultranvan (approved):

connectors/totalenergies/src/index.js
KillianCourvoisier: Yup, this is planed, it should be added anytime from now :)

connectors/totalenergies/src/index.js
KillianCourvoisier: The only thing I may try to avoid this is to get the parent and go down in it's child, but i'm pretty sure this is not robust enough too ...

PR #468, opened 2022-11-16
Fix/resync cookies on resume

Ldoppea:
> **Note**
> This PR contains some commits that are not related to the main PR feature but are needed
> All commits have their own description so it is easier to review this PR commit-by-commit

In 6d38a288164c5800520b8964c33f0a0e0ca7dc2a we implemented Cookies re-sync feature on `SupervisedWebView` component

This would allow to enforce cookies restauration when `SupervisedWebView` is restored in order to avoid authentication loss

But we found that there is still some rare cases when we loose authentication to the cozy-stack

We suspect that the CookieManager also have a separated lifecycle and can be disposed independently from the `SupervisedWebView`

To mitigate this potential issue, we want to re-sync cookies every time the App resumes (not only when `SupervisedWebView` is restored)

___

TODO:

- [ ] Fix `tsc` error due to CozyClient type definition on `lockScreenFunctions.ts`
- [ ] Move `resyncCookiesOnAppResume` outside of App.js like in #465 (see `useGlobalAppState.ts`)


Ldoppea:
This PR is ready for review

Ldoppea:
@acezard I added a new TS rule to enforce index checks when manipulating arrays.

Due to this rule I had to edit the ErrorScreen.jsx file. I made a simple test to ensure it still works, but I would appreciate if you double check this https://github.com/cozy/cozy-flagship-app/pull/468/commits/808aadac9537791dcb328dd8cad557b582e45043

src/hooks/useCookieResyncOnResume.ts
```diff
@@ -0,0 +1,51 @@
+import { AppState, AppStateStatus } from 'react-native'
+import { useEffect } from 'react'
+
+import Minilog from '@cozy/minilog'
+import CozyClient, { useClient } from 'cozy-client'
+
+import { resyncCookies } from '/libs/httpserver/httpCookieManager'
+
+const log = Minilog('useGlobalAppState')
```
acezard, line 9: Shouldn't it be 'useCookieResyncOnResume'?

src/hooks/useCookieResyncOnResume.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    appState = nextAppState
+  }
+
+/**
+ * In some scenario CookieManager cookies are not applied to the WebView on
+ * iOS reload (when killed while app in background state)
+ * To prevent this we enforce resetting cookies on App's resume
+ */
+export const useCookieResyncOnResume = (): void => {
```
acezard, line 37: As `<App />` is impossible to test, maybe it would be worthy to have a unit test for this hook (using a fake App component, not using renderHook())

src/libs/httpserver/httpCookieManager.ts
```diff
@@ -75,10 +82,14 @@ const parseCookie = cookieString => {
  * @param {CozyClient} client - CozyClient instance
  * @returns {Promise<boolean>}
  */
-export const setCookie = async (cookieString, client) => {
+export const setCookie = async (
```
acezard, line 85: These functions seem interesting to test (only the exported ones)

src/libs/httpserver/server-helpers.ts
```diff
@@ -16,7 +16,7 @@ export const addBodyClasses = (HTMLstring: string): string => {
 
   const hasClasses = /<body.*?class=("|')(.*?)("|')/i.exec(HTMLstring)
 
-  if (!hasClasses)
+  if (!hasClasses || hasClasses[2] === undefined)
```
acezard, line 19: We should explain this condition, why are we looking for the 3rd class to be undefined?

src/libs/intents/localMethods.spec.js
```diff
@@ -7,28 +7,35 @@ import { localMethods, asyncLogout } from './localMethods'
 
 jest.mock('react-native-keychain')
 jest.mock('../RootNavigation.js')
+jest.mock('@react-native-cookies/cookies', () => ({
+  clearAll: jest.fn()
+}))
 
 describe('asyncLogout', () => {
+  const client = {
```
acezard, line 15: I think this whole file was a bad idea from the start. We should have one file per tested method, in .ts.
Maybe for this PR it is possible to have a `asyncLogout.spec.ts` and ignore the rest of the file?

src/libs/intents/localMethods.ts
```diff
@@ -15,21 +15,27 @@ import { setFlagshipUI } from './setFlagshipUI'
 import { showInAppBrowser, closeInAppBrowser } from './InAppBrowser'
 import { toggleSetting } from '/libs/intents/toggleSetting'
 import { isBiometryDenied } from '/libs/intents/setBiometryState'
-import AsyncStorage from '@react-native-async-storage/async-storage'
+import { clearData } from '/libs/localStore/storage'
+import { clearCookies } from '/libs/httpserver/httpCookieManager'
 
-export const asyncLogout = async (): Promise<null> => {
+export const asyncLogout = async (client?: CozyClient): Promise<null> => {
```
acezard, line 21: As for the test, I think this function should have its own file (like all cozy-intents functions anyway). Not saying we should refactor everything asap, but doing it when adding/updating code

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "moduleResolution": "node",
     "noEmit": true,
     "noFallthroughCasesInSwitch": true,
+    "noUncheckedIndexedAccess": true,
```
acezard, line 16: Good find 👍

acezard (approved):
None of my comments are blocking 👍

src/libs/httpserver/httpCookieManager.ts
Ldoppea: Those tests already exist in: https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/httpserver/httpCookieManager.spec.js

Is that what you are talking about?

src/libs/httpserver/server-helpers.ts
Ldoppea: If you are talking about the `if` check, this is just because we use `hasClasses[2]` two lines later and the `noUncheckedIndexedAccess` rule now enforces this check

If you are talking about the use of `hasClasses[2]`, this is because the regex is formed like this with 3 capturing groups. Maybe we should add a comment with some exemples that match the regex?
Also maybe we should use non-capturing groups instead for the quotes: `<body.*?class=(?:"|')(.*?)(?:"|')`

src/libs/httpserver/httpCookieManager.ts
acezard: Yea my bad!

src/libs/httpserver/server-helpers.ts
acezard: I understand, and I was the one to write the regex if memory serves me right... I guess it proves that the piece of code is obscure without a brief comment explaining what's going on in this condition.

src/hooks/useCookieResyncOnResume.ts
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/compare/d79a634f6a0e19820c433eb60a85f072b7327c74..d97e25177cceed3efb818a22c57d1d3212b358ba

src/libs/intents/localMethods.spec.js
Ldoppea: I would prefer to do this in another commit

PR #469, opened 2022-11-23
fix: Get initial offline route correctly

acezard:
Strangely enough, the error page used `options` instead of `initialParams` to set the error type. So in some cases there was nothing to display because we didn't have any params even if the ErrorScreen was asked.

### DIFF
--- a/src/config/dev.ts
+++ b/src/config/dev.ts
@@ -1,6 +1,7 @@
 export const devConfig = {
   disableGetIndex: false, // Make cozy-app hosted by webpack-dev-server work with HTTPServer
   forceHideSplashScreen: false, // Hide react-native splash screen renders in development mode,
-  ignoreLogBox: true, // Hide react-native logbox renders in development mode,
-  sentry: false // Allow Sentry logging in development mode
+  ignoreLogBox: false, // Hide react-native logbox renders in development mode,
+  forceOffline: false, // Force offline mode in development mode,
+  sentry: false // Allow Sentry logging in development mode,
 }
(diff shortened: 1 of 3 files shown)


src/config/dev.ts
```diff
@@ -0,0 +1,7 @@
+export const devConfig = {
+  disableGetIndex: false, // Make cozy-app hosted by webpack-dev-server work with HTTPServer
+  forceHideSplashScreen: false, // Hide react-native splash screen renders in development mode,
+  ignoreLogBox: false, // Hide react-native logbox renders in development mode,
+  isOffline: false, // Force offline mode in development mode,
```
Ldoppea, line 5: I would prefer `forceOffline` to be consistent with `forceHideSplahScreen` but also to avoid confusion in code when reading `if (isOffline)` (i.e. if false, then we are online)

src/libs/services/NetService.js
```diff
@@ -46,9 +47,11 @@ const handleOffline = () =>
...
 const toggleNetWatcher = makeNetWatcher()
 
-export const NetService = {
+const NetService = {
   handleOffline,
-  isConnected,
-  isOffline,
-  toggleNetWatcher
+  isConnected: devConfig.isOffline ? isOffline : isConnected,
+  isOffline: devConfig.isOffline ? isConnected : isOffline,
```
Ldoppea, line 53: My understanding is that `devConfig.isOffline` is just inverting the result here. So the app would consider you to be online if you temporarily lose the connexion (although highly improbable). Is that intented?

Ldoppea (approved):

PR #470, opened 2022-11-24
feat: configure Netinfo plugin

acezard:
# Feature

We do not want to ping google servers to check if the user is connected or not

# Implementation
- [x] Update `Netinfo` to the latest version, using the newest features
- [x] Use Cloud URL if app Client is null
- [x] Use Stack URL if app Client is defined


src/libs/services/NetService.ts
```diff
@@ -2,16 +2,33 @@ import NetInfo, {
...
 import { reset } from '/libs/RootNavigation'
 import { routes } from '/constants/routes'
-import { showSplashScreen } from './SplashScreenService'
+import { showSplashScreen } from '/libs/services/SplashScreenService'
 
 const log = Minilog('NetService')
 
+const configureService = (client?: CozyClient): void => {
+  NetInfo.configure({
+    reachabilityUrl: client
```
Ldoppea, line 19: Care, after the first login, then `client` is never cleared even after login out (only restarting the app can do that).

This may query the wrong cozy if the user logout the login to another one (the bug would only occur during login).

I had to handle this on [this commit](https://github.com/cozy/cozy-flagship-app/pull/468/commits/37508c62f0c96d824027a82ea7abce021a5cc7c8) then [this commit](https://github.com/cozy/cozy-flagship-app/pull/468/commits/24dfd8c15463104a9cc2b36bee2bc6873beec748#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67R101)
Unfortunately they are not merged yet, i'll try to do it in the following days.

acezard: I think this is a low-priority PR, should I wait for your work to be merged?

Crash--: In order to respond to @LucsT's case:
- You need to add a reachabilityTest 
- You need to add reachabilityShouldRun() => true 
- Add useNativeReachability false 

But I don't know how the `NetInfoState` reacts to the connectivity change where this is no `nativeReachability`. 

So I'm wondering if we should not just keep the NetInfo as is in this PR, but also add a `fetch` (not the netinfo one) to the `reachabilityUrl` when launching the app.

Crash--: > In my opinion we could add a React error handler and redirect to the offline screen if the error is network related

If you know how to do that then go for it ;)

Ldoppea (approved):

PR #471, opened 2022-11-25
feat: update every testing library to latest

acezard:
Also removed some typing libraries that appear of no use,
since the typings are embedded in the actual libraries now.
Which also avoids mismatched types/implementation versions.


package.json
```diff
@@ -84,12 +84,10 @@
     "@babel/preset-typescript": "7.16.7",
     "@babel/runtime": "^7.17.9",
     "@svgr/webpack": "5.5.0",
-    "@testing-library/react-hooks": "^7.0.2",
-    "@testing-library/react-native": "^9.1.0",
-    "@types/jest": "^28.1.4",
```
Ldoppea, line 89: Isn't this needed for those typings? https://github.com/cozy/cozy-flagship-app/pull/470/commits/865b856ec737e57679fba1eea7ef9f17fe0b1952#diff-96cc3b955ac66dd55214d227f6c48d0c0b7ff2a8ae79a06a6d0c56690110dac3R7

acezard: From what I see the actual jest package comes with its typings now

acezard: @Ldoppea 
image

acezard: @Ldoppea also: https://github.com/facebook/jest/blob/6e5b1d60a1214e792b5229993b5475445e9c1a6e/packages/jest/package.json#L17

Ldoppea (approved):

PR #472, opened 2022-11-25
refactor: typescript improvements

acezard:
Further actions have been taken toward typings for the continued health of the application.

- [x] Upgrade TypeScript environment
- [x] Upgrade the linting environment
- [x] Convert every file in `/libs/intents` to TypeScript
- [x] Fix every **tsc**/**eslint** errors that arise in those files


package.json
```diff
@@ -87,21 +87,24 @@
...
+    "@types/react-native": "0.66.4",
+    "@typescript-eslint/eslint-plugin": "^5.44.0",
     "babel-loader": "8.2.2",
     "babel-plugin-module-resolver": "4.1.0",
     "babel-plugin-optional-require": "0.3.1",
     "babel-plugin-react-native-web": "0.17.7",
     "eslint": "^8.23.0",
-    "eslint-config-cozy-app": "^5.1.0",
+    "eslint-config-cozy-app": "^5.2.0",
+    "eslint-plugin-prettier": "4.2.1",
```
acezard, line 98: Ok so this is annoying but for now, the cozy eslint config's plugins aren't reachable by the flagship eslint config. For now, it's best to add the plugins in the app while working towards a solution in `eslint-config-cozy-app`

src/libs/intents/localMethods.spec.ts
```diff
@@ -45,9 +48,9 @@ describe('asyncLogout', () => {
 
 describe('backToHome', () => {
   it('should handle Navigation', async () => {
-    const client = {}
+    const client = {} as unknown as CozyClient
 
-    localMethods(client).backToHome()
+    await (localMethods(client) as NativeMethodsRegister).backToHome()
```
Ldoppea, line 53: Why do we need to specify `as NativeMethodsRegister` here? There seems to be no ambiguity from the import.

src/libs/intents/localMethods.spec.ts
```diff
@@ -59,23 +62,32 @@ test('fetchSessionCode should return only a session code', async () => {
...
+      fetchSessionCode
+    })
+  }
   fetchSessionCode.mockResolvedValue({
     twoFactorToken: 'token'
   })
 
-  await expect(localMethods(client).fetchSessionCode()).rejects.toThrowError(
+  await expect(
+    localMethods(client as CozyClient).fetchSessionCode()
```
Ldoppea, line 89: nit: I would prefer to put the `as CozyClient` on the declaration (`const client = ...`)

src/libs/intents/localMethods.ts
```diff
@@ -34,8 +36,9 @@ export const asyncLogout = async (client?: CozyClient): Promise<null> => {
 
 // Since logout is used from localMethods
 // it can't be async for now.
-const logout = (client?: CozyClient): (() => Promise<null>) => {
-  return async () => asyncLogout(client)
+const logout = async (client?: CozyClient): Promise<null> => {
+  await asyncLogout(client)
+  return null
```
Ldoppea, line 41: Why do we need a `return null` here? Doesn't `asyncLogout` already do this?

src/libs/intents/localMethods.ts
```diff
@@ -46,25 +49,22 @@ const backToHome = (): Promise<null> => {
...
   return async function fetchSessionCode() {
     if (!client) {
       return null
     }
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
-    const sessionCodeResult = await client.getStackClient().fetchSessionCode()
 
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    const sessionCodeResult = await (
+      client.getStackClient() as {
```
Ldoppea, line 62: Maybe we should declare those types somewhere else. I think that doing this inline makes the code less readable (too much noise).

src/libs/intents/localMethods.ts
```diff
@@ -91,27 +91,43 @@ export const internalMethods = {
     setFlagshipUI(
       intent,
       EnvService.nameIs(strings.environments.test)
-        ? internalMethods.setFlagshipUI.caller?.name // eslint-disable-line @typescript-eslint/no-unnecessary-condition
+        ? // @TODO: find why the function caller was determined to be possibly undefined
```
Ldoppea, line 94: Should we fix this TODO before merge?

src/libs/intents/localMethods.spec.ts
Ldoppea: Edit: ok I just saw the `CustomMethods` thing

src/libs/intents/localMethods.ts
```diff
@@ -91,27 +91,43 @@ export const internalMethods = {
...
+  async (): Promise<null> => {
+    await method()
+
+    return null
+  }
+
 export const localMethods = (
   client: CozyClient | undefined
-): NativeMethodsRegister => {
+): NativeMethodsRegister | CustomMethods => {
```
Ldoppea, line 120: Why is this new type needed?

Ldoppea (approved):

src/libs/intents/localMethods.ts
acezard: Resolved in another PR

src/libs/intents/localMethods.spec.ts
acezard: Updated

src/libs/intents/localMethods.ts
acezard: It should be added into cozy-intent NativeMethodsRegister when I have some time. I think it's ok to have this temporarily though

src/libs/intents/localMethods.ts
acezard: Correct, fixed

src/libs/intents/localMethods.ts
acezard: I agree, I will start adding these types into cozy-client directly so we can remove them. In the meantime I improved readability by removing inline typing

PR #473, opened 2022-11-29
feat: handle failure scenario in mailto handling

acezard:
In case the user didn't have any mailing app, we weren't handling anything. This has been fixed.

This PR also includes several multiples improvements discovered while developing the feature


src/screens/error/ErrorScreen.tsx
```diff
@@ -1,30 +1,63 @@
...
+  event: WebViewMessageEvent
+): Promise<undefined[]> =>
+  await Promise.all(
+    Object.keys(handlers).map(async eventId => {
+      if (!event.nativeEvent.data.includes(eventId)) return undefined
+
+      try {
+        await handlers[eventId]()
+      } catch (error) {
+        log.error('Error while handling event', eventId, error)
```
Ldoppea, line 49: Are we sure we want to silently catch ALL events? Here our need is to silently catch only the `mailto` one.

src/hooks/useAppBootstrap.js
```diff
@@ -140,7 +140,11 @@ export const useAppBootstrap = client => {
 
         setInitialRoute({
           stack: isHome ? fallback : undefined,
-          root: !isHome ? fallback : undefined
+          root: isConnected
+            ? !isHome
+              ? fallback
+              : undefined
+            : strings.errorScreens.offline
```
Ldoppea, line 147: I don't understand why this is needed (and also why it isn't needed for the `stack` prop). Can you describe a scenario in the commit message?

src/screens/error/ErrorScreen.tsx
acezard: Updated so now the event has to declare if it wants to be caught or not. I set `backButton` to `false` and `mailto` to `true`

src/hooks/useAppBootstrap.js
acezard: This is once more an issue with app bootstrap complexity. This provides us initial error param as offline, whereas before it was handled with a race condition on the handleOffline() function. My main objective in terms of refactoring is to tackle this file and make initial routes understandable.

src/screens/error/ErrorScreen.tsx
Ldoppea: Why not just moving the try-catch in the handler that needs it?

Ldoppea (approved):

PR #479, opened 2022-12-02
fix: Reconfigure NetService after logout

Ldoppea:
In #468 we fixed logout method to correctly logs out cozy-client

This state change should be detected by NetService so it can reconfigure itself to ping the Cloudery instead of the previous cozy-stack

> **Note**
> This PR can be merged before #468 , it will just do nothing until this other PR is merged

### DIFF
--- a/src/libs/services/NetService.ts
+++ b/src/libs/services/NetService.ts
@@ -17,7 +17,7 @@ const log = Minilog('NetService')
 
 const configureService = (client?: CozyClient): void => {
   NetInfo.configure({
-    reachabilityUrl: client
+    reachabilityUrl: client?.isLogged
       ? `${(client.getStackClient() as { uri: string }).uri}/${
           strings.reachability.stack
         }`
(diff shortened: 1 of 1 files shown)


src/libs/services/NetService.ts
```diff
@@ -27,7 +27,17 @@ const configureService = (client?: CozyClient): void => {
 
 export const useNetService = (client?: CozyClient): void =>
   useEffect(() => {
-    configureService(client)
+    const configure = (): void => {
+      configureService(client)
+    }
+
+    client?.on('logout', configure)
```
acezard, line 34: A good solution I didn't know we had events like that 👍.
Are we sure that the client ref in the hook won't change on the logout event? I think it should not, but if it did that would configure twice.

acezard (approved):

acezard:
I added two small commits after reviewing the NetService file and finding two small issues.

PR #488, opened 2022-12-09
feat: add NetBoundary component

acezard:
It will show offline screen if app is offline on boot.
It will redirect to app if connexion is back.
It will not redirect to offline is connexion is lost.


src/screens/error/ErrorScreen.tsx
```diff
@@ -14,7 +14,7 @@ const log = Minilog('ErrorScreen')
 /**
  * Typings
  */
-interface ErrorScreenProps {
+export interface ErrorScreenProps {
```
Ldoppea, line 17: Seems like you exported this interface but did not use it elsewhere. Is this intended?

Ldoppea (approved):

src/screens/error/ErrorScreen.tsx
acezard: You're right I did use it at first. The export is not needed anymore

PR #489, opened 2022-12-09
fix: catch error in getting caller name

acezard:
This happens on cozy-banks, probably linked to bad cozy-ui/cozy-intent versions

Simple fix native side is to ignore caller name if we don't have access to it

### DIFF
--- a/src/libs/intents/localMethods.ts
+++ b/src/libs/intents/localMethods.ts
@@ -87,13 +87,19 @@ const openAppOSSettings = async (): Promise<null> => {
 }
 
 export const internalMethods = {
-  setFlagshipUI: (intent: FlagshipUI): Promise<null> =>
-    setFlagshipUI(
-      intent,
-      EnvService.nameIs(strings.environments.test)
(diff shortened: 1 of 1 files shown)


src/libs/intents/localMethods.ts
```diff
@@ -87,13 +87,19 @@ const openAppOSSettings = async (): Promise<null> => {
...
-        : ''
-    )
+  setFlagshipUI: (intent: FlagshipUI): Promise<null> => {
+    const caller = (): string => {
+      if (!EnvService.nameIs(strings.environments.test)) return 'unknown'
+
+      try {
+        return internalMethods.setFlagshipUI.caller.name
+      } catch (error) {
+        return 'unknown'
```
Ldoppea, line 97: What is the advantage of try...catching this instead of using optional chaining operators?

acezard: Even with them, it throws an error. I'm not sure exactly what is the problem, but it's linked to security. It doesn't allow me to get the caller name in some contexts

Ldoppea (approved):

PR #498, opened 2022-12-15
fix: inject client in asyncLogout

acezard:
This fixes a regression. A previous commit removed the client injection
into the logout intent, so the functionality was disabled.
We correct that mistake by adding the client in asyncLogout call

### DIFF
--- a/src/libs/intents/localMethods.ts
+++ b/src/libs/intents/localMethods.ts
@@ -34,12 +34,6 @@ export const asyncLogout = async (client?: CozyClient): Promise<null> => {
   return Promise.resolve(null)
 }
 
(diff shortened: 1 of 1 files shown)


src/libs/intents/localMethods.ts
```diff
@@ -34,12 +34,6 @@ export const asyncLogout = async (client?: CozyClient): Promise<null> => {
   return Promise.resolve(null)
 }
 
-// Since logout is used from localMethods
-// it can't be async for now.
```
acezard, line 38: Well that's definitely a wrong statement now, this function has no use anymore

Ldoppea (approved):

PR #499, opened 2022-12-15
docs: Add port and url to use when working on Android with locally ho…

zatteo:
…sted webviews

When blindly following these steps, I forgot to add a custom port like 8080. So app could not connect to the stack. I also did not write explicitly http which can lead to connection issues.


Crash-- (approved):
<3

PR #501, opened 2022-12-16
fix: Force LockScreen to be on top of the UI

Ldoppea:
On iOS when the users opens a file (PDF, Image, etc), then a QuickLook
controller is displayed on top of the app with the file content

This conflicts with the Lock feature that should show the Lock UI on
top of everything

With current implementation, the QuickLook controller would still be
displayed on top of the Lock screen (this happens if the users send the
apps into the background while viewing a file)

By wrapping the Lock screen into a `FullWindowOverlay` we ensure that
it will be displayed on top of every other elements

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -398,8 +398,9 @@ PODS:
     - React
   - RNKeychain (8.0.0):
     - React-Core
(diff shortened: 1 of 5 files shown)


Crash-- (approved):
Nice 🎉

PR #510, opened 2022-12-19
Fix/alan connector cookies

KillianCourvoisier:
Fixing Alan connector. Using cookieManager from the RNLauncher to be able to retrieve the token hidden in the httpOnly cookies to make the requests on Alan's API.

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/connectors/alan/src/index.js
+++ b/connectors/alan/src/index.js
@@ -180,7 +180,7 @@ class TemplateContentScript extends ContentScript {
         userCredentials
       })
     }
-    if(document.location.href.includes(`${HOMEPAGE_URL}`) && document.querySelector('a[href="#"]') || document.querySelector('div[class="murray__NavListItem"]')){
+    if(document.location.href.includes(`${HOMEPAGE_URL}`) && document.querySelector('a[href="#"]') || document.querySelector('div[class="ListItem ListItem__Clickable ListCareEventItem"]')){
       this.log('Auth Check succeeded')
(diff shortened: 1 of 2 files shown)


connectors/alan/src/index.js
```diff
@@ -119,6 +117,7 @@ class TemplateContentScript extends ContentScript {
     await this.waitForElementInWorker('a[href="/login"]')
     await this.clickAndWait('a[href="/login"]', 'a' )
     await this.waitForElementInWorker('a')
+    // await this.waitForElementInWorker('[pause]')
```
Crash--, line 125: to remove?

connectors/alan/src/index.js
```diff
@@ -165,7 +164,7 @@ class TemplateContentScript extends ContentScript {
         userCredentials
       })
     }
-    if(document.location.href.includes(`${HOMEPAGE_URL}`) && document.querySelector('a[href="#"]') || document.querySelector('div[class="murray__NavListItem"]')){
+    if(document.location.href.includes(`${HOMEPAGE_URL}`) && document.querySelector('a[href="#"]') || document.querySelector('div[class="ListItem ListItem__Clickable ListCareEventItem"]')){
```
Crash--: So, this is not only "use cookie manager to retrieve token". You also changed a selector. Maybe a dedicated commit?

PR #511, opened 2022-12-19
fix: downgrade commons-io in gzip lib to support Android 7+

acezard:
Sentry reported a bug caused by `react-native-gzip`,
created by `commons-io@2.6`.

We decide to patch the library with `patch-package`
to use a commons-io version
compatible with Android 7+

The application was tested on AVD with v7.1 and was not able to reproduce the error

### DIFF
--- a/patches/@fengweichong+react-native-gzip+2.0.0.patch
+++ b/patches/@fengweichong+react-native-gzip+2.0.0.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/@fengweichong/react-native-gzip/android/build.gradle b/node_modules/@fengweichong/react-native-gzip/android/build.gradle
+index c2e330f..f28740e 100644
+--- a/node_modules/@fengweichong/react-native-gzip/android/build.gradle
++++ b/node_modules/@fengweichong/react-native-gzip/android/build.gradle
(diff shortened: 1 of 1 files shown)


Ldoppea:
Did you try to run this on lower Android versions like Android 5? As this is our minimum supported version it would be great to check this.

acezard:
> Did you try to run this on lower Android versions like Android 5? As this is our minimum supported version it would be great to check this.

I only tried on 7.1
But I'm not sure I did the correct scenario to reproduce the bug, do you know how to exactly reproduce it so I can be 100% sure?

Ldoppea:
> do you know how to exactly reproduce it so I can be 100% sure?

You have to start the app and wait 10s for the cozy-app's bundle download to be done. If you get this without error in the console, it should be OK

image


The gzip extraction is done in `cozyAppBundle.extractCozyAppArchive`

Ldoppea:
You can also change `BUNDLE_UPDATE_DELAY_IN_MS` so you won't have to wait 10s

acezard:
> You can also change `BUNDLE_UPDATE_DELAY_IN_MS` so you won't have to wait 10s

Ok so I tested correctly without knowing! Will try v5

PR #516, opened 2022-12-23
feat: Add redux store for handling current running connector and logs

Ldoppea:
We want to keep track of the current running connector so we can
prevent multiple connectors to run at the same time. Also we want to
persist this info so we can mark the connector as failed after a
restart if the app crashed during its execution

We also want to persist connectors logs so we can still send them to the
cozy-stack if the app crashes before being able to send them

This commits add the foundation for managing connectors through redux

The complete logic is not implemented as we only implement the store
container with all the redux logic

___

Usage:

```jsx
import { useConnectorState } from 'src/providers/ConnectorState/ConnectorStateProvider'

// in component's logic
const { addLog, currentRunningConnector, setCurrentRunningConnector } = useConnectorState()

const doSomething = () => {
  setCurrentRunningConnector('some_slug')
  addLog('some logs')
}
```

___

TODO:

- [x] Check how to clear the store on logout (⚠️ connector logs must be sent to stack before clearing the store)
- [x] Verify if the currentRunningConnector should be persisted or not


acezard:
Are we sure we want to use Redux here? It's very verbose, maybe we can look at lighter alternatives?

Ldoppea:
> Are we sure we want to use Redux here? It's very verbose, maybe we can look at lighter alternatives?

What alternative do you have in mind?

Here we chose react-redux because we want a state management that automatically persist its data, and because it is already used by cozy-client (so we keep the same stack).

Crash--:
@Ldoppea We should talk about this. If you need slice within redux, you should combine the one from cozy-client and the one you want to add and then make a `setStore` call to cozy-client. https://github.com/cozy/cozy-banks/blob/e1a2b5435f4be2d55a744190aafc243005c857f1/src/main.jsx#L118 for instance (to be able to do that, when you initiliaze cozy-client, you should set store to false (https://github.com/cozy/cozy-banks/blob/e1a2b5435f4be2d55a744190aafc243005c857f1/src/ducks/client/web.js#L41) 

You should also be aware, that we have a new API from CozyClient to persist queries. See https://github.com/cozy/cozy-notes/pull/327 for how to handle it. (I know this is not what you need for now, but maybe it can provide you some guidance / vision)

Ldoppea:
Latest decision:
- We won't merge the store with the cozy-client one for now. But we expect to do it in a later PR

This PR is ready for review

> **Warning**
> Do not review the logic inside of `sendConnectorsLogs()` and `useConnectors > doAddLog()`, those are demo codes and are expected to be replaced in #534

src/redux/store.ts
```diff
@@ -0,0 +1,42 @@
...
+})
+
+const persistedReducer = persistReducer(persistConfig, rootReducter)
+
+export const store = configureStore({
+  reducer: persistedReducer,
+  middleware: getDefaultMiddleware =>
+    getDefaultMiddleware({
+      serializableCheck: {
+        ignoredActions: [FLUSH, REHYDRATE, PAUSE, PERSIST, PURGE, REGISTER]
```
Crash--, line 33: I still don't understand this need. Is it because of TS?

Crash-- (approved):

src/redux/store.ts
Crash--: OK next time I'll read the commit message https://github.com/cozy/cozy-flagship-app/pull/516/commits/d11100b0df56e6108d920ab7d49ec39741514ee2 

So now my question is: why do we use redux-tookit?

Ldoppea: Woups I missed this notification.

I used it because it is the official recommendation: https://redux.js.org/introduction/getting-started

PR #524, opened 2022-12-26
fix: Set correct caret color on LockScreen on iOS

Ldoppea:
The `cursorColor` prop only works on Android so we want to use `selectionColor` on iOS instead

The drawback of `selectionColor` is that the caret AND the selection highlight are impacted. This is not a problem on iOS as an opacity is applied on the highlight color. However on Android we get white text on white highlight, which is not usable. So we want to apply `selectionColor` only on iOS

More info:
- https://reactnative.dev/docs/textinput#cursorcolor-android
- https://reactnative.dev/docs/textinput#selectioncolor

### Result on iOS
image
image

### Result on Android
image
image

### What would happen on Android if we applied the `selectionColor` prop
image

> **Note**
> We may want to use Secondary color on Android for the selection color. This is what we do in CozyPass Mobile (cf cozy/cozy-pass-mobile@d9b7dbc145ab0622ccc4bc357c228cd4758ad2fb). But the react-native API only changes the highlight color and not the selection handle. So we will have to investigate a better solution.
image

### DIFF
--- a/src/ui/TextField/index.tsx
+++ b/src/ui/TextField/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { TextInput, TextInputProps, View } from 'react-native'
+import { Platform, TextInput, TextInputProps, View } from 'react-native'
 import { MaskInputProps } from 'react-native-mask-input'
 
 import { Typography } from '/ui/Typography'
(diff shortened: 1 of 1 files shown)


acezard (approved):

acezard:
👍

PR #525, opened 2022-12-27
Feat  store and update fetched connectors

acezard:
We want to put the CCCs into production. It will therefore be required to no longer put the connectors in the flagship application but in the registry, then retrieve them at the right time.

This PR consists of retrieving the sources of the connector during its installation, then during its execution.

* [x] Remove old code
* [x] Install a new connector
* [x] Update an existing connector
* [x] Fetch a connector from the file system
* [x] Clear old connector versions from the file system


src/libs/ReactNativeLauncher.js
```diff
@@ -59,14 +60,34 @@ class ReactNativeLauncher extends Launcher {
...
+
+    try {
+      bundle = await getConnectorBundle({ client, slug })
+    } catch (err) {
+      log.error('Error while getting connector bundle from cache', err)
+    }
+
+    if (!bundle)
+      throw new Error(
+        `Error while reading connector from cache, falsy value received ${slug} connector`
```
acezard, line 87: Message not great

src/libs/ReactNativeLauncher.js
```diff
@@ -3,9 +3,10 @@ import Minilog from '@cozy/minilog'
 
 import ContentScriptBridge from './bridge/ContentScriptBridge'
 import Launcher from './Launcher'
-import * as ConnectorInstaller from './ConnectorInstaller'
 import { saveCookie, getCookie, removeCookie } from './keychain'
 import CookieManager from '@react-native-cookies/cookies'
+import { updateCozyAppBundle } from './cozyAppBundle/cozyAppBundle'
```
acezard, line 8: relative imports

src/libs/cozyAppBundle/cozyAppBundle.spec.ts
```diff
@@ -18,11 +19,8 @@ import {
 } from './cozyAppBundleConfiguration'
 
 jest.mock('react-native-fs', () => ({
-  DocumentDirectoryPath: 'SOME_DocumentDirectoryPath',
-  downloadFile: jest.fn(),
-  exists: jest.fn(),
-  mkdir: jest.fn(),
-  unlink: jest.fn()
+  ...jest.requireActual('../../../__tests__/__mocks__/react-native-fs'),
```
acezard, line 22: requireMock ?

src/libs/cozyAppBundle/cozyAppBundle.ts
```diff
@@ -6,64 +6,91 @@ import {
...
+import { handleCleanup } from '/libs/cozyAppBundle/functions'
 
-const log = logger('AppBundle')
+export const log = logger('AppBundle')
 
 const BUNDLE_UPDATE_DELAY_IN_MS = 10000
 
+type AppType = 'konnectors' | 'apps'
+
+export interface AppInfo {
```
acezard, line 25: All those type declarations look bad

src/libs/cozyAppBundle/cozyAppBundle.ts
```diff
@@ -6,64 +6,91 @@ import {
...
-      return await updateCozyAppBundle({ slug, client })
-    } catch (err) {
-      log.error(
-        `Something went wront while updating ${slug} bundle: ${err.message}`
-      )
-    }
-  }, delayInMs)
-}
+}: CozyAppToUpdate): Promise<void> =>
+  new Promise(resolve => {
```
acezard, line 59: doesn't look good

src/libs/cozyAppBundle/cozyAppBundle.ts
```diff
@@ -6,64 +6,91 @@ import {
...
+              err
+            )}`,
+            resolve()
+          )
+        )
+    }, delayInMs)
+  })
 
 /**
  * Check the cozy-app version on cozy-stack and update the local CozyAppBundle
```
acezard, line 75: Update comments to reflect added connector feature

src/libs/cozyAppBundle/cozyAppBundle.ts
```diff
@@ -94,26 +121,36 @@ export const updateCozyAppBundle = async ({ slug, client }) => {
...
     fqdn,
     slug,
     version: stackVersion,
     folder: normalizeVersion(stackVersion) + tarPrefix
   })
+
+  await handleCleanup({
+    slug,
+    client,
+    versionsToKeep: [
```
acezard, line 138: make it declarative

src/libs/cozyAppBundle/cozyAppBundle.ts
```diff
@@ -129,23 +166,31 @@ const deleteVersionBundleFromLocalFilesIfExists = async ({
   }
 }
 
-const normalizeVersion = version => {
+const normalizeVersion = (version: string): string => {
```
acezard, line 169: it does nothing, remove it

src/libs/cozyAppBundle/cozyAppBundle.ts
```diff
@@ -169,60 +215,66 @@ const downloadAndExtractCozyAppVersion = async ({
...
   slug,
   version,
   destinationPath,
-  client
-}) => {
-  const stackClient = client.getStackClient()
+  client,
+  type
+}: CozyAppWithDest): Promise<void> => {
+  const stackClient = client.getStackClient() as {
```
acezard, line 251: find a better way to cast cozyclient

src/screens/connectors/LauncherView.js
```diff
@@ -33,22 +18,6 @@ const colors = getColors()
 
 const DEBUG = false
 
-const embeddedConnectors = {
```
acezard, line 36: comment out

src/libs/ReactNativeLauncher.js
```diff
@@ -59,14 +60,34 @@ class ReactNativeLauncher extends Launcher {
...
 
-  async ensureConnectorIsInstalled({ connector, client }) {
-    await ConnectorInstaller.ensureConnectorIsInstalled({
-      ...connector,
-      client
-    })
-    const content = await ConnectorInstaller.getContentScript(connector)
-    const manifest = await ConnectorInstaller.getManifest(connector)
-    return { content, manifest }
+  async ensureConnectorIsInstalled({ connector: { slug }, client }) {
```
acezard, line 63: connector must be a connector string slug

src/libs/cozyAppBundle/cozyAppBundle.ts
```diff
@@ -0,0 +1,268 @@
...
+  destinationPath,
+  slug,
+  type,
+  version
+}: AppInfo & {
+  destinationPath: string
+  type: AppType
+  version: string
+}): Promise<void> => {
+  const stackClient = client.getStackClient() as {
```
Ldoppea, line 239: Can we use inline type here, as in https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/intents/localMethods.ts#L59 ?

__mocks__/react-native-fs-mock.js
```diff
@@ -0,0 +1,45 @@
+const rnfsMock = {
+  CachesDirectoryPath: jest.fn(),
+  DocumentDirectoryPath: jest.fn(),
+  ExternalDirectoryPath: jest.fn(),
+  ExternalStorageDirectoryPath: jest.fn(),
+  LibraryDirectoryPath: jest.fn(),
+  MainBundlePath: jest.fn(),
+  PicturesDirectoryPath: jest.fn(),
+  TemporaryDirectoryPath: jest.fn(),
```
Ldoppea, line 9: `XXXPath` members are not functions. They are just strings.

src/libs/cozyAppBundle/cozyAppBundle.ts
```diff
@@ -0,0 +1,268 @@
...
+  slug: string
+}
+
+/**
+ * After the specified delay, check the cozy-app version on cozy-stack and
+ * update the local CozyAppBundle if necessary
+ *
+ * This method should run in background and cannot be awaited
+ */
+export const updateCozyAppBundleInBackground = ({
```
Ldoppea, line 39: ⚠️  As you removed the JSDoc, we lost the mention to default value and so it won't appear in the autocomplete documentation.
This is not really a problem as this is an internal method, but we should find a way to document TS code correctly in the future.

src/libs/ReactNativeLauncher.js
```diff
@@ -59,14 +60,34 @@ class ReactNativeLauncher extends Launcher {
...
-    })
-    const content = await ConnectorInstaller.getContentScript(connector)
-    const manifest = await ConnectorInstaller.getManifest(connector)
-    return { content, manifest }
+  async ensureConnectorIsInstalled({ slug, client }) {
+    try {
+      await updateCozyAppBundle({
+        slug,
+        client,
+        type: 'konnectors'
```
Ldoppea, line 68: nit: as this is type enforced, you can also declare `BUNDLE_TYPE_KONNECTOR` and `BUNDLE_TYPE_APP` consts in `cozyAppBundle.ts`

src/screens/connectors/LauncherView.js
```diff
@@ -73,20 +42,18 @@ class LauncherView extends Component {
 
   async initConnector() {
     const { client, launcherContext } = this.props
-    let result = null
-    let connector = embeddedConnectors[launcherContext.job.message.konnector]
```
Ldoppea, line 77: Didn't we conclude that we should keep the "local" connectors mecanism until we are able to easily declare them through a local cozy-stack (compatible with `build --watch`)?

__mocks__/react-native-fs-mock.js
acezard: Updated mock file with typings, missing methods, and mock primitive values with their intended values instead of mock fn

src/libs/cozyAppBundle/cozyAppBundle.ts
acezard: Improved cozy client typings app-wide

src/libs/cozyAppBundle/cozyAppBundle.ts
acezard: Agree

src/libs/ReactNativeLauncher.js
acezard: I'm not too sold as there are only two values, typescript will forbid you to use anything else than what is expected. Now, this file is in js, but since I'm working on it believe me it won't stay like that for long 😂

src/screens/connectors/LauncherView.js
acezard: Honestly there is already an easy method to test them, it works well https://zoe-notes.mycozy.cloud/public/?sharecode=xmKI8NByNkbP

For real time usage, I can implement something but I need to see with connector team how they want to handle that.

Ldoppea: The real time scenario seems to be critical in my point of view. The connector team works with this feature on a daily basis.

Let @KillianCourvoisier, @LucsT and @doubleface confirm that we can merge this as they will be the first ones impacted.

Ldoppea (approved):

PR #533, opened 2023-01-05
feat: implement cozyprovider for LauncherView

acezard:
We create a new CozyClient each time a slug changes or each time
the LauncherView is remounted.

Example of a JWT token as a bearer in request header sent from LauncherView:

```
{
  "aud": "konn",
  "iat": 1673344160,
  "iss": "anc.10-0-2-2.nip.io:8080",
  "sub": "template"
}
```


acezard:
cozy-client must be updated or it won't work (using a new oauth.method)

src/libs/connectorLibs/getLauncherClient.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Gets the client to be used by the launcher.
+ *
+ * The launcher is the webview that displays the login form for a given connector.
+ * This function creates a new client with the right token for the connector.
+ *
+ * @param {CozyClient} client - The main client
+ * @param {string} slug - The connector slug
+ * @param {React.Dispatch<React.SetStateAction<CozyClient | undefined>>} callback - A callback invoked with the new client
```
acezard, line 16: Should be react agnostic

src/libs/connectorLibs/getLauncherClient.ts
```diff
@@ -0,0 +1,56 @@
...
+      uri
+    })
+
+    callback?.(client)
+
+    return client
+  } catch (error) {
+    throw new Error(
+      `Failed to create launcher client for ${slug}.\n,
+      ${getErrorMessage(error)}`
```
Ldoppea, line 38: nit: I'm wondering if we should start using the `cause` param to pass the original error. But i don't know if that would give more context in Sentry (cf [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause))

src/libs/connectorLibs/getLauncherClient.ts
```diff
@@ -0,0 +1,56 @@
...
+    )
+  }
+}
+
+export const useLauncherClient = (
```
Ldoppea, line 43: Should we put this entire file in the `hook` folder instead? The `connectorLibs` should be for connectors `contentScript` (if I understand it correctly)

src/libs/connectorLibs/getLauncherClient.ts
acezard: I'm adding it

src/libs/connectorLibs/getLauncherClient.ts
acezard: Agreed

src/libs/connectorLibs/getLauncherClient.ts
```diff
@@ -0,0 +1,56 @@
...
+  launcherContext: LauncherContext
+): { launcherClient?: CozyClient } => {
+  const [launcherClient, setLauncherClient] = useState<CozyClient>()
+  const client = useClient()
+
+  useEffect(() => {
+    const slug = launcherContext.job?.message?.konnector
+
+    if (slug) void getLauncherClient(client, slug, setLauncherClient)
```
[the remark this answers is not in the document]
acezard: With this:
```
  useEffect(() => {
    const slug = launcherContext.job?.message?.konnector

    if (slug) void getLauncherClient(client, slug, setLauncherClient)
  }, [client, launcherContext.state, launcherContext.job?.message?.konnector])
```

I'm not sure that launcherContext.state is safe indeed. I'm going to only check the connector slug, if it ever changes we should be good to create a new client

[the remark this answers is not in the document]
Ldoppea: I just found another possible bug: after the first `setLauncherClient(client)` call, the `launcherClient` will be `undefined` anymore.

So `isClientReady(launcherClient)` will always be `true` after the first connector run. Then if we run another connector we won't have any control on the `launcherClient` value. `LauncherView` would then start to use the old client and the switch to the correct one during is lifecycle.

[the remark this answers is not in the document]
acezard: Was seen in peer review ✔️

Ldoppea (approved):

PR #534, opened 2023-01-06
feat: Sending konnectors logs to stack in batch

LucsT:
This PR send konnectors logs in batch to the stack.

They respect the format of stack.
They are filtered on possible level (or debug is the default level) : debug, info, warn, error
They are log into the flagship app too corresponding to their level.

### DIFF
--- a/connectors/connectorLibs/ContentScript.js
+++ b/connectors/connectorLibs/ContentScript.js
@@ -362,11 +362,26 @@ export default class ContentScript {
   /**
    * Send log message to the launcher
    *
-   * @param {string} : the log message
-   * @todo Use cozy-logger to add logging level and other features
+   * @param {"debug"|"info"|"warn"|"error"} level: the log level
+   * @param {string} message: the log message
    */
-  log(message) {
-    this.bridge.emit('log', message)
+  log(level, message) {
+    const allowedLevels = ['debug', 'info', 'warn', 'error']
+    if (!message) {
+      log.warn(
+        `you are calling log without message, use log(level,message) instead`
+      )
+      return
+    }
+    if (!allowedLevels.includes(level)) {
+      level = 'debug'
+    }
+    const now = new Date()
+    this.bridge.emit('log', {
+      timestamp: now,
+      level,
+      msg: message
+    })
   }
 
   /**
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -104,7 +104,7 @@ class ReactNativeLauncher extends Launcher {
   async stop({ message } = {}) {
     const context = this.getStartContext()
     const client = context.client
-    sendConnectorsLogs(client)
+    await sendConnectorsLogs(client)
     if (message) {
       await this.updateJobResult({
         state: 'errored',
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -7,6 +7,7 @@ import Launcher from './Launcher'
 import { getConnectorBundle } from '/libs/cozyAppBundle/cozyAppBundle.functions'
 import { saveCookie, getCookie, removeCookie } from './keychain'
 import { updateCozyAppBundle } from '/libs/cozyAppBundle/cozyAppBundle'
+import { sendConnectorsLogs } from '/libs/connectors/sendConnectorsLogs'
 
 const log = Minilog('ReactNativeLauncher')
 
@@ -101,6 +102,9 @@ class ReactNativeLauncher extends Launcher {
   }
 
   async stop({ message } = {}) {
+    const context = this.getStartContext()
+    const client = context.client
+    sendConnectorsLogs(client)
     if (message) {
       await this.updateJobResult({
         state: 'errored',
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -27,6 +27,16 @@ class ReactNativeLauncher extends Launcher {
     this.workerListenedEventsNames = ['log', 'workerEvent', 'workerReady']
   }
 
+  setLogger(onKonnectorLog) {
+    this.logger = onKonnectorLog
+  }
+
+  log(logContent) {
+    const context = this.getStartContext()
+    const slug = context.manifest.slug
+    this.logger({ ...logContent, slug })
+  }
+
   async init({ bridgeOptions, contentScript }) {
     log.debug('bridges init start')
     const promises = [
--- a/src/libs/connectors/sendConnectorsLogs.spec.js
+++ b/src/libs/connectors/sendConnectorsLogs.spec.js
@@ -1,0 +1,177 @@
+import CozyClient from 'cozy-client'
+import { sendConnectorsLogs } from './sendConnectorsLogs'
+import { store } from '/redux/store'
+
+jest.genMockFromModule('cozy-client')
+jest.mock('cozy-client')
+
+const mockFetchJSON = jest.fn()
+
+const mockCozyClient = {
+  getStackClient: jest.fn().mockReturnValue({ fetchJSON: mockFetchJSON })
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+CozyClient.mockImplementation(() => mockCozyClient)
+
+jest.mock('/redux/store', () => ({
+  store: {
+    getState: jest.fn(),
+    dispatch: jest.fn()
+  }
+}))
+
+describe('Konnectors logs', () => {
+  it('Should send no log', async () => {
+    const client = new CozyClient()
+
+    store.getState.mockReturnValue({
+      connectorLogs: {
+        logs: {}
+      }
+    })
+
+    await sendConnectorsLogs(client)
+    expect(mockFetchJSON).not.toHaveBeenCalled()
+    expect(store.dispatch).not.toHaveBeenCalled()
+  })
+
+  it('Should send one log', async () => {
+    const client = new CozyClient()
+
+    store.getState.mockReturnValue({
+      connectorLogs: {
+        logs: {
+          myslug: [
+            {
+              level: 'debug',
+              msg: 'Log Message',
+              timestamp: '2023-01-05T10:12:32Z'
+            }
+          ]
+        }
+      }
+    })
+
+    await sendConnectorsLogs(client)
+    expect(mockFetchJSON).toHaveBeenCalledWith(
+      'POST',
+      '/konnectors/myslug/logs',
+      [
+        {
+          level: 'debug',
+          msg: 'Log Message',
+          timestamp: '2023-01-05T10:12:32Z'
+        }
+      ]
+    )
+    expect(store.dispatch).toHaveBeenCalledWith({
+      payload: { number: 1, slug: 'myslug' },
+      type: 'connectorLogs/removeLogs'
+    })
+  })
+
+  it('Should send 2 logs', async () => {
+    const client = new CozyClient()
+
+    store.getState.mockReturnValue({
+      connectorLogs: {
+        logs: {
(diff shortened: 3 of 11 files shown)


Crash--:
@LucsT you should rebase and fix lint issues 🙏

src/libs/ReactNativeLauncher.js
```diff
@@ -26,6 +26,16 @@ class ReactNativeLauncher extends Launcher {
     this.workerListenedEventsNames = ['log', 'workerEvent', 'workerReady']
   }
 
+  setLogger(onKonnectorLog) {
+    this.logger = onKonnectorLog
+  }
+
+  async log(logContent) {
```
Crash--, line 33: why this method is async ? Since there is no away in it?

src/providers/ConnectorState/ConnectorLogsSlice.ts
```diff
@@ -2,31 +2,55 @@ import { createSlice, PayloadAction } from '@reduxjs/toolkit'
...
-      state.logs = [...state.logs, action.payload]
+    addLog: (state, action: PayloadAction<LogObj>) => {
+      if (state.logs[action.payload.slug] === undefined) {
+        state.logs[action.payload.slug] = []
+      }
+      state.logs[action.payload.slug]?.push(action.payload)
     },
-    spliceLogs: (state, action: PayloadAction<number>) => {
-      state.logs = state.logs.slice(action.payload)
+    removeLogs: (state, action: PayloadAction<removeLogInfo>) => {
```
Crash--, line 37: We have `addLog` in singular but `removeLogs` in plural. Is it wanted?

src/providers/ConnectorState/ConnectorLogsSlice.ts
```diff
@@ -2,31 +2,55 @@ import { createSlice, PayloadAction } from '@reduxjs/toolkit'
...
+      const result = state.logs[action.payload.slug]?.slice(
+        action.payload.number
+      )
+      if (result?.length) {
+        state.logs[action.payload.slug] = result
+      } else {
+        delete state.logs[action.payload.slug]
```
Crash--, line 44: delete is extremely slow in javascript. Setting undefined is way better. 

You can find a lot of readings about that on the internet, here is one of the first link I found https://webkit.org/blog/10298/inline-caching-delete/

src/providers/ConnectorState/LogLoop.ts
```diff
@@ -1,12 +1,16 @@
...
+import { removeLogs } from '/providers/ConnectorState/ConnectorLogsSlice'
 import { store } from '/store'
+import { default as CozyClientType } from 'cozy-client'
+
+interface CozyClient extends CozyClientType {
+  getStackClient: () => {
+    uri: string
+    fetchJSON: <T>(method: string, path: string, body: unknown) => Promise<T>
+  }
+}
```
Crash--, line 13: Should we create this interface elsewhere in the project in order to be able to reuse it easily? Don't we already have a place to store these interfaces? @acezard @Ldoppea

src/providers/ConnectorState/LogLoop.ts
```diff
@@ -24,31 +28,34 @@ export class LogLoop {
...
+        try {
+          // Sending
+          await client
+            .getStackClient()
+            .fetchJSON('POST', `/konnectors/${slug}/logs`, cleanedLogs)
+          // Deleting
+          store.dispatch(removeLogs({ slug, number: cleanedLogs.length }))
+        } catch (e) {
+          log.error(`Error while sending ${slug} logs`)
+          log.error(e)
```
Crash--, line 56: Is it ok to log the error object directly? What will happen if the reason is itself an object?

src/providers/ConnectorState/LogLoop.ts
```diff
@@ -24,31 +28,34 @@ export class LogLoop {
...
+        }
+      }
     } catch (e) {
       log.error(e)
```
Crash--, line 60: You should definitively add a string before to tell us more about this log.

src/screens/connectors/LauncherView.js
```diff
@@ -18,7 +18,7 @@ import { WebView } from 'react-native-webview'
 // import redConnector from '../../../connectors/red/dist/webviewScript'
 // import sfrConnector from '../../../connectors/sfr/dist/webviewScript'
 // import totalenergiesConnector from '../../../connectors/totalenergies/dist/webviewScript'
-// import alanConnector from '../../../connectors/alan/dist/webviewScript'
+import alanConnector from '../../../connectors/alan/dist/webviewScript'
```
Crash--, line 21: to remove

src/libs/ReactNativeLauncher.js
LucsT: Shouldn't !

src/screens/home/HomeScreen.js
```diff
@@ -1,17 +1,26 @@
...
 export const HomeScreen = ({ route, navigation }) => {
   const [barStyle, setBarStyle] = useState(StatusBarStyle.Light)
   const [debug] = useState(false)
   const [launcherContext, setLauncherContext] = useState({ state: 'default' })
+  const { addLog } = useConnectorState()
 
+  const onKonnectorLog = async logObj => {
+    konnLog[logObj.level](`${logObj.slug}: ${logObj.msg}`)
+    addLog(logObj)
+  }
```
Crash--, line 23: It seems weird to me, that this is the HomeScreen job to `addLog`. Shouldn't be the responsibility of the `LauncherView`? @Ldoppea

src/libs/ReactNativeLauncher.js
LucsT: Done

src/providers/ConnectorState/ConnectorLogsSlice.ts
LucsT: Should stay like this.

src/libs/ReactNativeLauncher.js
```diff
@@ -26,6 +26,16 @@ class ReactNativeLauncher extends Launcher {
...
+    const context = this.getStartContext()
+    const slug = context.manifest.slug
+    this.logger({ ...logContent, slug })
```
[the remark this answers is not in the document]
LucsT: Passing 2 arguments in a action payload is impossible.
```addLog: (state, action: PayloadAction<LogObj>)```

We have to structurate this object in order to transmit the message/level and the slug

File: connectors/connectorLibs/ContentScript.js [hunk 10]
```diff
@@ -361,11 +362,25 @@ export default class ContentScript {
...
-   * @todo Use cozy-logger to add logging level and other features
    */
-  log(message) {
-    this.bridge.emit('log', message)
+  log(level, message) {
+    const allowedLevels = ['debug', 'info', 'warn', 'error']
+    if (message === undefined) {
+      message = level
+      level = 'debug'
+    }
```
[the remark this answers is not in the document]
LucsT: We discuss about it. And we choose log(level, message) for two reasons :
 - It matches not client side connectors
 - It seems more logic and readable to have the level in first argument
 
 What should I do ?
 Remove the magic adding debug level, and we change it in the few connectors ?

src/providers/ConnectorState/LogLoop.ts
LucsT: Done

src/providers/ConnectorState/LogLoop.ts
LucsT: It seems better than nothing. What do you recommend instead?

connectors/connectorLibs/ContentScript.js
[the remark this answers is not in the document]
Crash--: >  Remove the magic adding debug level, and we change it in the few connectors ?

Yep

src/screens/home/HomeScreen.js
Ldoppea: This is because we chose to handle logs by the Flagship part (vs Launcher) so we can use the same code to send them on app's restart. Due to this we can use only the Flagship route (the one that does not require the connector's dedicated token).

Also the connector doesn't have to know what is done with the logs. They can be send instantaneously, They can be sent in batch, or not sent due to an error and then sent on next app's start.

So we considered that the responsibility split is more clear if the Launcher only know the `log()` interface and then the Flagship part does all the batch/persist logic.

connectors/connectorLibs/ContentScript.js
```diff
@@ -4,6 +4,7 @@ import Minilog from '@cozy/minilog'
 
 import LauncherBridge from './bridge/LauncherBridge'
 import { kyScraper as ky, blobToBase64 } from './utils'
+import { debug } from 'post-me'
```
Crash--, line 7: Why this import? Is it used?

src/libs/connectors/sendConnectorsLogs.spec.js
```diff
@@ -0,0 +1,177 @@
+import CozyClient from 'cozy-client'
```
acezard, line 1: This whole file should be in .ts (new file + tested file is written in ts)

src/screens/home/HomeScreen.tsx
```diff
@@ -25,6 +30,16 @@ export const HomeScreen = ({
     state: string
     value?: Record<string, unknown>
   } | null>({ state: 'default' })
+  const { addLog } = useConnectors()
+
+  const onKonnectorLog = (logObj: LogObj): void => {
```
acezard, line 35: this logic should be written somewhere else, View files should strive to be simple

src/libs/connectors/sendConnectorsLogs.spec.js
LucsT: As discussed we @Ldoppea and you, we will keep it in JS, too much difficulty to made a clean cozy-client mock in .ts

connectors/connectorLibs/ContentScript.js
```diff
@@ -361,11 +361,26 @@ export default class ContentScript {
   /**
    * Send log message to the launcher
    *
+   * @param {string} : the log level
    * @param {string} : the log message
-   * @todo Use cozy-logger to add logging level and other features
    */
-  log(message) {
-    this.bridge.emit('log', message)
+  log(level, message) {
```
doubleface, line 367: We will need to port this to cozy-ccc-libs

connectors/connectorLibs/ContentScript.js
```diff
@@ -361,11 +361,26 @@ export default class ContentScript {
   /**
    * Send log message to the launcher
    *
+   * @param {string} : the log level
```
Crash--, line 364: * @param {"debug"|"info"|"warn"|"error"} level: the log level

LucsT: Updated

connectors/connectorLibs/ContentScript.js
LucsT: PR done

src/libs/ReactNativeLauncher.js
```diff
@@ -91,6 +102,9 @@ class ReactNativeLauncher extends Launcher {
   }
 
   async stop({ message } = {}) {
+    const context = this.getStartContext()
+    const client = context.client
+    sendConnectorsLogs(client)
```
Ldoppea, line 107: Should we add an `await` here?

Ldoppea (approved):

src/screens/home/HomeScreen.tsx
LucsT: Done

src/libs/ReactNativeLauncher.js
LucsT: Done

src/screens/home/HomeScreen.js
Crash--: @Ldoppea WDYT about my previous comment?

PR #535, opened 2023-01-06
refactor: clean up src/ directory

acezard:
index.js will be the only file in the root directory.
some directories were deleted (providers, config) and their files relocated.
all the pathes were then updated.

this should be squashed into one commit when merging


Ldoppea:
@acezard : Seems like your commits do not contain everything. I see that some files are created but with no related deletion elsewhere.

acezard:
> @acezard : Seems like your commits do not contain everything. I see that some files are created but with no related deletion elsewhere.

Not sure, on my end everything should be committed. api-keys.json is ignored though.

Ldoppea:
As an example I see the new dev-config.ts, but it is considered as new file, and I don't see any reference to old dev.ts

acezard:
> As an example I see the new dev-config.ts, but it is considered as new file, and I don't see any reference to old dev.ts

it seems to be the only file isn't it? it must be a problem of assume unchanged on my end

Ldoppea:
`App.js` is more an entrypoint than a "screen" in my point of view.

But more important, this is an "origin" file from react-native's bootstrap project. So moving it would make difficult for us to upgrade react-native in the future using the upgrade tool.
This is also the case for `app.json`

acezard:
> `App.js` is more an entrypoint than a "screen" in my point of view.
> 
> But more important, this is an "origin" file from react-native's bootstrap project. So moving it would make difficult for us to upgrade react-native in the future using the upgrade tool. This is also the case for `app.json`

the same applies for index.js?
but anyway that's a good argument when updating react-native. I can move them back. Any other change causing issues?

Ldoppea:
> the same applies for index.js?

Yes, but this is the entry point, so the app would not even run if you move it without also editing the build configuration

> Any other change causing issues?

Everything else seems ok (or I missed it)

You can see an untouched version of rn 0.66 here: https://github.com/Ldoppea/react-native-bug-reports

src/constants/dev-config.ts
```diff
@@ -1,7 +1,7 @@
 export const devConfig = {
   disableGetIndex: false, // Make cozy-app hosted by webpack-dev-server work with HTTPServer
   forceHideSplashScreen: false, // Hide react-native splash screen renders in development mode,
-  ignoreLogBox: false, // Hide react-native logbox renders in development mode,
```
Ldoppea, line 4: you should revert this line

Ldoppea (approved):

PR #539, opened 2023-01-09
fix: Template connector manifest url to https

doubleface:
Or else, the pilot webview will be init on http://books.toscrape.com and
will fail to fetch https files.

#### Checklist

* [X] Tested on Android


LucsT:
Oui on fait du fetch. via saveFiles (mais ce n'est pas fetch de node, c'est la lib 'ky')

PR #543, opened 2023-01-12
feat: use `CozyFonts` app attribute from the stack

acezard:
No implementation needed as it is handled declarative way


src/libs/httpserver/indexDataFetcher.ts
```diff
@@ -15,38 +30,29 @@ export const fetchAppDataForSlug = async (slug, client) => {
...
-      subdomain: cozyDataAttributes.SubDomain,
+      subdomain: cozyDataAttributes.SubDomain as 'nested' | 'flat',
       token: cozyDataAttributes.Token,
       tracking: cozyDataAttributes.Tracking
     }
 
     let cozyDataString = JSON.stringify(cozyData)
     cozyDataString = replaceAll(cozyDataString, '"', '&#34;')
 
-    cozyDataAttributes.Flags = replaceAll(
```
acezard, line 38: I remember this hack and removed it, somehow I didn't see any issue in the produced HTML

src/libs/httpserver/indexDataFetcher.ts
```diff
@@ -1,12 +1,27 @@
...
+import { replaceAll } from '/libs/functions/stringHelpers'
+
+interface FetchAppDataForSlugResult {
+  cookie: string
+  templateValues: {
+    [P in keyof Partial<Omit<AppAttributes, 'CozyData'>>]: string
+  } & {
+    CozyData: string
+  }
+}
```
Ldoppea, line 16: This type seems from a mix between an old type organisation and a new one.

Here you Omit `CozyData` from `AppAttributes` but `AppAttributes` does not contain this attribute
And then you add `CozyData` to the final type again. This seems to cancel the Omitted one.

src/libs/httpserver/indexDataFetcher.ts
Ldoppea: Are you sure this isn't needed anymore? 🤔

src/libs/httpserver/indexDataFetcher.ts
acezard: Fixed but CozyData isn't in AppAttributes though

src/libs/httpserver/indexDataFetcher.ts
acezard: Yeah it seems so, because the stack changed the way strings are formatted. I couldn't reproduce any issue with removing all the formatting we did before

Crash--: > Yeah it seems so, because the stack changed the way strings are formatted. 

This is not something I'm aware of. @nono, does the stack changed something about cozy-data.flags encoding?

acezard: > > Yeah it seems so, because the stack changed the way strings are formatted.
> 
> This is not something I'm aware of. @nono, does the stack changed something about cozy-data.flags encoding?

I seemed certain about that but it's just an observation. Seems like what we get from the stack is escaped correctly and we do not need manipulation like before. Or maybe it was like that from the start? Or maybe I'm wrong, but from my end, the cozy-data and the other attributes are all correctly injected into this PR

nono: I don't think the stack has changed something about that. Maybe, when upgrading the go version used to compile the code, a change in the stdlib has an effect on that.

Ldoppea (approved):

PR #544, opened 2023-01-16
feat: obfuscate phone numbers in Sentry logs

acezard:
We do not want to see PII data in Sentry logs, we just delete anything detectable as a phone number in an error message.


src/libs/monitoring/scrubbing.ts
```diff
@@ -0,0 +1,12 @@
+import Sentry from '@sentry/react-native'
+
+import { devlog } from '/libs/services/EnvService'
+
+export const scrubPhoneNumbers = (event: Sentry.Event): Sentry.Event => {
+  if (event.message) {
+    event.message = event.message.replace(/tel:[^ ]*/g, 'tel:XXXXXXXXXX')
```
Ldoppea: What if the tel is in the form `+33 7 12345678`/`07 12 34 56 78`? Or in the US format `(234)-555-1234`?

src/libs/services/EnvService.ts
```diff
@@ -1,37 +1,56 @@
...
 
 if (devConfig.sentry) toggleLocalSentry(true)
 
 const hasSentryEnabled = enableSentryOn.some(
   environment => environment === name
 )
 
+export const isDev = (): boolean => nameIs(strings.environments.test)
+
+export const devlog = isDev()
```
Ldoppea: nit: can you edit the commit's message to say that you added this method? Or even better extract this in another commit

Ldoppea:
This PR is about preventing Sentry to send phone numbers, but did you check why the original error occurred?

src/libs/monitoring/scrubbing.ts
acezard: I don't think it matters since we replace everything after "tel:"

acezard: @Ldoppea I modified the scrubbing so we can have outputs like `tel:(***)-***-****` now

Ldoppea (approved):

PR #547, opened 2023-01-19
feat: add Android development CI

acezard:
### What this workflow does

- Build a signed APK
- Expose the APK in the job artefact (as `app-x86_64-release-signed.apk` in a `app-release.zip`)
- Provide a way to manually trigger it (once this branch has been merged only)

### What this PR changes to the application

- Some secrets will have to be provided by the developer from their .env (like API keys)
- The local build will always be unsigned by default
- **Sentry** default properties are versioned (as suggested by **Sentry** itself)
- **Cozy-home** will always be installed after a `yarn install`

### Actions used

**From GitHub:**

- [actions/checkout@v2](https://github.com/actions/checkout)
- [actions/setup-java@v1.4.3](https://github.com/actions/setup-java)
- [actions/cache@v2](https://github.com/actions/cache)
- [actions/upload-artifact@v3](https://github.com/actions/upload-artifact)

**Third-party:**

- [r0adkll/sign-android-release@v1](https://github.com/r0adkll/sign-android-release)

This action is depreciated and unmaintained at this time, but still works for now. We can observe that in the Node12 usage for instance. It should be advised to fork the action and adapt it to our needs and update everything to current standards (no need to write one from scratch).

### Mandatory environment secrets:

- `ANDROID_ALIAS`
- `ANDROID_KEY_PASSWORD`
- `ANDROID_KEY_STORE_PASSWORD`
- `ANDROID_SAFETY_NET_API_KEY`
- `ANDROID_SIGNING_KEY`
- `SENTRY_AUTH_TOKEN`

### How was it tested

- Initially used to trigger workflow on push to iterate rapidly
- Installed the produced APK on a physical device

### Suggestions

- Automate artefact upload to a remote location rather than manually download it
- Add a release system linked to GitHub releases to trigger builds


android/app/build.gradle
```diff
@@ -154,14 +154,6 @@ android {
             keyAlias 'androiddebugkey'
             keyPassword 'android'
         }
-        release {
-            if (project.hasProperty('MYAPP_UPLOAD_STORE_FILE')) {
```
Ldoppea, line 158: As you removed this config, we should edit the build process documentation, as we may still want to manually build a signed APK locally in some scenario

src/constants/api-keys.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  androidSafetyNetApiKey:
+    process.env['ANDROID_SAFETY_NET_API_KEY'] ?? 'testEnvFallback'
```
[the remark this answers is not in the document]
acezard: Totally agree, I will add a .env file with a readme in another PR if that suits you

android/app/build.gradle
acezard: True. I will check documentation because I'm unsure how to do it at the moment, then add it to the readme once I managed to do it locally

src/constants/api-keys.js
Ldoppea, line 3: Did you test this on runtime?

When reading the CI files, it seems like you are defining `ANDROID_SAFETY_NET_API_KEY` env variable for the build process (`Build Android App Bundle` task), but are you sure this will be set also for the runtime process?

android/sentry.properties
```diff
@@ -0,0 +1,3 @@
+defaults.url=https://errors.cozycloud.cc/
+defaults.org=cozycloud
+defaults.project=app-amirale
```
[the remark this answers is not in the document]
acezard: I think these infos are pretty safe, it's easier to use sentry locally if we keep them. And yes the example become useless

src/constants/api-keys.js
acezard: I figured we only needed it at build time, I shall check if the variable is injected in runtime

Ldoppea: `ANDROID_SAFETY_NET_API_KEY` is used on runtime during the login process. The mobile should do the flagship certification and call Google APIs.

acezard: using react-native-config to inject the secret into runtime as well as using it directly during build time

android/app/build.gradle
acezard: Actually, I would like to update the whole readme file, I think some other info has to be updated. Could that be done in another pull request?

PR #550, opened 2023-01-20
fix: Rework `useAppBootsrap` and fix deeplinks implementations

Ldoppea:
With previous implementation we had to use `setInitialRoute()` and `setInitialScreen()` to define the initial app state

This was problematic as it was difficult to understand which route should be defined as `initial route` and which one should be defined as `initial screen`

Also the usage of `initialRoute.stack = <some_route>` and `initialScreen.stack = <some_route>` with addition to the presence of `intialRoute.route = routes.stack` made it even more confusing about what is a `root` and what is a `stack`

This complexity was due to the fact that we split the routes declarations between `RootNavigator` and `StackNavigator`. Here again the rule defining where to put each route was unclear

To simplify this we chose to refactor the code to use one single `Navigator` for every routes

This simplifies the API as we only need to use `setInitialRoute()` now, and its parameters only consist of a `route` and a set of `params`

The only exception is the nested `MainAppNavigator` that has been introduced in the `fix: Handle deeplinks on cold start` commit. See the commit description to understand why it is needed
___

TODO:

- [x] Tests on iOS
- [x] Tests on Android
- [x] Document test scenario in a Paper document

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -64,16 +64,35 @@ const App = ({ setClient }) => {
     return null
   }
 
+  const MainAppNavigator = () => (
+    <Root.Navigator
+      initialRouteName="default"
+      screenOptions={{ headerShown: false }}
+    >
+      <Stack.Screen
+        name="default"
+        component={HomeScreen}
+        initialParams={initialRoute.params}
+      />
+    </Root.Navigator>
+  )
+
   const RootNavigator = () => (
     <Root.Navigator
       initialRouteName={initialRoute.route}
       screenOptions={{ headerShown: false }}
     >
-      <Stack.Screen
-        name={routes.home}
-        component={HomeScreen}
-        initialParams={initialRoute.params}
-      />
+      {/*
+      We embed routes.home into its own Navigator so we prevent initialParams
+      to be re-applied when calling goBack() from the CozyApp route
+      i.e:
+       - We open the app with a deeplink https://links.mycozy.cloud/flagship/drive/?fallback=http://drive.cozy.tools:8080/
+       - routes.home is called with {initialParams:{cozyAppFallbackURL: 'http://drive.cozy.tools:8080/'}}
+       - routes.home reads initialParams and redirects to routes.cozyapp
+       - user press the back button so the app redirects to routes.home
+       - routes.home SHOULD NOT read initialParams and SHOULD NOT redirect to routes.cozyapp
+      */}
+      <Root.Screen name={routes.home} component={MainAppNavigator} />
 
       <Stack.Screen
         name={routes.authenticate}
--- a/src/App.js
+++ b/src/App.js
@@ -55,7 +55,7 @@ const App = ({ setClient }) => {
 
   useNetService(client)
 
-  const { initialScreen, initialRoute, isLoading } = useAppBootstrap(client)
+  const { initialRoute, isLoading } = useAppBootstrap(client)
 
   useGlobalAppState()
   useCookieResyncOnResume()
@@ -64,34 +64,34 @@ const App = ({ setClient }) => {
     return null
   }
 
-  const StackNavigator = () => (
-    <Stack.Navigator
-      initialRouteName={client ? routes.home : initialScreen.stack}
+  const RootNavigator = () => (
+    <Root.Navigator
+      initialRouteName={initialRoute.route}
       screenOptions={{ headerShown: false }}
     >
       <Stack.Screen
         name={routes.home}
         component={HomeScreen}
-        {...(initialRoute.stack ? { initialParams: initialRoute.stack } : {})}
+        initialParams={initialRoute.params}
       />
 
       <Stack.Screen
         name={routes.authenticate}
-        initialParams={initialScreen.params}
+        initialParams={initialRoute.params}
       >
         {params => <LoginScreen setClient={setClient} {...params} />}
       </Stack.Screen>
 
       <Stack.Screen
         name={routes.onboarding}
-        initialParams={initialScreen.params}
+        initialParams={initialRoute.params}
       >
         {params => <OnboardingScreen setClient={setClient} {...params} />}
       </Stack.Screen>
 
       <Stack.Screen
         name={routes.instanceCreation}
-        initialParams={initialScreen.params}
+        initialParams={initialRoute.params}
       >
         {params => <CreateInstanceScreen {...params} />}
       </Stack.Screen>
@@ -99,15 +99,6 @@ const App = ({ setClient }) => {
       <Stack.Screen name={routes.welcome}>
         {params => <WelcomeScreen setClient={setClient} {...params} />}
       </Stack.Screen>
-    </Stack.Navigator>
-  )
-
-  const RootNavigator = () => (
-    <Root.Navigator
-      initialRouteName={initialScreen.root}
-      screenOptions={{ headerShown: false }}
-    >
-      <Root.Screen name={routes.stack} component={StackNavigator} />
 
       <Root.Screen
         name={routes.error}
@@ -122,7 +113,6 @@ const App = ({ setClient }) => {
           presentation: 'transparentModal',
           animationEnabled: false
         }}
-        {...(initialRoute.root ? { initialParams: initialRoute.root } : {})}
       />
 
       <Root.Screen
--- a/src/App.js
+++ b/src/App.js
@@ -55,10 +55,7 @@ const App = ({ setClient }) => {
 
   useNetService(client)
 
-  const { initialScreen, initialRoute, isLoading } = useAppBootstrap(
-    client,
-    setClient
-  )
+  const { initialScreen, initialRoute, isLoading } = useAppBootstrap(client)
 
   useGlobalAppState()
   useCookieResyncOnResume()
--- a/src/hooks/useAppBootstrap.functions.ts
+++ b/src/hooks/useAppBootstrap.functions.ts
@@ -9,6 +9,10 @@ interface OnboardingParams {
   fqdn: string | null
 }
 
+const MAIN_APP = 'home'
+const FALLBACK_PARAM = 'fallback'
+const UNIVERSAL_LINK_BASE_PATH = 'flagship'
+
 export const parseOnboardingURL = (
   url: string | null
 ): OnboardingParams | undefined => {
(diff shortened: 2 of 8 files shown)


src/hooks/useAppBootstrap.functions.ts
```diff
@@ -0,0 +1,75 @@
...
+import { getErrorMessage } from '/libs/functions/getErrorMessage'
+
+const log = Minilog('useAppBootstrap.functions')
+
+interface OnboardingParams {
+  onboardUrl: string | null
+  fqdn: string | null
+}
+
+export const parseOnboardingURL = (
```
acezard, line 12: I think as a pure function it should be unit tested

src/hooks/useAppBootstrap.functions.ts
```diff
@@ -0,0 +1,75 @@
...
+    return undefined
+  }
+}
+
+interface FallbackUrl {
+  mainAppFallbackURL: string | undefined
+  cozyAppFallbackURL: string | undefined
+}
+
+export const parseFallbackURL = (url: string | null): FallbackUrl => {
```
acezard, line 46: I think as a pure function it should be unit tested (2)

src/hooks/useAppBootstrap.js
```diff
@@ -1,72 +1,24 @@
...
-  }
-
-  try {
-    const makeURL = new URL(url)
-    const fallback = makeURL.searchParams.get('fallback')
-    const isHome = makeURL.pathname.split('/')[1] === 'home'
+import {
+  parseFallbackURL,
+  parseOnboardingURL
+} from './useAppBootstrap.functions'
```
acezard, line 16: think we should still use aliased import

src/hooks/useAppBootstrap.js
```diff
@@ -1,72 +1,24 @@
...
-  } catch (error) {
-    log.error(
-      `Something went wrong while trying to parse fallback URL data: ${error.message}`
-    )
-    return defaultParse
-  }
-}
+if (devConfig.ignoreLogBox) LogBox.ignoreAllLogs()
 
 export const useAppBootstrap = client => {
```
acezard, line 20: just thinking: at some point we should get rid of `useAppBootstrap` and separate the logic like you did in this PR into multiple modules that we can compose in a lifecycle app hook or even during runtime after initial boot if needed

src/libs/intents/localMethods.ts
```diff
@@ -30,7 +30,7 @@ export const asyncLogout = async (client?: CozyClient): Promise<null> => {
   await deleteKeychain()
   await clearCookies()
   await clearData()
-  RootNavigation.reset(routes.stack, { screen: 'welcome' })
+  RootNavigation.reset(routes.welcome, { screen: 'welcome' })
```
acezard, line 33: note for later: `reset('foo', { screen : 'foo' })` looks weird, we should rewrite reset so the reset route is also automatically the first indexed route. Actually I don't know why it would be otherwise

src/screens/home/components/HomeView.js
```diff
@@ -114,6 +125,35 @@ const HomeView = ({ route, navigation, setLauncherContext, setBarStyle }) => {
     }
   }, [uri, client, route, navigation, session])
 
+  useEffect(
```
acezard, line 128: I think we should strive to not add any more logic into view components at this point.
Using the pure functions (when possible) with hook-as-a-bridge seem like a good replacement.
Also anyway the logic that I see here should not be triggered by a View in any case (seems higher level in terms of abstraction), but I might have missed something

src/App.js
```diff
@@ -50,10 +50,7 @@ const App = ({ setClient }) => {
 
   useNetService(client)
 
-  const { initialScreen, initialRoute, isLoading } = useAppBootstrap(
-    client,
-    setClient
-  )
+  const { initialRoute, isLoading } = useAppBootstrap(client)
```
acezard, line 53: maybe we could use `initialParams` instead of `initialRoute.params`? I think that would go well with the spirit of this PR

src/hooks/useAppBootstrap.functions.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  if (url === null) {
+    return defaultParse
+  }
+
+  try {
+    const makeURL = new URL(url)
+    const fallback = makeURL.searchParams.get('fallback') ?? undefined
+    const isHome =
+      makeURL.pathname.startsWith('/home') ||
```
acezard, line 60: there are a number of "magic strings" here and I'm not sure I'm too comfortable with that. For instance `"home"` already exists in `routes.js`. But what if tomorrow some other route name replaces `home`? Maybe we should use something like `mainPage` or `defaultPage`, just suggestions, trying to pass an idea

src/App.js
Ldoppea: `initialRoute` and `initialParams` are part of the same concept (i.e. deep links) so I would prefer to have them "linked" into the same `useState` (even if today react is able to handle multiple setState in a single re-render)

However I can split them in the hook's return. Something like:
```js
  return {
    client,
    initialRoute: initialRoute.route,
    initialParams: initialRoute.params
    isLoading
  }
```
But I'm not sure this will improve clarity that much

src/hooks/useAppBootstrap.functions.ts
Ldoppea: I didn't add tests because those are already covered by `useAppBootstrap.spec.js`. I see this extraction more like an implementation detail. But this is pretty subjective. If I add unit tests, those will be kind of duplicates. 
What do you think about this?

src/hooks/useAppBootstrap.functions.ts
Ldoppea: What about this? https://github.com/cozy/cozy-flagship-app/compare/d65f4daf44c04c41d3ab436de7f31e9def13486c..c15517c19bb895b6b1682f74e176cd27f4d98b8b

I'm not sure of how this can change in the future.

For me the `/home` path in UniversalLinks is about some external app that requests to display `cozy-home`. It shouldn't have to worry about our MainApp concept. But maybe it should?
@Crash-- do you have any thoughts on this?

src/hooks/useAppBootstrap.js
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/pull/550/commits/cd72b2044f0c573367989e1bbbd0dd2f1cad6805

src/hooks/useAppBootstrap.js
Ldoppea: Yes, now that I extracted the URL parse logic it would be easy to split this into 2 hooks, one for the bootstrap on cold start, one for the URL interception on runtime.

However I'm not sure if the hideSplashScreen part can be easily removed from the bootstrap hook.

Ldoppea: I propose to do this in another PR if you are ok

src/hooks/useAppBootstrap.functions.ts
Crash--: I'm not sure to really understand the discussion here, but here is what I can say: 
- We'll be able to specify a new defaultApp (maybe what you call mainapp?) for a cozy. For now, this defaultApp is the Cozy-Home, it'll change in the future. 
- If you have universal link for `links.mycozy.cloud/flagship/toto`, then we should open the `toto` app. If we have link to `/flagship/home`, then we have to open the `home` (this is the slug so cozy-home). If we only have `/flagship`, then we should open the defaultApp.

src/libs/intents/localMethods.ts
Ldoppea: > we should rewrite reset so the reset route is also automatically the first indexed route

I'm not sure about this. In the NetService you use reset to go back to the callbackRoute that is not necessary the first indexed route

Also you can open the app with the first route being `home` but on logout you want to reset to the `welcome` route.
However i'm not sure why we need the `{ screen: ... }` param here.

src/screens/home/components/HomeView.js
Ldoppea: > the logic that I see here should not be triggered by a View in any case (seems higher level in terms of abstraction), but I might have missed something

Here I got the same problem you had in the past for the lockscreen.
If the initial route is `drive` we want to display the CozyAppScreen, but we still want to load the HomeScreen in background. We also want the back button to correctly go back to the `home` route.

The solution I found was to navigate to `home` first then the home detects it should open `drive`.

However I agree if you want to split this component into a view/hook organisation. I propose to do it in another PR

src/hooks/useAppBootstrap.functions.ts
Ldoppea: Yes that was the point: mainapp = defaultapp

My understanding is that today `flagship/home` will open the "main screen" (a.k.a CozyHomeScreen as it is named today)
But in the future we will have to check if `flagship/home` is the default app or if we should open it as a cozy-app. But maybe /home just won't exist on cozys with a different defaultapp?

Crash--: > But in the future we will have to check if flagship/home is the default app or if we should open it as a cozy-app (using CozyAppScreen). 

Oh, this is a nice remark, because, yes, /home will still be there even with a different defaultapp. Since this is the slug of an app.

src/hooks/useAppBootstrap.functions.ts
acezard: I agree with you, it can be considered an implementation detail and thus shouldn't be tested.

src/hooks/useAppBootstrap.functions.ts
acezard: Edit: implem detail so no need

src/hooks/useAppBootstrap.js
acezard: Yes it was definitely for another PR

src/libs/intents/localMethods.ts
acezard: Well that's the point, with the new routing, reset() doesn't need two params, but just one, since we don't have two navigators (well we still do, but one has only one route so we can consider it as a route).

But anyway this is to be considered, it's not directly linked to your PR but the question is raised from the new routing

src/screens/home/components/HomeView.js
acezard: Yeah agree for the render problem.

Ok for refactoring later

src/App.js
acezard: alright agree with you

src/hooks/useAppBootstrap.functions.ts
acezard: yeah that's why I think we should use semantic variables right now as it's very easily done, just to keep in mind that nothing is hardcoded nor it should be

acezard (approved):

PR #551, opened 2023-01-20
feat: migrate existing CI from Travis to GitHub Actions

acezard:
## What does this do?

Remove travis.yml and replace it with a GitHub Action (https://docs.github.com/en/actions/migrating-to-github-actions/migrating-from-travis-ci-to-github-actions).

## Why did you do this?

This feels logical because we are going to use GitHub Actions for Continuous Delivery, so it makes sense to do the same for Integration especially since the migration is easy. Also, it's easier to leverage the GitHub ecosystem with its Actions than with Travis. Overall, I feel this is a net gain for not much work.

## Who/what does this impact?

This should impact no one, the CI will still run like before. The only difference is that it's triggered only when pull requesting to master, to avoid wasting computing time for unused runs.

## How did you test this?

It was tested directly on this PR, by getting a successful check from the GitHub Action that was created.


package.json
```diff
@@ -16,8 +16,9 @@
     "postinstall": "patch-package",
     "pre-commit": "yarn lint",
     "start": "react-native start",
-    "test": "jest connectors/**/* src/**/* scripts/* --detectOpenHandles --forceExit",
+    "test": "jest connectors/**/* src/**/* scripts/* --detectOpenHandles --forceExit --coverage",
```
Ldoppea, line 19: adding `--coverage` by default would slow down local tests in `--watch` mode, why do we need to add it here compared to putting it in the action's code?

Ldoppea (approved):

package.json
acezard: true, I guess we don't need it in the CI either anyway since we don't take any action with the result

PR #557, opened 2023-01-25
feat: add .env file

acezard:
## What does this do?

This will be how we manage build variables now.

## Why did you do this?

Rationalization of the app secrets and structure that match Github secrets. This change was kindled by the implementation of the CI/CD process.

## Who/what does this impact?

When developing locally, you definitely don't need all the variables.
Only the auth token for Sentry is mandatory.
You will need more credentials if you need
to produce actual builds locally.

## How did you test this?

Tested locally and in CI/CD environments.

## Handling breaking changes

- Write your Sentry auth key in local .env (SENTRY_AUTH_TOKEN)
- (optional) Write your Safetynet key in local .env (ANDROID_SAFETY_NET_API_KEY)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,32 @@
+# Copy this file and rename it to .env and fill the missing values as needed.
+# Only the SENTRY_AUTH_TOKEN is mandatory when developing locally. 
+# You will need more credentials if you need to produce actual signed builds.
+
+# --- ANDROID BUILD ---
+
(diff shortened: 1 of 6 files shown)


Crash--:
Don't hesitate to ping every one to tell us that there is a breaking change.

Crash-- (approved):

sblaisot (approved):

PR #558, opened 2023-01-25
fix: update connector launcher

acezard:
## What does this do?

- Handle stack-only connectors (no registry publication)
- Provide the correct clients to both Launcher and LauncherView
- Read main.js instead of index.js in the cache

## Why did you do this?

Connectors weren't launching properly, especially if not published in the registry

## Who/what does this impact?

- Everyone working on connectors should perceive better workflow and less bugs

## How did you test this?

- Manually, installing `git://github.com/konnectors/template_ccc.git#build` on my staging environment (not local dev stack)

### DIFF
--- a/src/libs/client.js
+++ b/src/libs/client.js
@@ -8,8 +8,9 @@ import { loginFlagship } from './clientHelpers/loginFlagship'
 import { normalizeFqdn } from './functions/stringHelpers'
 import { SOFTWARE_ID, SOFTWARE_NAME } from './constants'
 
-import { androidSafetyNetApiKey } from '/constants/api-keys'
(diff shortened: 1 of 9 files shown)


src/libs/client.js
```diff
@@ -421,12 +421,12 @@ export const fetchCozyAppArchiveInfoForVersion = async (
       tarPrefix
     }
   } catch (e) {
-    if (e.status === 404) {
-      return {
-        tarPrefix: ''
-      }
-    }
+    log.error(
```
acezard, line 424: mostly when the connector is on the stack but not on the registry

doubleface (approved):

PR #560, opened 2023-01-25
feat: Remove cozy-client-js when creating or updating file

LucsT:
This PR remove create of cozy-client-js and use new cozy client.
Aim to be stacked with : https://github.com/cozy/cozy-flagship-app/pull/555

### DIFF
--- a/src/libs/connectorLibs/saveFiles.js
+++ b/src/libs/connectorLibs/saveFiles.js
@@ -1,11 +1,9 @@
 import Minilog from '@cozy/minilog'
 import get from 'lodash/get'
 import omit from 'lodash/omit'
-import { Client } from 'cozy-client-js'
 import { Q } from 'cozy-client'
 import retry from 'bluebird-retry'
 
-let cozy
 let client
 
 const log = Minilog('saveFiles')
@@ -25,7 +23,6 @@ const saveFiles = async (entries, folderPath, options = {}) => {
   }
 
   client = options.client
-  cozy = initCozyClientJs(client)
 
   const saveOptions = {
     folderPath,
@@ -124,7 +121,7 @@ const saveEntry = async function (client, entry, options) {
         interval: 1000,
         throw_original: true,
         max_tries: options.retry,
-        args: [client, entry, options, method, file ? file._id : undefined]
+        args: [client, entry, options, method, file ? file : undefined]
       }).catch(err => {
         if (err.message === 'BAD_DOWNLOADED_FILE') {
           log.warn(
@@ -272,13 +269,13 @@ async function getFileFromPath(client, entry, options) {
   }
 }
 
-async function createFile(client, entry, options, method, fileId) {
+async function createFile(client, entry, options, method, file) {
   const folder = await client
     .collection('io.cozy.files')
     .statByPath(options.folderPath)
   let createFileOptions = {
     name: getFileName(entry),
-    dirID: folder.data._id
+    dirId: folder.data._id
   }
   if (options.contentType) {
     createFileOptions.contentType = options.contentType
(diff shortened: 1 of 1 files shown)


src/libs/connectorLibs/saveFiles.js
```diff
@@ -301,7 +301,9 @@ async function createFile(client, entry, options, method, fileId) {
 
   let fileDocument
   if (method === 'create') {
-    fileDocument = await cozy.files.create(toCreate, createFileOptions)
+    fileDocument = await client
+      .collection('io.cozy.files')
+      .createFile(toCreate, createFileOptions)
```
Crash--, line 306: Tu devrais plutôt utiliser le save() dans ce cas. 

Pour toutes les taches « CRUD » tu ne devrais jamais utiliser la collection directement car elle bypass plein de mécanisme de Cozy client. Un jour, on pourra accéder depuis le client directement à toutes les méthodes des collections aujourd’hui seules les crud sont accessibles.

paultranvan: > car elle bypass plein de mécanisme de Cozy client.

Plus spécifiquement, tu bypass notamment le store en faisant ça. Aujourd'hui on essaye de n'utiliser les collections que pour des méthodes dont on a pas accès directement depuis le client, ou pour des cas très spécifiques où on veut explicitement ne pas utiliser le store.

LucsT: Using save for creating and updating files

src/libs/connectorLibs/saveFiles.js
```diff
@@ -305,14 +307,23 @@ async function createFile(client, entry, options, method, fileId) {
...
-      createFileOptions
-    )
+    const clientResponse = client.save({
+      _id: file._id,
+      _rev: file._rev,
+      _type: 'io.cozy.files',
+      data: toCreate,
+      ...createFileOptions
+    })
+    fileDocument = clientResponse.data
```
paultranvan, line 326: This might be out of scope of this PR, but the `method` param to determine if it's a create or update feels a bit weird and unnecessary. We could simply save the file, and let cozy-client deal with the create or update, which is based on the existence or not of the  `_rev` attribute : https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L792-L797

src/libs/connectorLibs/saveFiles.js
```diff
@@ -305,14 +307,23 @@ async function createFile(client, entry, options, method, fileId) {
 
   let fileDocument
   if (method === 'create') {
-    fileDocument = await cozy.files.create(toCreate, createFileOptions)
+    const clientResponse = await client.save({
+      _type: 'io.cozy.files',
```
paultranvan, line 311: nit: a constant var could be used here, rather than a hardcoded string

paultranvan (approved):
Approving as it seems fine for the PR scope, but I have concerns about the API

src/libs/connectorLibs/saveFiles.js
```diff
@@ -272,13 +272,15 @@ async function getFileFromPath(client, entry, options) {
   }
 }
 
-async function createFile(client, entry, options, method, fileId) {
+
+async function createFile(client, entry, options, method, file) {
```
paultranvan, line 276: Probably out of scope, but it feels weird to have the `file` in last position, especially behind `options` and `method`. I would have put it just after the `client` (and removed the method, see other comment)

src/libs/connectorLibs/saveFiles.js
LucsT: Yeah we already speak about that with @doubleface in the case of save. But we want to erase cozy-client-js first, propagate into the future ccc_lib before beginning to optimized the code

paultranvan: > But we want to erase cozy-client-js first, propagate into the future ccc_lib before beginning to optimized the code


Ok, but I feel like it would have been more efficient to do all the API improvements at the same time :)

Anyways, maybe we should list all the possible improvements in a trello card? Especially since there are some crossed remarks with @Crash-- in the other PR: https://github.com/cozy/cozy-flagship-app/pull/555

PR #570, opened 2023-02-01
fix: remove warning from CI check

acezard:
'CozyProvider' is defined but never used

Will fix the only quality_checks warning


acezard:
Force merging it as it's trivial and validated by the CI check

PR #571, opened 2023-02-01
feat: add connector timeout

acezard:
## What does this do?

- When LauncherView is instantiated, it starts a 15 minutes timer. It will kill itself after this time has passed if it didn't stop the timer beforehand.

## Why did you do this?

- We're doing this to cleanup broken states in connectors

## Who/what does this impact?

- People working on connectors might experience less bugs

## How did you test this?

- Unit test for the timeout function
- Manually for the actual integration test

### DIFF
--- a/src/libs/functions/convertMinutesToMs.ts
+++ b/src/libs/functions/convertMinutesToMs.ts
@@ -1,0 +1,2 @@
+export const minutesToMilliseconds = (minutes: number | string): number =>
+  Number(minutes) * 60 * 1000
--- a/src/screens/connectors/LauncherView.js
+++ b/src/screens/connectors/LauncherView.js
@@ -11,6 +11,10 @@ import { getDimensions } from '/libs/dimensions'
 import ReactNativeLauncher from '/libs/ReactNativeLauncher'
 import { getColors } from '/ui/colors'
 import strings from '/constants/strings.json'
+import {
+  startTimeout,
+  stopTimeout
+} from '/screens/connectors/core/handleTimeout'
 
 const log = Minilog('LauncherView')
 
@@ -70,6 +74,7 @@ class LauncherView extends Component {
   async componentDidMount() {
     this.launcher = new ReactNativeLauncher()
     this.launcher.setLogger(this.props.onKonnectorLog)
+
     this.launcher.setStartContext({
       ...this.props.launcherContext,
       client: this.props.client,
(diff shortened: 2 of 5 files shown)


Ldoppea:
`await this.launcher.start({ initConnectorError })` contains many steps (ensureAuthenticated, fetch files, etc)

How does this behaves if the `stop()` is called in the middle of the `start()` scenario?

Ldoppea:
What I'm wondering is if:
- The connector starts and is veeery slow
- The timeout occurs and call `setLauncherContext({ state: 'default' })`
  - This would unmount the LauncherView
  - But the `await this.launcher.start(...)` Promise may continue to live in memory
- If the connector was inside of the `fetch` it may (not sure which one would happen)
  - hang forever because the post-me message would never answer
  - throw an exception because the post-me message would also timeout
- In case of an exception, it may call `stop()` a 2nd time [here](https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/ReactNativeLauncher.js#L143)
  - So the stack would receive 2 `updateJobResult` call
  
Did you try this scenario?

acezard:
> What I'm wondering is if:
> 
> * The connector starts and is veeery slow
> * The timeout occurs and call `setLauncherContext({ state: 'default' })`
>   
>   * This would unmount the LauncherView
>   * But the `await this.launcher.start(...)` Promise may continue to live in memory
> * If the connector was inside of the `fetch` it may (not sure which one would happen)
>   
>   * hang forever because the post-me message would never answer
>   * throw an exception because the post-me message would also timeout
> * In case of an exception, it may call `stop()` a 2nd time [here](https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/ReactNativeLauncher.js#L143)
>   
>   * So the stack would receive 2 `updateJobResult` call
> 
> Did you try this scenario?

Didn't, I need to get some insight on the subject. I assumed one stop was definitive

doubleface:
@Ldoppea Right, in that case, stop is actually called multiple times.

I thought that the close method of the ReactNativeLauncher would stop the execution but it doesn't.

I think we should use an AbortController to trigger the end of the start() promise when the stop function has been called (and not call stop() one more time in that case)

doubleface (approved):
Let's handle the question raised in https://github.com/cozy/cozy-flagship-app/pull/571#issuecomment-1429848192 in another PR where :

 - we allow the start method to be aborted with AbortController (available in react native)
 - call the abort signal in the stop method
 - do not call the stop method when aborted

PR #572, opened 2023-02-01
Feat  forbid parallel connectors   rework

acezard:
## What does this do?

- Forbid the launch of a new connector if one is already running
- Display a dialog with confirm so the user can understand what happened

## Why did you do this?

- This is needed because connectors can't run correctly in parallel at the moment so we have to avoid this situation in the first place

## Who/what does this impact?

- For the moment it should reduce the number of bugs for local development. Although if a connector never stops, you can never open another one so it can be a big problem. But it's easy to hack the feature to disable it.

## How did you test this?

- Test manually only. I wanted to add integration tests but it's too complex at the moment with the current architecture. I will work on having a LauncherView mock that can be useful and usable for more testing

### DIFF
--- a/src/screens/home/HomeScreen.tsx
+++ b/src/screens/home/HomeScreen.tsx
@@ -6,36 +6,31 @@ import {
   RouteProp
 } from '@react-navigation/native'
 
-import CozyClient from 'cozy-client'
-
 import HomeView from '/screens/home/components/HomeView'
 import LauncherView from '/screens/connectors/LauncherView'
 import { StatusBarStyle } from '/libs/intents/setFlagshipUI'
 import { styles } from '/screens/home/HomeScreen.styles'
-import { useLauncherClient } from '/hooks/useLauncherClient'
-import { LauncherContext } from '/libs/connectors/models'
+import { useLauncherContext } from '/screens/home/hooks/useLauncherContext'
 
 interface HomeScreenProps {
   navigation: NavigationProp<ParamListBase>
   route: RouteProp<ParamListBase>
 }
 
-const isLauncherReady = (
-  context: LauncherContext
-): context is LauncherContext => context.state !== 'default'
-
-const isClientReady = (client: CozyClient | undefined): client is CozyClient =>
-  Boolean(client)
-
 export const HomeScreen = ({
   navigation,
   route
 }: HomeScreenProps): JSX.Element => {
   const [barStyle, setBarStyle] = useState(StatusBarStyle.Light)
-  const [launcherContext, setLauncherContext] = useState<LauncherContext>({
-    state: 'default'
-  })
-  const { launcherClient } = useLauncherClient(launcherContext.value)
+  const {
+    LauncherDialog,
+    canDisplayLauncher,
+    launcherClient,
+    launcherContext,
+    resetLauncherContext,
+    setLauncherContext,
+    trySetLauncherContext
+  } = useLauncherContext()
 
   return (
     <View style={styles.container}>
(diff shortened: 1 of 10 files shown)


src/hooks/useLauncherClient.ts
```diff
@@ -9,17 +9,17 @@ interface useLauncherClientReturn {
   launcherClient?: CozyClient
 }
 export const useLauncherClient = (
-  launcherContext?: LauncherContext
+  launcherContextValue?: LauncherContext['value']
```
Ldoppea, line 12: Why not exporting `LauncherContextValue` from `models`?

src/screens/home/HomeScreen.tsx
```diff
@@ -35,7 +27,9 @@ export const HomeScreen = ({
   const [launcherContext, setLauncherContext] = useState<LauncherContext>({
     state: 'default'
   })
-  const { launcherClient } = useLauncherClient(launcherContext.value)
+  const { canDisplayLauncher, launcherClient } =
+    useLauncherWrapper(launcherContext)
+  const [concurrentConnector, setConcurrentConnector] = useState('')
```
Ldoppea, line 32: I would set `useState(undefined)` by default

src/screens/home/HomeScreen.tsx
```diff
@@ -45,17 +39,32 @@ export const HomeScreen = ({
         navigation={navigation}
         route={route}
         setBarStyle={setBarStyle}
-        setLauncherContext={setLauncherContext}
+        setLauncherContext={(candidateContext: LauncherContext): void => {
+          if (launcherContext.value && candidateContext.value) {
```
Ldoppea, line 43: Is it possible to have `candidateContext.value` being falsy?

acezard: It should be but typescript thinks it is (because we made it that way). Have to find a better way to type infer

src/screens/home/HomeScreen.tsx
acezard: Yeah I wanted to stay always a string but can work too

src/hooks/useLauncherClient.ts
acezard: indeed, will update

acezard: ✔️

src/screens/home/HomeScreen.tsx
acezard: ✔️

src/screens/home/hooks/useLauncherWrapper.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (!isLauncherReady(launcherContext)) {
+      return false
+    }
+
+    if (!isClientReady(launcherClient)) {
+      return false
+    }
+
+    return true
+  }
```
Ldoppea, line 31: nit: can be simplified as
```js
const canDisplayLauncher = (): boolean =>
  isLauncherReady(launcherContext) && isClientReady(launcherClient)
```

With that form, I'm wondering if we should still have an entire hook for this as we can just have those two lines in `useLauncherContext` so we won't have to understand the difference and relationship between the `Context` and the `Wrapper`

```js
const { launcherClient } = useLauncherClient(launcherContext.value)
const canDisplayLauncher = (): boolean =>
  isLauncherReady(launcherContext) && isClientReady(launcherClient)
```

But this is very subjective, I know it it easier for you to read split logic contrary to me who prefer condensed logic.

src/screens/home/HomeScreen.tsx
```diff
@@ -45,17 +40,19 @@ export const HomeScreen = ({
...
         <LauncherView
           launcherClient={launcherClient}
           launcherContext={launcherContext.value}
-          retry={(): void => setLauncherContext({ state: 'default' })}
+          retry={resetLauncherContext}
           setLauncherContext={setLauncherContext}
         />
       )}
+
+      {LauncherDialog}
```
Ldoppea, line 55: nit: with that name we expect to potentially see something different than an error, with maybe logic inside. Should we rename it LauncherErrorDialog?

Ldoppea (approved):

src/screens/home/hooks/useLauncherWrapper.ts
acezard: Agree, I will handle a refactor into another PR, I have to think about a pattern

src/screens/home/HomeScreen.tsx
acezard: Yes, that was my intent actually, the HomeScreen doesn't really want or need to know what's going to display in that LauncherDialog props. One thing is sure it will be a Dialog, but not necessarily linked to an error.

PR #581, opened 2023-02-03
doc: update codeowners

acezard:
## What does this do?

- codeowners updated to flagship devs

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @acezard @Ldoppea @doubleface
+* @acezard @Ldoppea @zatteo


CODEOWNERS
```diff
@@ -1 +1 @@
-* @acezard @Ldoppea @doubleface
```
Ldoppea, line 1: Why do we remove doubleface?

renovate.json
```diff
@@ -1,5 +1,14 @@
...
   "extends": [
     "cozy"
+  ],
+  "packageRules": [
+    {
+      "matchUpdateTypes": [
+        "patch"
+      ],
+      "matchCurrentVersion": ">= 1.0.0",
+      "automerge": true
```
Ldoppea, line 11: Does this means that renovate can automerge its PRs without our approval? I'm not sure I would be confident with that.

Ldoppea (commented):
Why did you remove dependabot? Is it because renovate does a similar thing?

CODEOWNERS
acezard: As the connector files are going to another repo, it makes sense to me. @doubleface is assigned on every pull request we make but 99% of the time he is not really concerned by the changes we make. What do you think @doubleface? I mean, this change is mainly to remove the amount of unnecessary notifications you receive

renovate.json
acezard: I allowed this only for patches in semver (not minor updates, are minor updates rarely strictly follow semver and are often breaking). What do you think? It would help reduce bot spam

acezard:
> Why did you remove dependabot? Is it because renovate does a similar thing?

yeah from what I understand dependabot is inferior to renovate. I don't think we need two bots for deps updating. I looked up and Renovate seems the better of the two. Any insight @Crash-- ?

acezard:
I think we should all @zatteo by the way now in codeowners for automatic review assign.
cc @Crash-- @Ldoppea

Crash--:
Can you split the PR with only the codeowners change please? 

The fact that you want to auto-merge minor is a no-go for me. You can do that for the cozy-* packages and maybe for third party lib like lodash or else, but you can't do that for react native ecosystem packages.

acezard:
> Can you split the PR with only the codeowners change please?
> 
> The fact that you want to auto-merge minor is a no-go for me. You can do that for the cozy-* packages and maybe for third party lib like lodash or else, but you can't do that for react native ecosystem packages.

Just to be sure we're clear on that, I wanted to automerge the patches, not the minor versions (which are indeed very often breaking). Agree on splitting PRs

Crash--:
I still have the same opinion on that subject, even if we are speaking about patches for the RN ecosystem.

Crash--:
Also about the bot spam, I'm not agree (but since I'm not maintaining this project this is just my opinion, nothing more). Having a PR with updated version is useful to know if you have missed versions or else. Yes, it's almost having one always opened PR for each dep but : 
- We talk only about active dep 
- If we merge quickly, there is not so much noise

zatteo (approved):

PR #594, opened 2023-02-07
Feat/receive notifications

zatteo:
It still need to be tested on iOS physical.

## What does this do?
- Request notification permission
- Save Firebase Cloud Messaging token when permission is granted or token is refreshed
- Remove Firebase Cloud Messaging token when permission is removed
- Navigate to specified app when notification is opened

## Why did you do this?

To receive notifications and navigate to specified app when notification is opened.

## Who/what does this impact?

/

## How did you test this?

To send test notifications (Android only), you can send POST requests using the Firebase Cloud Messaging and a oauth2 token. [More information](https://blog.mestwin.net/send-your-test-fcm-push-notification-quickly-with-curl/). I also documented this in  this PR in the last commit.


src/libs/notifications/notifications.js
```diff
@@ -0,0 +1,28 @@
+import { navigateToApp } from '/libs/functions/openApp'
+import { generateWebLink } from 'cozy-client'
+
+const navigateFromNotification = (client, navigate, notification) => {
```
Crash--, line 4: Could be awesome to create a type for the notification object

src/libs/notifications/notifications.js
```diff
@@ -0,0 +1,28 @@
...
+  if (!url || !slug) return
+
+  const subDomainType = client.getInstanceOptions().capabilities.flat_subdomains
+    ? 'flat'
+    : 'nested'
+
+  const href = generateWebLink({
+    cozyUrl: client.getStackClient().uri,
+    subDomainType,
+    pathname: '/',
```
Crash--, line 16: I'm wondering if the pathname should not be part of the notification's content. It can be /public, /export or else. It should not be hardcoded.

src/libs/notifications/notifications.js
zatteo: I tried to convert notifications.js in ts in a new commit, tell me what you think @Crash-- @acezard

acezard (commented):
It looks good to me, but I would also have notifications tests in typescript since the implementation file itself is in typescript. It makes tests more robust albeit admittedly trickier to write.

docs/how-to-debug-notifications.md
```diff
@@ -0,0 +1,77 @@
+# How to debug notifications
+
+## Get Firebase project id
```
Ldoppea, line 3: In this section I'm missing two things:
- Where should I put this code?
- Why should I do this? (i.e. is this a prerequisite do to tests? What should I do with the projectId? etc)

docs/how-to-debug-notifications.md
```diff
@@ -0,0 +1,77 @@
...
+
+Two Firebase projects have been created, one for development environnement (_cozy-flagship-app-push-dev_) and one for production environment (_cozy-flagship-app-push-prod_).
+
+```
+import { utils } from   '@react-native-firebase/app'
+
+console.log(utils()._app._options.projectId) // cozy-flagship-app-push-dev or cozy-flagship-app-push-prod
+```
+
+## Get FCM token
```
Ldoppea, line 13: nit: Same here, why should I get the FCM token? (I read the next section so I have an answer but it would be great to have the answer directly in this section)

src/libs/notifications/notifications.ts
```diff
@@ -0,0 +1,105 @@
...
+  capabilities: {
+    flat_subdomains: string
+  }
+}
+
+export const navigateFromNotification = (
+  client: CozyClient,
+  notification: FirebaseMessagingTypes.RemoteMessage
+): void => {
+  const { data: { pathname, url, appName } = {} as NotificationData } =
```
Ldoppea, line 27: Can you describe why `appName` is better than `slug` here? Simple question as `slug` is a convention for cozy-apps in our team then I would like to understand the reason here.

src/libs/notifications/notifications.ts
```diff
@@ -0,0 +1,105 @@
...
+import { navigateToApp } from '/libs/functions/openApp'
+import { saveNotificationDeviceToken } from '/libs/client'
+
+interface NotificationData {
+  pathname: string
+  url: string
+  appName: string
+}
+
+interface InstanceOptions {
```
Ldoppea, line 17: nit: this type may be declared in `@types/cozy-client.d.ts`

Ldoppea (approved):
Looks great 👍 

I added some comments, if you take them into account and then if you are confident to merge without another review, then do it.

Do we plan to implement something in cozy-settings (or somewhere else) to warn the user that they rejected the notification authorization at OS level and to guide them if they want to re-enable them?

docs/how-to-debug-notifications.md
```diff
@@ -0,0 +1,77 @@
+# How to debug notifications
+
+## Get Firebase project id
+
+Two Firebase projects have been created, one for development environnement (_cozy-flagship-app-push-dev_) and one for production environment (_cozy-flagship-app-push-prod_).
```
Crash--, line 5: Maybe we should be clearer here.

It happens a lot to develop on a local machine but being connected on a production's cozy. In that case, if we need to test notification, then we should use the push-prod token. 

Also, we can change the "stack environnement" when we log on a "production build" of the flagship app. We're able to chose "other domain" (or enter a mail address linked to a stg or int or dev cozy) but in that case, notification will not work since we'll use the "production token" but for a "non production cozy". 

I think it's acceptable, but it means that to validate the push notification we need to build a "prod" version of the app, but with the "dev token" in order to connect on "non prod cozy". Once it's tested and validated, then we'll be able to deploy the new config on the "prod cozy" and now we should build a "prod version of the app" with the "prod token". 


I'm wrong?

ios/CozyReactNative/CozyReactNative.entitlements
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>aps-environment</key>
+	<string>development</string>
```
Crash--, line 6: CI should be updated to know to change to production if we're making a release?

ios/CozyReactNative/Info.plist
```diff
@@ -62,13 +66,20 @@
...
-	<string/>
+	<string></string>
 	<key>UIAppFonts</key>
 	<array>
 		<string>Lato-Bold.ttf</string>
 		<string>Lato-Regular.ttf</string>
 	</array>
+	<key>UIBackgroundModes</key>
+	<array>
+		<string>fetch</string>
```
Crash--, line 80: Do we really need `fetch`?

src/libs/notifications/notifications.ts
Crash--: We should have both 
Capture d’écran 2023-02-22 à 13 03 30

And in that case, we should use the `slug` instead of `appName`. 

`appName` is the name displayed to the user, `slug` is the technical name. So if we need to redirect someone on the right app, we have to use the `slug` and not the `appName`

https://github.com/cozy/cozy-stack/blob/63782e4e45faed57e3fa5fe0ae5c49480a5f6aaf/docs/client-app-dev.md?plain=1#L161-L162

ios/CozyReactNative/Info.plist
zatteo: I blindly followed rnfirebase setup for this https://rnfirebase.io/messaging/usage/ios-setup#enable-background-modes

docs/how-to-debug-notifications.md
zatteo: Firebase projects are linked to environments :
- on Android by the location of google-services.json (prod version app get the file from android/app/google-services.json and dev version app get the file from android/app/src/debug/google-services.json). It is how it works automatically.
- on iOS by adding the Firebase configuration file to the appropriate target in Xcode

So it is possible to build a "prod" version of the app with a "dev" Firebase conf but not very practical because you need to update Xcode conf or move conf files. What do you think ? Should we find another way ? Maybe using Firebase prod env in cozy int to be able to test a production app with a production Firebase to our int cozy ?

---

On the other side, if you want to connect to a production cozy from local machine: easy. yarn android --variant=release -> Firebase prod and you can connect to a prod cozy with notifications.

ios/CozyReactNative/CozyReactNative.entitlements
zatteo: I need to check

docs/how-to-debug-notifications.md
Crash--: > Firebase projects are linked to environments :

Yeah but in Cozy's world, environments can mean "cozy stack environment". It's rarely means "localhost dev / prod". 

> So it is possible to build a "prod" version of the app with a "dev" Firebase conf but not very practical because you need to update Xcode conf or move conf files. What do you think ? Should we find another way ? 

I don't think we have a way to handle that correctly, we should only improve the documentation to explain that point. 

Maybe we can find a way to show an alert within the app if we're connected to a "non prod cozy-stack" with the "prod firebase key" (and vice et versa) Is it possible?

Crash--: > On the other side, if you want to connect to a production cozy from local machine: easy. yarn android --variant=release -> Firebase prod and you can connect to a prod cozy with notifications.

Yes, but we lose logs and debug stuff, no?

docs/how-to-debug-notifications.md
zatteo: I updated the documentation. Tell me if it is enough or not.

docs/how-to-debug-notifications.md
zatteo: For android, I created 2 product flavors following [this issue](https://github.com/invertase/react-native-firebase/issues/3504#issuecomment-616796121). You can check the last commit.

At the end, we can do : 
- `yarn android --variant=prodDebug` => Firebase in "prod" and app in "dev"
- `yarn android --variant=prodRelease` =>  Firebase in "prod" and app in "prod"
- `yarn android --variant=devDebug` => Firebase in "dev" and app in "dev"
- `yarn android --variant=devRelease` =>  Firebase in "dev" and app in "prod"

`yarn android`does not work anymore because it takes _debug_ as a variant default value and _debug_ does not exist anymore.

zatteo: @Crash-- @Ldoppea what do you think of this solution ?

src/hooks/useNotifications.js
```diff
@@ -0,0 +1,37 @@
...
+        setAreNotificationsEnabled(true)
+      } else if (permission === messaging.AuthorizationStatus.DENIED) {
+        await removeNotificationDeviceToken(client)
+      }
+    })()
+  }, [client])
+
+  useEffect(() => {
+    if (areNotificationsEnabled) {
+      handleNotificationTokenReceiving(client)
```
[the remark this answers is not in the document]
zatteo: I am not sure of what will happen. So I unregistered to the two handlers. Safer and cleaner.

zatteo:
> Do we plan to implement something in cozy-settings (or somewhere else) to warn the user that they rejected the notification authorization at OS level and to guide them if they want to re-enable them?

It is not planned at the moment, but one day it will be clearly the case. And I think we will be able to reuse what you did for biometry permission.

ios/CozyReactNative/CozyReactNative.entitlements
zatteo: See https://github.com/cozy/cozy-flagship-app/pull/594/commits/94870f87f817fe593704fb38fe2c9196638459a6

No changes on iOS

src/hooks/useNotifications.js
```diff
@@ -0,0 +1,42 @@
...
+
+import { useClient } from 'cozy-client'
+import { removeNotificationDeviceToken } from '/libs/client'
+import {
+  handleNotificationTokenReceiving,
+  handleNotificationOpening,
+  requestAndGetNotificationPermission
+} from '/libs/notifications/notifications'
+
+export const useNotifications = () => {
```
acezard, line 12: This got some logic, we should extract it as pure functions and write easy unit tests for it.
Also as a new file, it should be pure TS if you are able to use it

src/hooks/useNotifications.js
```diff
@@ -0,0 +1,42 @@
...
+  const client = useClient()
+
+  const [areNotificationsEnabled, setAreNotificationsEnabled] = useState(false)
+
+  useEffect(() => {
+    ;(async function () {
```
acezard, line 18: While I agree with the spirit I think IIFE are not that great with the mandatory semi-colon prefix.
Best to name things a bit more even if it feels verbose, a good middle ground is more robust

acezard (approved):

Ldoppea (approved):

src/hooks/useNotifications.js
zatteo: It is hard to separate more because we have some React setXXX.

But I migrated to ts (and it helped to spot some issues).

PR #595, opened 2023-02-07
feat: disconnect connectors on end and on reboot

acezard:
## What does this do?

- Add to url list on launch
- Delete linked cookies after completion
- Delete all cookies stored after crash reboot

## Why did you do this?

- This allows better handling of end of life in the connector lifecycle

## Who/what does this impact?

- At this time only developers working on connectors

## How did you test this?

- Unit tests

### DIFF
--- a/src/redux/ConnectorState/ConnectorUrlsSlice.ts
+++ b/src/redux/ConnectorState/ConnectorUrlsSlice.ts
@@ -1,0 +1,35 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+
+import { RootState } from '/redux/store'
+
+export interface ConnectorUrlsState {
+  urls: string[]
+}
+
+const initialState: ConnectorUrlsState = {
+  urls: []
+}
+
+export const connectorUrlsSlice = createSlice({
+  name: 'connectorUrls',
+  initialState,
+  reducers: {
+    updateList: (state, action: PayloadAction<string>) => {
+      state.urls = Array.from(new Set([...state.urls, action.payload]))
+    },
+    clearList: state => {
+      state.urls = []
+    },
+    removeUrlFromStore: (state, action: PayloadAction<string>) => {
+      state.urls = state.urls.filter(url => url !== action.payload)
+    }
+  }
+})
+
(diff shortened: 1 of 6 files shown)


src/redux/ConnectorState/ConnectorUrlsSlice.ts
```diff
@@ -0,0 +1,35 @@
...
+const initialState: ConnectorUrlsState = {
+  urls: []
+}
+
+export const connectorUrlsSlice = createSlice({
+  name: 'connectorUrls',
+  initialState,
+  reducers: {
+    updateList: (state, action: PayloadAction<string>) => {
+      state.urls = [...state.urls, action.payload]
```
Ldoppea, line 18: Maybe we can dedupe urls?

src/screens/connectors/LauncherView.js
```diff
@@ -104,6 +108,7 @@ class LauncherView extends Component {
     if (this.launcher.removeAllListener) {
       this.launcher.removeAllListener()
     }
+    cleanConnectorCookies(this.state.worker.url)
```
Ldoppea, line 111: Shouldn't this be `cleanAllConnectorCookies()` ?

acezard: My aim was to only delete the cookies of the current connector URL, isn't that so? I only cleanAll() on reboot

acezard: I think you're right though, it's far easier to remove everything on clear. And since we don't do parallel, I don't think it's an issue if we end up deleting something that should not be there anyway

src/redux/ConnectorState/ConnectorUrlsSlice.ts
acezard: after fixing the other issue, I'm not sure it's very important (shouldn't happen too)

src/screens/connectors/LauncherView.js
acezard: 💀: after more testing, it seems I'm unable to selectively delete cookies on Android (either by deleting them, tampering with their value, or tampering with their expiration time). This is a HUGE blocker to this feature right now

PR #596, opened 2023-02-09
Fix/open app open iab

Crash--:
We don't want the third party dev to know about the flagship
behavior.

The dev should call openApp() when they want to open a app
even if the app will be opened on an other cozy (for instance
when sharing "cozy to cozy" a note)

So let's check if the URL of the client is the same as the
destination. If not, let's open the IAB.

### DIFF
--- a/src/libs/functions/urlHelpers.js
+++ b/src/libs/functions/urlHelpers.js
@@ -1,7 +1,7 @@
 import { Platform } from 'react-native'
 import { InAppBrowser } from 'react-native-inappbrowser-reborn'
 
-import { deconstructCozyWebLinkWithSlug } from 'cozy-client'
+import { deconstructCozyWebLinkWithSlug } from 'cozy-client/dist/helpers'
 
 import Minilog from '@cozy/minilog'
 
--- a/src/libs/functions/urlHelpers.js
+++ b/src/libs/functions/urlHelpers.js
@@ -40,7 +40,41 @@ export const checkIsRedirectOutside = ({ currentUrl, destinationUrl }) => {
 
   return false
 }
+/**
+ *
+ * @param {object} params
+ * @param {string} params.cozyUrl The URL of the cozy, from client.stackclient.uri
+ * @param {string} params.destinationUrl The URL we want to redirect the user
+ * @param {'flat' | 'nested'} params.subdomainType - the Cozy's subdomain type
+ * @returns {boolean} True is the destination is the same cozy
+ */
+export const isSameCozy = ({
+  cozyUrl,
+  destinationUrl,
+  subdomainType = 'flat'
+}) => {
+  try {
+    const currentUrlData = deconstructCozyWebLinkWithSlug(
+      cozyUrl,
+      subdomainType
+    )
+    const destinationUrlData = deconstructCozyWebLinkWithSlug(
+      destinationUrl,
+      subdomainType
+    )
+    if (
+      currentUrlData.cozyBaseDomain !== destinationUrlData.cozyBaseDomain ||
+      currentUrlData.cozyName !== destinationUrlData.cozyName
+    ) {
+      return false
+    }
 
+    return true
+  } catch (err) {
+    log.error('Error while calling isSameCozy', err)
+    return false
+  }
+}
 /**
  * Compare current URL and target URL and detect if the app should navigate to another cozy-app
  * or if it is another type of navigation (same cozy-app or outgoing link)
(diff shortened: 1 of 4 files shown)


Crash--:
@Ldoppea I've a question concerning the scenario with a native app. Does an href is provided?

src/libs/intents/localMethods.ts
```diff
@@ -121,7 +121,7 @@ export const localMethods = (
     hideSplashScreen: nativeMethodWrapper(hideSplashScreen),
     logout: () => asyncLogout(client),
     openApp: (href, app, iconParams) =>
-      openApp(RootNavigation, href, app, iconParams),
+      openApp(client, RootNavigation, href, app, iconParams),
```
acezard, line 124: even if we're in TS (thankfully here), I think that's starting to be a lot of arguments for that function.

This might reveal a bigger issue with the openApp systems, why do we need so many parameters?

Crash--: This method is called from the outside of the React Context. So we need to get access to the router and to the client. We can inject them by magic by using some kind of middleware, but I always prefer readability / verbosity versus magic. 

Or... we change to be event based. Like that we can listen to a specific message in a React context and we'll not have to pass those 2 args. But this is not something I want to go in.

src/libs/functions/urlHelpers.js
```diff
@@ -40,7 +40,41 @@ export const checkIsRedirectOutside = ({ currentUrl, destinationUrl }) => {
...
+      currentUrlData.cozyBaseDomain !== destinationUrlData.cozyBaseDomain ||
+      currentUrlData.cozyName !== destinationUrlData.cozyName
+    ) {
+      return true
+    }
 
+    return false
+  } catch (err) {
+    log.error('Error while calling checkIsCozyUrlIsTheDestination', err)
+    return false
```
Ldoppea, line 75: I would return `true` instead.

If we don't succeed to compare URL I prefer to have a false positive and open IAB than the reverse. This would prevent the user to be stuck in a random url.
Also the more common error source would be an exception while parsing the destination URL, which should happen only if it is from another server with different subdomain type (or just a badly formated URL)

src/libs/functions/urlHelpers.js
```diff
@@ -40,7 +40,41 @@ export const checkIsRedirectOutside = ({ currentUrl, destinationUrl }) => {
...
 }
+/**
+ *
+ * @param {object} params
+ * @param {string} params.cozyUrl The URL of the cozy, from client.stackclient.uri
+ * @param {string} params.destinationUrl The URL we want to redirect the user
+ * @param {'flat' | 'nested'} params.subdomainType - the Cozy's subdomain type
+ * @returns {boolean} True is the destination is not on the cozy
+ */
+export const checkIsCozyUrlIsTheDestination = ({
```
Ldoppea, line 51: The function's name seems off here.

First it has 2 `is` which is hard to understand. Shouldn't the first `is` be an `if`?

But even with that, if I call `checkIfCozyUrlIsTheDestination` I expect to get `true` if the sourceURL and the destinationURL are the same. But here `true` means that there are different (so we are leaving the cozy). So I would reverse the name or the logic.

Proposal: rename it to something like `checkIsRedirectOutsideCurrentCozy`

src/libs/functions/urlHelpers.js
Crash--: Oh you right, nice catch

src/libs/functions/urlHelpers.js
Crash--: > First it has 2 is which is hard to understand. Shouldn't the first is be an if?

Yep. Sorry for that. 

You right, I'll change the logic. But in fact there is already a named `checkIsRedirectOutside` that doesn't not really what I wanted. I'll think of something.

Ldoppea:
> @Ldoppea I've a question concerning the scenario with a native app. Does an href is provided?

Yes there should be an href for them too. The scenario can be tested by clicking cozy-pass from the Home screen. This uses the AppLinker component like for other cozy apps tiles

Crash--:
OK so it makes sens to open the IAB if the href is not on the Cozy. Because if not, then we'll open the Native App, but with a big chance of having the user not connected on the right cozy. OK with that?

Ldoppea:
By reading the question a new time I just want to clarify my answer:
- If the question is if an href is sent to the `openApp` method, then yes
- If the question is if that href is sent to the native app, then no

Ok to intercept any request not related to the current cozy without exception.

src/libs/functions/urlHelpers.js
Ldoppea: What if you inverse the logic and name it `isSameCozy`?

Crash--:
@Ldoppea ready for a review and you can merge it by yourself after that :)

Ldoppea (approved):

PR #597, opened 2023-02-09
feat: Use cache for fetch apps query

zatteo:
`attemptFetchApps` used directly `fetchJSON` bypassing cozy-client cache. Now we use `fetchQueryAndGetFromState` which uses cozy-client cache.


Crash-- (approved):

acezard (commented):
We should add a small PR description

PR #607, opened 2023-02-14
feat: Ensure account and trigger on the flagship app side

doubleface:
Now, the Launcher creates the account and trigger, not harvest.
The trigger is also launched.

This allows for faster account and trigger creation since the flagship
app will only have to check clisk specific aspects.

At the moment, account and trigger are still created even if the login fails.
What is missing is a way to display an error in the amiral application or harvest without a job
And we need to have a job result to trace connections attempts which have failed in logs

- feat: Use connector slug in launcher
- feat: Upgrade cozy-client to v35
- feat: Ensure account and trigger on flagship app side
- docs: Add @ts-check

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android
* [X] All changes have test coverage

### DIFF
--- a/src/libs/Launcher.js
+++ b/src/libs/Launcher.js
@@ -1,3 +1,4 @@
+// @ts-check
 /* eslint-disable no-unused-vars */
 import { Q, models } from 'cozy-client'
 import Minilog from '@cozy/minilog'
@@ -16,13 +17,19 @@ const log = Minilog('Launcher')
  * @interface
  */
 export default class Launcher {
+  constructor() {
+    this.setUserData({
+      sourceAccountIdentifier: null
+    })
+  }
   /**
    * Inject the content script and initialize the bridge to communicate it
    *
-   * @param  {Object} options.bridgeOptions : options which will be given as is to the bridge. Bridge options depend on the environment of the launcher
-   * @param  {String} options.contentScript : source code of the content script which will be injected
+   * @param  {Object} options               - options object
+   * @param  {Object} options.bridgeOptions - options which will be given as is to the bridge. Bridge options depend on the environment of the launcher
+   * @param  {String} options.contentScript - source code of the content script which will be injected
    *
-   * @return {Bridge}
+   * @returns {Promise<void>}
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async init({ bridgeOptions, contentScript }) {}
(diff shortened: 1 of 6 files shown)


src/libs/Launcher.js
```diff
@@ -169,17 +189,71 @@ export default class Launcher {
...
+  async ensureAccountTriggerAndLaunch() {
+    const startContext = this.getStartContext()
+    let { trigger, account, connector, client, job, launcherClient } =
+      startContext
+
+    if (!account) {
+      log.debug(
+        `ensureAccountAndTriggerAndJob: found no account in start context. Creating one`
+      )
+      const accountResponse = await client.create(
```
Crash--, line 209: you should use client.save()

src/libs/Launcher.js
```diff
@@ -169,17 +189,71 @@ export default class Launcher {
...
+    if (!trigger) {
+      log.debug(
+        `ensureAccountAndTriggerAndJob: found no trigger in start context. Creating one`
+      )
+      const triggerResponse = await client.create(
+        'io.cozy.triggers',
+        models.trigger.triggers.buildTriggerAttributes({
+          account,
+          konnector: connector
+        })
```
Crash--, line 225: same here. create() is only useful if you use the third arg (aka reference), if not, save is the way to go

src/libs/Launcher.js
```diff
@@ -169,17 +189,71 @@ export default class Launcher {
...
       )
     }
   }
 
+  /**
+   * Ensures that account and triggers are created and launch the trigger
+   */
+  async ensureAccountTriggerAndLaunch() {
+    const startContext = this.getStartContext()
+    let { trigger, account, connector, client, job, launcherClient } =
```
Crash--, line 202: please, make us a favor and make a PR to change `connector` to `konnector` every where in this project. Having both is confusing ...

src/libs/Launcher.js
```diff
@@ -1,5 +1,6 @@
+// @ts-check
 /* eslint-disable no-unused-vars */
-import { Q, models } from 'cozy-client'
+import CozyClient, { Q, models } from 'cozy-client'
```
Crash--, line 3: If `CozyClient` is imported for the JSDoc, you should remove it and use the `import('cozy-client').default` or something like that syntax.

acezard (approved):

src/libs/Launcher.js
doubleface: https://github.com/cozy/cozy-flagship-app/pull/617

PR #619, opened 2023-02-23
fix: handle nested domains for `isSameCozy`

acezard:
## What does this do?

- This will check when opening an URL if it's from the same nested cozy

## Why did you do this?

- Previously we determined that a nested cozy could never navigate to itself

## Who/what does this impact?

- It impacts everyone with a nested cozy, so mainly dev environments

## How did you test this?

- Manually, no unit tests as of now. The shape of deconstructed cozy web link is strange in nested mode though, I'm not sure if it's returning correct data

### DIFF
--- a/src/libs/functions/openApp.js
+++ b/src/libs/functions/openApp.js
@@ -113,11 +113,11 @@ const openConnectorInHome = (navigation, connector) => {
  * @returns {Promise}
  */
 export const openApp = (client, navigation, href, app, iconParams) => {
-  const subdomainType = client.capabilities?.flat_subdomains ? 'flat' : 'nested'
+  const subDomainType = client.capabilities?.flat_subdomains ? 'flat' : 'nested'
   const shouldOpenInIAB = !isSameCozy({
     cozyUrl: client.getStackClient().uri,
     destinationUrl: href,
-    subdomainType
+    subDomainType
   })
 
   if (shouldOpenInIAB) {
--- a/src/libs/functions/urlHelpers.js
+++ b/src/libs/functions/urlHelpers.js
@@ -1,9 +1,9 @@
-import { Platform } from 'react-native'
 import { InAppBrowser } from 'react-native-inappbrowser-reborn'
-
-import { deconstructCozyWebLinkWithSlug } from 'cozy-client/dist/helpers'
+import { Platform } from 'react-native'
 
 import Minilog from '@cozy/minilog'
+import { deconstructCozyWebLinkWithSlug } from 'cozy-client/dist/helpers'
+import { generateWebLink } from 'cozy-client'
 
 const log = Minilog('CozyWebView')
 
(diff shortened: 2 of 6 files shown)


src/libs/functions/urlHelpers.js
```diff
@@ -62,6 +62,21 @@ export const isSameCozy = ({
       destinationUrl,
       subdomainType
     )
+
+    if (subdomainType === 'nested') {
+      // http://dev.10-0-2-2.nip.io:8080 to http://contacts.dev.10-0-2-2.nip.io:8080/
+      const nestedName = currentUrlData.slug // dev
```
Crash--, line 68: For me, we should not have a slug there. It should be empty for `http://dev.10-0-2-2.nip.io:8080` but maybe you don't know that `dev` is not a slug ATM. @Ldoppea don't we have a method in cozy-client that handles that?

Ldoppea: I think I understand the problem here and seems like we are not fixing it on the more optimal place.

Here the method is based on `cozy-client/deconstructCozyWebLinkWithSlug`. The method's name explicitly states that it expects to have URLs with slugs (this is the meaning of `WithSlug`). If the provided URL doesn't contain any slug then we won't be able to correctly deconstruct the link.

Here the `isSameCozy` method is called only from `openApp()` that uses `client.getStackClient().uri` for cozyUrl, so we are sure that the url WON'T contain any slug.

Instead of adding this specific code I propose to either:
- Enforce a slug from the caller (i.e use `generateWebLink` in `openApp` or something similar)
or
- Implement a method in cozy-client called `deconstructCozyWebLinkWithoutSlug`

src/libs/functions/openApp.js
```diff
@@ -113,11 +113,11 @@ const openConnectorInHome = (navigation, connector) => {
  * @returns {Promise}
  */
 export const openApp = (client, navigation, href, app, iconParams) => {
-  const subdomainType = client.capabilities?.flat_subdomains ? 'flat' : 'nested'
+  const subDomainType = client.capabilities?.flat_subdomains ? 'flat' : 'nested'
```
Ldoppea, line 116: nit: `subdomain` is more correct than `subDomain` as it is a single world in english

Ldoppea (approved):
I added some tests to bullet proof this method

Note that this won't work if cozyUrl is `http://claude.mycozy.cloud` and if destinationUrl is also `http://claude.mycozy.cloud` with nested domains.

I'm not sure if we need this scenario to be functional so feel free to merge without fix, but in that case we should at least document this behavior.

Also current behavior seems similar to the old one if we just add the generateWebLink(). But new implementation is longer/more complex to read. But maybe I'm missing something.

src/libs/functions/openApp.js
acezard: yeah but it's the name in the cozy client function :/

acezard:
> I added some tests to bullet proof this method
> 
> Note that this won't work if cozyUrl is `http://claude.mycozy.cloud` and if destinationUrl is also `http://claude.mycozy.cloud` with nested domains.
> 
> I'm not sure if we need this scenario to be functional so feel free to merge without fix, but in that case we should at lease document this behavior.
> 
> Also current behavior seems similar to the old one if we just add the generateWebLink(). But new implementation is longer/more complex to read. But maybe I'm missing something.

sorry, not sure to understand how the destinationUrl could be the same as the client base uri?

Crash--:
> sorry, not sure to understand how the destinationUrl could be the same as the client base uri?

Today, in order to open the "default app" on a cozy, instead of requesting the default_app you can just go to the client base uri, and then the stack will redirect you to the default_app. 

Not sure this is something we will do with the openApp() call since we need to define the slug of the app to go on, but if this underlying method is used somewhere else we have to be careful.

Crash-- (approved):

PR #629, opened 2023-02-27
feat: allow direct konnector fetching

acezard:
## What does this do?

- Enable dev user to fetch Konnector file from any server

## Why did you do this?

- Crucial for easy development flow

## Who/what does this impact?

- Anyone who wants to work on Konnectors in real time

## How did you test this?

- Tested locally and manually
- I would consider this feature experimental for now

### DIFF
--- a/src/constants/dev-config.ts
+++ b/src/constants/dev-config.ts
@@ -9,5 +9,6 @@ export const devConfig = {
   enableReduxLogger: false, // Outputs to console every Redux action with payload, prev and next state
   forceHideSplashScreen: false, // Hide react-native splash screen renders, useful for debugging in case of a webview crash
   forceOffline: false, // Force offline mode by overwriting the NetInfo module and returning a fake offline state,
-  ignoreLogBox: false // Hide react-native LogBox renders but still display logs to the console,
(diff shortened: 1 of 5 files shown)


src/constants/dev-config.ts
```diff
@@ -9,5 +9,6 @@ export const devConfig = {
   enableReduxLogger: false, // Outputs to console every Redux action with payload, prev and next state
   forceHideSplashScreen: false, // Hide react-native splash screen renders, useful for debugging in case of a webview crash
   forceOffline: false, // Force offline mode by overwriting the NetInfo module and returning a fake offline state,
-  ignoreLogBox: false // Hide react-native LogBox renders but still display logs to the console,
+  ignoreLogBox: true, // Hide react-native LogBox renders but still display logs to the console,
+  konnectorServer: 'http://172.28.246.230:3000' // Use local konnectors instead of the ones from the stack, e.g. "http://172.31.47.56:3000"
```
Crash--, line 13: Il me semble que côté stack, on va ajouter un --konnectorDir comme on a un --appDir de façon justement à avoir du refresh. Et je préfère largement la solution par la stack si jamais y a des droits qui changent 🤔 . 

Ça se marche pas sur les pieds ?

PR #633, opened 2023-02-28
feat: Download files from konnectors in worker

LucsT:
Use new call to download file in konnectors.

This call is implemented in cozy-clisk in https://github.com/konnectors/libs/pull/920
We will need konnectors up to date (contentScript from clisk) when using this in flagship app.


LucsT:
Rebased on master to include some recent fix.

Crash-- (approved):

PR #635, opened 2023-03-02
Open app at onboarding end with email link

zatteo:
**End goal** : be able to open a specific app at the end of the onboarding.

The way to get the _onboarded_redirection_ is not sure for the moment. So it is a first possibility based on the onboarding email, which will be useful for testing purpose anyway.

**About the code**

~~I am not very happy about the result especially because I added code in a lot of different places. Feel free to challenge it. What do you think about the onboardingRedirection state used to pass the redirection between onboarding and the useAppBoostrap hook for example?~~

**How to test**
 - Encode your onboarded_redirection param like _drive/#/_ (root of Cozy Drive app) => _drive%2F%23%2F_
 - Encode your email in the onboard_url param like _oib35043@zslsz.com_ => _oib35043%40zslsz.com_
 - Forge the link and open it with adb or xcrun `adb shell am start -W -a android.intent.action.VIEW -d 'cozy://onboarding?flagship=true\&onboarded_redirection=drive%2F%23%2F\&onboard_url=https%3A%2F%2Fmanager.cozycloud.cc%2Fv2%2Fcozy%2Fonboard%3Femail%3Doib35043%40zslsz.com%26skip_email_step%3Dtrue'` (for Android, [you need to add a backslash to escape '&' for search params](https://stackoverflow.com/a/44782797))
 - Follow the onboarding
 - Check that you are correctly redirected to your onboarded_redirection
 
**Remaining**
 - [x] test on iOS
 - [x] add documentation about how to test in "Scénario de tests"

### DIFF
--- a/src/hooks/useAppBootstrap.functions.ts
+++ b/src/hooks/useAppBootstrap.functions.ts
@@ -6,6 +6,7 @@ const log = Minilog('useAppBootstrap.functions')
 
 interface OnboardingParams {
   onboardUrl: string | null
+  onboardedRedirection: string | null
   fqdn: string | null
 }
 
@@ -23,6 +24,9 @@ export const parseOnboardingURL = (
 
     const onboardingUrl = new URL(url)
     const onboardUrl = onboardingUrl.searchParams.get('onboard_url')
+    const onboardedRedirection = onboardingUrl.searchParams.get(
+      'onboarded_redirection'
+    )
     const fqdn = onboardingUrl.searchParams.get('fqdn')
 
     if (!onboardUrl && !fqdn) {
@@ -31,6 +35,7 @@ export const parseOnboardingURL = (
 
     return {
       onboardUrl,
+      onboardedRedirection,
       fqdn
     }
   } catch (error) {
--- a/src/hooks/useAppBootstrap.js
+++ b/src/hooks/useAppBootstrap.js
@@ -13,6 +13,9 @@ import {
   parseOnboardingURL
 } from '/hooks/useAppBootstrap.functions'
 import { useSplashScreen } from '/hooks/useSplashScreen'
+import { formatOnboardingRedirection } from '/libs/functions/formatOnboardingRedirection'
+
+let OnboardingRedirection = ''
 
 export const useAppBootstrap = client => {
   const [initialRoute, setInitialRoute] = useState('fetching')
@@ -28,7 +31,11 @@ export const useAppBootstrap = client => {
         const onboardingParams = parseOnboardingURL(onboardingUrl)
 
         if (onboardingParams) {
-          const { onboardUrl, fqdn } = onboardingParams
+          const { onboardUrl, onboardedRedirection, fqdn } = onboardingParams
+
+          if (onboardedRedirection) {
+            OnboardingRedirection = onboardedRedirection
+          }
 
           if (onboardUrl) {
             return setInitialRoute({
@@ -50,6 +57,20 @@ export const useAppBootstrap = client => {
             route: routes.welcome
           })
         }
+      } else if (OnboardingRedirection) {
+        const onboardingRedirectionURL = formatOnboardingRedirection(
+          OnboardingRedirection,
+          client
+        )
+
+        OnboardingRedirection = ''
+
+        return setInitialRoute({
+          route: routes.home,
+          params: {
+            cozyAppFallbackURL: onboardingRedirectionURL
+          }
+        })
       } else {
         const payload = await Linking.getInitialURL()
         const { mainAppFallbackURL, cozyAppFallbackURL } =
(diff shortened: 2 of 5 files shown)


src/hooks/useAppBootstrap.functions.ts
```diff
@@ -23,14 +24,18 @@ export const parseOnboardingURL = (
 
     const onboardingUrl = new URL(url)
     const onboardUrl = onboardingUrl.searchParams.get('onboard_url')
+    const onboardedRedirection = onboardingUrl.searchParams.get(
+      'onboarded_redirection'
+    )
     const fqdn = onboardingUrl.searchParams.get('fqdn')
 
-    if (!onboardUrl && !fqdn) {
+    if (!onboardUrl && !onboardedRedirection && !fqdn) {
```
Ldoppea, line 32: I'm not sure this check should be added. If we don't have `onboardUrl` nor `fqdn`, then there would be no sense to have an `onboardedRedirection`.

If the deep link contains only `onboardedRedirection` but no `onboardUrl` nor `fqdn`, then the app would break as it would try to initiate onboarding without enough params.

So I would return `undefined` in that case (so I would keep the old check)

src/libs/functions/formatOnboardingRedirection.ts
```diff
@@ -0,0 +1,25 @@
+import CozyClient, {
+  generateWebLink,
+  deconstructRedirectLink
+} from 'cozy-client'
+
+export const formatOnboardingRedirection = (
```
Ldoppea, line 6: Is there a reason why we implement this method here and not in cozy-client like you did for `deconstructRedirectLink`?

src/hooks/useAppBootstrap.js
```diff
@@ -90,10 +99,13 @@ export const useAppBootstrap = client => {
       const onboardingParams = parseOnboardingURL(url)
 
       if (onboardingParams) {
-        const { onboardUrl, fqdn } = onboardingParams
+        const { onboardUrl, onboardedRedirection, fqdn } = onboardingParams
```
Ldoppea, line 102: This should be also done on cold boot (here you are in URL interception when app is already running). Cold boot part is in the first useEffect

Ldoppea (commented):
> I am not very happy about the result especially because I added code in a lot of different places. Feel free to challenge it. What do you think about the onboardingRedirection state used to pass the redirection between onboarding and the useAppBoostrap hook for example?

I'm wondering if we can keep this info only in `useAppBootstrap` as it is responsible for the initial URL parsing AND for the final redirection. I have an intuition on how to handle this, but I'm not sure if I have all scenario in mind.

We can do a call to discuss about this if you want to

src/hooks/useAppBootstrap.functions.ts
zatteo: 100% correct.

zatteo:
> > I am not very happy about the result especially because I added code in a lot of different places. Feel free to challenge it. What do you think about the onboardingRedirection state used to pass the redirection between onboarding and the useAppBoostrap hook for example?
> 
> I'm wondering if we can keep this info only in `useAppBootstrap` as it is responsible for the initial URL parsing AND for the final redirection. I have an intuition on how to handle this, but I'm not sure if I have all scenario in mind.
> 
> We can do a call to discuss about this if you want to

I updated like you said, it works very smoothly ! It works with initial link and subscription link.

When we are already connected and we open an onboarding link, what should we do (whatever the _onboarded_redirection_ params) ? At the moment
- when app is running and we open an onboarding link, it redirects to the onboarding path
- when app is not running and we open an onboarding link, it ignores the onboarding path and just log the user (I think it is the good behavior)

src/libs/functions/formatOnboardingRedirection.ts
zatteo: Not really. But maybe we can keep it here because it is more high level.

src/hooks/useAppBootstrap.js
```diff
@@ -72,9 +91,7 @@ export const useAppBootstrap = client => {
   useEffect(() => {
     if (initialRoute !== 'fetching' && isLoading) {
       setIsLoading(false)
-      if (initialRoute.route !== routes.home) {
```
Ldoppea, line 75: Why did you need to remove this?

src/hooks/useAppBootstrap.js
```diff
@@ -50,6 +57,18 @@ export const useAppBootstrap = client => {
             route: routes.welcome
           })
         }
+      } else if (OnboardingRedirection) {
+        const onboardingRedirectionURL = formatOnboardingRedirection(
+          OnboardingRedirection,
+          client
+        )
+
+        return setInitialRoute({
```
Ldoppea, line 66: Maybe we should clear `OnboardingRedirection` or it will be used if the user logs out and tries to login into another Cozy.

Ldoppea (approved):
Wayyy cleaner 👍

src/hooks/useAppBootstrap.js
Ldoppea: I'm wondering about the side effects, so please double check you didn't break the splashscreen management that should happens only after app and konnector tiles are loaded on the home.

src/hooks/useAppBootstrap.js
zatteo: Done

src/hooks/useAppBootstrap.js
zatteo: I agree it makes little sense to remove it. I do not remember why exactly I removed this. Sometime splash screen did not hide. But I tried again and it works with it so I add it back.

PR #636, opened 2023-03-03
feat: Create a clisk timeout trigger

doubleface:
This trigger is created when a clisk konnector is created and is
scheduled to run the cliskTimeout service 20min after the creation
of the job and will force the termination of the job

This trigger will be automatically removed when the job is stopped by
the flagship application

Depends on https://github.com/cozy/cozy-home/pull/1897 to be deployed on home application

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android
* [X] All changes have test coverage

### DIFF
--- a/src/libs/Launcher.js
+++ b/src/libs/Launcher.js
@@ -11,6 +11,8 @@ import { ensureKonnectorFolder } from './folder'
 import { saveCredential, getCredential, removeCredential } from './keychain'
 import { dataURItoArrayBuffer } from './utils'
 
+import { constants } from '/screens/konnectors/constants/konnectors-constants'
+
 const log = Minilog('Launcher')
 
 /**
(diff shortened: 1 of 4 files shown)


src/libs/Launcher.js
```diff
@@ -368,6 +370,52 @@ export default class Launcher {
...
+    if (!job) {
+      throw new Error('createTimeoutTrigger: no job found in context')
+    }
+
+    const { data: timeoutTrigger } = await client.save({
+      _type: 'io.cozy.triggers',
+      worker: 'service',
+      type: '@at', // did not use @in because only @at triggers remove themselves after use
+      arguments: new Date(
+        Date.now() + constants.serviceTimeoutDuration
```
Ldoppea, line 391: Is is possible that the device clock is not in sync with the server? (I'm wondering if the SSL certificates would throw before this step or not)

src/libs/Launcher.js
```diff
@@ -368,6 +370,52 @@ export default class Launcher {
...
+
+  /**
+   * Removes the timeout trigger when the job is already done
+   *
+   * @returns {Promise<void>}
+   */
+  async removeTimeoutTrigger() {
+    const { client, timeoutTrigger } = this.getStartContext() || {}
+    if (timeoutTrigger) {
+      await client.destroy(timeoutTrigger)
```
Ldoppea, line 415: Can this throw if `timeoutTrigger` is not found on cozy-stack?

src/libs/Launcher.js
doubleface: I used the same code as the one in banks : https://github.com/cozy/cozy-banks/blob/e39fb584e39917a5329d79762a6f16c73d79e105/src/targets/services/konnectorAlerts/createTriggerAt.js#L75

But I think if the device clock is out of sync, it won't work as expected.

No idea about the ssl certificates

src/libs/Launcher.js
doubleface: Yes it could. I added try catch arount it

Ldoppea (approved):
I think we can merge the PR even if we are not sure about the clock offset question. I asked about it to the team, so based on the answer we will create another PR.

src/libs/Launcher.js
doubleface: Yes it would, but @in triggers do not remove themselves after run and that is why I used @at

PR #640, opened 2023-03-03
feat: Removing all konnectors and remaining tools, libs.

LucsT:
This is cleaning all konnectors things in the app.

- Konnectors sources (we still have git for a 2-3 remaining not migrated konnectors)
- Script for build
- Many associated libs
- ContentScript (now in clisk)

Please double check that i did not miss a lib. Thank's.


yarn.lock
```diff
@@ -11183,7 +11149,7 @@ klona@^2.0.4:
   resolved "https://registry.yarnpkg.com/klona/-/klona-2.0.4.tgz#7bb1e3affb0cb8624547ef7e8f6708ea2e39dfc0"
   integrity sha512-ZRbnvdg/NxqzC7L9Uyqzf4psi1OM4Cuc+sJAkQPjO6XkQIJTNbfK2Rsmbw8fx1p2mkZdp2FZYo2+LwXYY/uwIA==
 
-ky@0.25.1, ky@^0.25.1:
+ky@^0.25.1:
```
LucsT, line 11152: I'm not sur about that for exemple. But seems hoisted by cozy-clisk only. So seems good to me.

Ldoppea (approved):

yarn.lock
Ldoppea: You can check this by doing a `yarn why ky`

But seems good to me too :)

doubleface (approved):

PR #647, opened 2023-03-06
feat: remove placeholder text

acezard:
Just display an empty view instead.
The "loading" text was clashing with the backdrop overlay and created a visual artifact.
You can see the bug it created on the following image:
image

### DIFF
--- a/src/screens/konnectors/LauncherView.js
+++ b/src/screens/konnectors/LauncherView.js
@@ -200,11 +200,7 @@ class LauncherView extends Component {
               />
             </View>
           </>
-        ) : (
-          <View>
-            <Text>Loading...</Text>
-          </View>
-        )}
+        ) : null}
       </>
     )
   }


acezard:
> I don't see the "loading..." text displayed in this screenshot. Surely I didn't understand everything, so if I rephrase: We were displaying "loading..." even with the "Connection in progress" stuff?
> 
> But what I see from this screenshot is that a wrong alignment between the modal, StatusBar and bottom bar

yeah not sure why we don't see the loading text in my screenshot. But that's what was displayed indeed. So in my screenshot case I'm not sure what happened, I guess I emptied the text, took a screenshot, then removed the <Text>

src/screens/konnectors/LauncherView.js
```diff
@@ -201,9 +201,7 @@ class LauncherView extends Component {
             </View>
           </>
         ) : (
-          <View>
-            <Text>Loading...</Text>
-          </View>
+          <View />
```
Crash--, line 204: can't we just return null instead of a view?

Ldoppea:
> ps: also yes, there is an issue in dev mode (localhost) where top/bottom margin values are wrong. I don't remember why, it's an old issue and since it doesn't happen in the actual build we didn't solve it

This may be related to the fact that safe-area injection is not done if you set `devConfig.disableGetIndex = true` (as we cannot intercept the index.html content.

acezard:
> > ps: also yes, there is an issue in dev mode (localhost) where top/bottom margin values are wrong. I don't remember why, it's an old issue and since it doesn't happen in the actual build we didn't solve it
> 
> This may be related to the fact that safe-area injection is not done if you set `devConfig.disableGetIndex = true` (as we cannot intercept the index.html content.

yes I believe this is the case indeed

src/screens/konnectors/LauncherView.js
acezard: It makes the most sense I agree, I'm updating it

Crash-- (approved):

PR #649, opened 2023-03-07
feat: Force stop the current job id if any on flagship app start

doubleface:
The current job id is saves in the redux store when the job is stated by the flagship app and removed on the end of the clisk konnector execution.

If the job id is still in redux store when reloading the flagship app, it means that the flagship app failed while running the clisk konnector job. The flagship app then mark the current job as "context deadline exceeded" which is the error code for timeout job error on cozy stack.

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/libs/konnectors/cleanKonnectorsOnBoot.ts
+++ b/src/libs/konnectors/cleanKonnectorsOnBoot.ts
@@ -4,9 +4,11 @@ import CozyClient from 'cozy-client'
 
 import { getErrorMessage } from '/libs/functions/getErrorMessage'
 import { sendKonnectorsLogs } from '/libs/konnectors/sendKonnectorsLogs'
+import { stopKonnectorJob } from '/libs/konnectors/stopKonnectorJob'
 import {
   CurrentKonnectorState,
-  setCurrentRunningKonnector
+  setCurrentRunningKonnector,
+  setCurrentRunningKonnectorJobId
 } from '/redux/KonnectorState/CurrentKonnectorSlice'
 import { store } from '/redux/store'
 
@@ -20,6 +22,12 @@ const getRunningKonnector = (state: {
   return state.currentKonnector.currentRunningKonnector
 }
 
+const getRunningKonnectorJobId = (state: {
+  currentKonnector: CurrentKonnectorState
+}): string | undefined => {
+  return state.currentKonnector.currentRunningKonnectorJobId
+}
+
 /**
  * Check for still-running konnectors, clean them and
  * send remaining konnectors' logs on cozy-stack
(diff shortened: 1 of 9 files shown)


src/libs/konnectors/cleanKonnectorsOnBoot.ts
```diff
@@ -38,6 +47,14 @@ export const cleanKonnectorsOnBoot = async (
     store.dispatch(setCurrentRunningKonnector())
   }
 
+  if (runningKonnectorJobId !== undefined) {
+    log.warn(
+      `Stopping the job ${runningKonnectorJobId} as context deadline exceeded`
```
Ldoppea, line 52: Should we use something like `as the connector was not cleaned on previous app session` instead of `as context deadline exceeded`? In my point of view we are not really in a "timeout" scenario here.

src/libs/konnectors/stopKonnectorJob.ts
```diff
@@ -0,0 +1,61 @@
...
+    log.error('Error while trying to stop job ${jobId}')
+    log.error(e)
+  }
+}
+
+/**
+ * Mark the given job as "context deadline exceeded"
+ */
+const stopJob = (client: CozyClient, jobId: string): Promise<unknown> => {
+  return client.save({
```
Ldoppea, line 26: Don't we need to check if the job is still running on cozy-stack first? What happens if we do this after the stack run the cleanup service?

src/libs/konnectors/stopKonnectorJob.ts
```diff
@@ -0,0 +1,61 @@
...
+): Promise<void> => {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const { data: triggers } = await client.query(
+    Q('io.cozy.triggers').where({
+      worker: 'service',
+      type: '@at',
+      'message.fields.cliskJobId': jobId
+    })
+  )
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
Ldoppea, line 54: I think we can remove those ESLint errors by typing the `io.cozy.triggers` output.

Something like:
```ts
interface IoCozyTrigger {
    _id: string
}
interface IoCozyTriggerResult {
    data: IoCozyTrigger[]
}

//...

const { data: triggers } = await client.query(
  Q('io.cozy.triggers').where({
    worker: 'service',
    type: '@at',
    'message.fields.cliskJobId': jobId
  })
) as IoCozyTriggerResult
```

Ldoppea (approved):
Looks good to me, modulo your answer on the "cozy-stack check" question 👍

src/libs/konnectors/stopKonnectorJob.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Mark the given job as "context deadline exceeded"
+ */
+const stopJob = (client: CozyClient, jobId: string): Promise<unknown> => {
+  return client.save({
+    _type: 'io.cozy.jobs',
+    _id: jobId,
+    _rev: true, // to force an update on Job collection
```
Crash--, line 29: same as here https://github.com/cozy/cozy-home/pull/1897#discussion_r1130794135 

we should not do that and pass the real rev 🙏

src/libs/konnectors/stopKonnectorJob.ts
doubleface: We can. I was trying to avoid to get the job. But since I will need to get the job to pass it's real revision, I will be able to check this too.

If we do this after the cleanup service of the stack, the last one to update the job will win.

src/libs/konnectors/stopKonnectorJob.ts
doubleface: I was trying to avoid to get the job for performance reason. But let's do the get

PR #650, opened 2023-03-07
feat: Wait for CozyApp to be loaded before loading HomeView if DeepLink

Ldoppea:
When opening the app from a deep link that target a CozyApp, the current implementation start loading the HomeView and the CozyAppScreen simultaneously

This is not optimal as we want to display CozyAppScreen as fast as possible. The HomeView only needs to be loaded when the user clicks on the Home button or the Back button

So we want to start loading the HomeView only after the CozyAppScreen finished loading

### DIFF
--- a/src/screens/home/components/HomeView.js
+++ b/src/screens/home/components/HomeView.js
@@ -19,6 +19,7 @@ import { useSession } from '/hooks/useSession'
 import { routes } from '/constants/routes'
 import { navigate } from '/libs/RootNavigation'
 import { getData, StorageKeys } from '/libs/localStore/storage'
+import { useHomeStateContext } from '/screens/home/HomeStateProvider'
 
 const unzoomHomeView = webviewRef => {
   webviewRef?.injectJavaScript(
@@ -42,6 +43,7 @@ let hasRenderedOnce = false
 const HomeView = ({ route, navigation, setLauncherContext, setBarStyle }) => {
   const client = useClient()
   const [uri, setUri] = useState('')
+  const { shouldWaitCozyApp, setShouldWaitCozyApp } = useHomeStateContext()
   const [trackedWebviewInnerUri, setTrackedWebviewInnerUri] = useState('')
   const nativeIntent = useNativeIntent()
   const session = useSession()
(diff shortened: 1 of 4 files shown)


src/screens/home/components/HomeView.js
```diff
@@ -169,10 +172,12 @@ const HomeView = ({ route, navigation, setLauncherContext, setBarStyle }) => {
             href: cozyAppFallbackURL,
             slug
           })
+        } else {
+          if (shouldWaitCozyApp === undefined) setShouldWaitCozyApp(false)
```
zatteo, line 176: What is the purpose of setting shouldWaitCozyApp to undefined by default ?

zatteo (approved):

Crash--:
> The HomeView only needs to be loaded when the user clicks on the Home button or the Back button

Not agreed with that. We want the navigation between the CozyApp and the "HomeView" as fast as possible even for the first click since this is the only way we have to navigate throught your cozy. 

The HomeView should be loaded as soon as we can. So after the loading (onLoadEnd) of the CozyApp?

src/screens/home/components/HomeView.js
Ldoppea: This is because a render can be done before the first `useEffect` execution. If defaulted to `false` the WebView would render without waiting for the `useEffect` to set the `shouldWaitCozyApp`.

Ldoppea:
> > The HomeView only needs to be loaded when the user clicks on the Home button or the Back button
> 
> Not agreed with that. We want the navigation between the CozyApp and the "HomeView" as fast as possible even for the first click since this is the only way we have to navigate throught your cozy.
> 
> The HomeView should be loaded as soon as we can. So after the loading (onLoadEnd) of the CozyApp?

This is what is done. By `the HomeView only needs to be loaded when the user clicks` I meant that in needs **to be loaded** and not **to start loading**.

This commit's implementation triggers the HomeView rendering just after the `onLoadEnd` (cf this [line](https://github.com/cozy/cozy-flagship-app/pull/650/files#diff-321301bbb3002ce9f236dc4e373c08558b0ca01a83053acaa98ff6d6db1b6a0cR71)) which means that when the users clicks on the Home button:
- either the Home finished loaded
- or the Home would be still loading (if the user clicks on the Home button fast enough)

Crash-- (approved):

PR #652, opened 2023-03-08
Context for ccc

KillianCourvoisier:
With @Ldoppea , we find a bug with the recent update regarding the account creation.
In this PR we fix it by giving the rest of the context in addition to the destructurate variables from the startContext.
We spot that an account is needed to save the user's credentials, so it add a little condition returning null if no accounts are present in the startContext at the time getCredentials is called.

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/libs/Launcher.js
+++ b/src/libs/Launcher.js
@@ -209,15 +209,8 @@ export default class Launcher {
   async ensureAccountTriggerAndLaunch() {
     const result = {}
     const startContext = this.getStartContext()
-    let {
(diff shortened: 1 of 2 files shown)


Crash--:
Can't we add tests for that?

doubleface:
J'ai ajouté un test unitaire, puisque c'est moi qui ai raté cette régression

Crash-- (approved):

PR #653, opened 2023-03-08
Feat/open cozy app at flagship opening

zatteo:
Manage _default_redirection_ from instance settings to open a Cozy app different from home when you open the app : 
- fetch it [from stack](https://docs.cozy.io/en/cozy-stack/settings/#get-settingsinstance)
- store it in AsyncStorage
- fetch it from AsyncStorage
- use it in useAppBootstrap according to the product request to redirect when you open the app

To test, just add _"default_redirection": "drive/#/folder"_ in your io.cozy.settings.instance database.

### DIFF
--- a/src/hooks/useAppBootstrap.js
+++ b/src/hooks/useAppBootstrap.js
@@ -14,7 +14,10 @@ import {
 } from '/hooks/useAppBootstrap.functions'
 import { useSplashScreen } from '/hooks/useSplashScreen'
 import { formatRedirectLink } from '/libs/functions/formatRedirectLink'
-import { getOrFetchDefaultRedirectionUrl } from '/libs/defaultRedirection/defaultRedirection'
+import {
+  getOrFetchDefaultRedirectionUrl,
+  getParamsWithDefaultRedirectionUrl
+} from '/libs/defaultRedirection/defaultRedirection'
 
 let OnboardingRedirection = ''
 
@@ -90,10 +93,7 @@ export const useAppBootstrap = client => {
 
         return setInitialRoute({
           route: routes.home,
-          params: {
-            mainAppFallbackURL: undefined,
-            cozyAppFallbackURL: defaultRedirectUrl
-          }
+          params: getParamsWithDefaultRedirectionUrl(defaultRedirectUrl, client)
         })
       }
     }
--- a/src/hooks/useAppBootstrap.js
+++ b/src/hooks/useAppBootstrap.js
@@ -14,6 +14,7 @@ import {
 } from '/hooks/useAppBootstrap.functions'
 import { useSplashScreen } from '/hooks/useSplashScreen'
 import { formatOnboardingRedirection } from '/libs/functions/formatOnboardingRedirection'
+import { getOrFetchDefaultRedirectionUrl } from '/libs/defaultRedirection/defaultRedirection'
 
 let OnboardingRedirection = ''
 
@@ -75,12 +76,23 @@ export const useAppBootstrap = client => {
         const payload = await Linking.getInitialURL()
         const { mainAppFallbackURL, cozyAppFallbackURL } =
           parseFallbackURL(payload)
+        if (mainAppFallbackURL || cozyAppFallbackURL) {
+          return setInitialRoute({
+            route: routes.home,
+            params: {
+              mainAppFallbackURL,
+              cozyAppFallbackURL
+            }
+          })
+        }
+
+        const defaultRedirectUrl = await getOrFetchDefaultRedirectionUrl(client)
 
         return setInitialRoute({
           route: routes.home,
           params: {
-            mainAppFallbackURL,
-            cozyAppFallbackURL
+            mainAppFallbackURL: undefined,
+            cozyAppFallbackURL: defaultRedirectUrl
           }
         })
       }
--- a/src/hooks/useAppBootstrap.js
+++ b/src/hooks/useAppBootstrap.js
@@ -13,7 +13,7 @@ import {
   parseOnboardingURL
 } from '/hooks/useAppBootstrap.functions'
 import { useSplashScreen } from '/hooks/useSplashScreen'
-import { formatOnboardingRedirection } from '/libs/functions/formatOnboardingRedirection'
+import { formatRedirectLink } from '/libs/functions/formatRedirectLink'
 import { getOrFetchDefaultRedirectionUrl } from '/libs/defaultRedirection/defaultRedirection'
 
 let OnboardingRedirection = ''
@@ -59,7 +59,7 @@ export const useAppBootstrap = client => {
           })
         }
       } else if (OnboardingRedirection) {
-        const onboardingRedirectionURL = formatOnboardingRedirection(
+        const onboardingRedirectionURL = formatRedirectLink(
           OnboardingRedirection,
           client
         )
--- a/src/libs/defaultRedirection/defaultRedirection.spec.ts
+++ b/src/libs/defaultRedirection/defaultRedirection.spec.ts
@@ -8,13 +8,13 @@ import {
   NAVIGATION_APP_SLUG,
   DEFAULT_REDIRECTION_DELAY_IN_MS,
   InstanceSettings,
-  isDefaultRedirectionUrlNavigationApp,
   fetchDefaultRedirectionUrl,
   setDefaultRedirectionUrl,
   getDefaultRedirectionUrl,
   fetchAndSetDefaultRedirectionUrl,
   fetchAndSetDefaultRedirectionUrlInBackground,
-  getOrFetchDefaultRedirectionUrl
+  getOrFetchDefaultRedirectionUrl,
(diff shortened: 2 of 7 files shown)


src/libs/defaultRedirection/defaultRedirection.ts
```diff
@@ -0,0 +1,131 @@
...
+  data: {
+    attributes: {
+      default_redirection: string
+    }
+  }
+}
+
+export const isDefaultRedirectionUrlNavigationApp = (
+  defaultRedirectionUrl: string,
+  client: CozyClient
```
Crash--, line 30: Does the CozyClient import at the beginning of the file mandatory to be able to type this arg?

src/libs/defaultRedirection/defaultRedirection.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export const fetchDefaultRedirectionUrl = async (
+  client: CozyClient
+): Promise<string | undefined> => {
+  let defaultRedirectionUrl
+
+  try {
+    const settingsCollection = client.collection(
+      'io.cozy.settings'
+    ) as SettingsCollection
```
Crash--, line 50: You should almost never use the `collection` directly. `client.query()` always. 

`client.query(Q('io.cozy.settings').getById('instance'))` should do the trick ;)

Crash--: Also, client.query should always have a fectchPolicy and be named. By default, the getById will be named automatically by cozy-client itself, but it's a good habit to do it: 
`as: io.cozy.settings/instance`

Also the settings of an instance should not be often changed, so you can use a long policy ;)

src/libs/functions/formatRedirectLink.spec.ts
```diff
@@ -1,7 +1,8 @@
 import CozyClient from 'cozy-client'
-import { formatOnboardingRedirection } from './formatOnboardingRedirection'
 
-describe('formatOnboardingRedirection', () => {
+import { formatRedirectLink } from './formatRedirectLink'
```
Crash--, line 3: from /libs/.. ?

src/libs/functions/formatRedirectLink.ts
```diff
@@ -3,7 +3,7 @@ import CozyClient, {
   deconstructRedirectLink
 } from 'cozy-client'
 
-export const formatOnboardingRedirection = (
+export const formatRedirectLink = (
```
Crash--, line 6: should be put in cozy-client?

zatteo: Second times we used this function, second person that talk about putting it in cozy-client => let's go

src/hooks/useAppBootstrap.js
```diff
@@ -75,12 +76,23 @@ export const useAppBootstrap = client => {
...
+
+        const defaultRedirectUrl = await getOrFetchDefaultRedirectionUrl(client)
 
         return setInitialRoute({
           route: routes.home,
           params: {
-            mainAppFallbackURL,
-            cozyAppFallbackURL
+            mainAppFallbackURL: undefined,
+            cozyAppFallbackURL: defaultRedirectUrl
```
Ldoppea, line 95: If `getOrFetchDefaultRedirectionUrl` returns `home` then we should set the `mainAppFallbackURL` instead.

src/libs/defaultRedirection/defaultRedirection.ts
Ldoppea: Yes it is, but my understanding is that `tsc` would remove that import at build

src/libs/defaultRedirection/defaultRedirection.ts
```diff
@@ -0,0 +1,131 @@
...
+    )
+      return
+
+    return localDefaultRedirectionUrl
+  } else {
+    const newDefaultRedirectionUrl = await fetchDefaultRedirectionUrl(client)
+
+    if (!newDefaultRedirectionUrl) return
+
+    await setDefaultRedirectionUrl(newDefaultRedirectionUrl)
```
Ldoppea, line 122: Here you can reuse `fetchAndSetDefaultRedirectionUrl()` (if you add a return to it)

Before
```ts
const newDefaultRedirectionUrl = await fetchDefaultRedirectionUrl(client)

if (!newDefaultRedirectionUrl) return

await setDefaultRedirectionUrl(newDefaultRedirectionUrl)
```

after
```ts
const newDefaultRedirectionUrl = await fetchAndSetDefaultRedirectionUrl(client)
```

Ldoppea: And after that, the code would be pretty similar between the `if` and the `else`, so you can condition only the `fetchAndSetDefaultRedirectionUrlInBackground` vs `fetchDefaultRedirectionUrl` and do the `if (isDefault...)` and the `return` outside of the if

src/libs/defaultRedirection/defaultRedirection.ts
zatteo: It seems so. I will ask confirmation.

src/libs/defaultRedirection/defaultRedirection.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export const isDefaultRedirectionUrlNavigationApp = (
+  defaultRedirectionUrl: string,
+  client: CozyClient
+): boolean => {
+  const subdomainType = client.capabilities.flat_subdomains ? 'flat' : 'nested'
+
+  const { slug } = deconstructCozyWebLinkWithSlug(
+    defaultRedirectionUrl,
+    subdomainType
```
Ldoppea, line 32: Note for another PR: we re-write this `flat_subdomains ? 'flat' : 'nested'` thing everywhere, we definitively needs a helper to convert `flat_domain:boolean` into `subdomainType:'flat'|'nested'`

src/hooks/useAppBootstrap.js
Ldoppea: After reading the `getOrFetchDefaultRedirectionUrl()` method I understand that it would return `undefined` if `home` is default. So you implementation would work.

But someone said that the default url may contain some `path`. Is that still true? If yes it seems like the current implementation would not handle something like `home/#/connected/alan`

src/libs/defaultRedirection/defaultRedirection.ts
zatteo: I am not sure about the long policy because this request will be used :
- the first time so we need fresh data
- when we open the app 10s later to update the value if the user updated it. So with a long policy the user will need to wait before seeing the change even after an app reload

src/libs/defaultRedirection/defaultRedirection.ts
zatteo: I TOTALLY agree. I noted it for myself yesterday. I am more thinking on cozy-client side because it also happens on front app. Something like client.getSubdomainType => return 'flat' or 'nested' (no hook no async). @Crash-- what do you think ?

src/libs/defaultRedirection/defaultRedirection.ts
Crash--: > the first time so we need fresh data

The first time, you'll not have fetched data, so the fetch policy will ne be applied. 

> when we open the app 10s later to update the value if the user updated it. So with a long policy the user will need to wait before seeing the change even after an app reload

Oh you right, I forgot about `to update the value if the user updated it` the fact that the user will be able to update the value.

src/libs/defaultRedirection/defaultRedirection.ts
zatteo: Update: `import type CozyClient from 'cozy-client'`

src/hooks/useAppBootstrap.js
zatteo: Right!  I did not handle the case of a home with a path default_redirection. Fixed.

Ldoppea (approved):

src/libs/defaultRedirection/defaultRedirection.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+jest.mock('@react-native-async-storage/async-storage')
+jest.mock('/libs/functions/formatRedirectLink')
+
+const mockedFormatRedirectLink = formatRedirectLink as jest.MockedFunction<
+  typeof formatRedirectLink
+>
+
+const NAVIGATION_APP_URL = `http://${NAVIGATION_APP_SLUG}.mycozy.test/#/alan`
+const DRIVE_URL = 'http://drive.mycozy.test/#/folder'
+const CONTACT_URL = 'http://contacts.mycozy.test/#/'
```
Crash--, line 30: Do ou only test nested domain? Can you add a test for a flat one? 
```
https://contacts-foo.mycozy.cloud
```

zatteo: Right. I added tests in the function checking flat/nested.

PR #662, opened 2023-03-13
fix: Use @in trigger for CliskTimeout service

doubleface:
This avoids problems when client and service clocks are not synchronized
and it is now confirmed that @in triggers are well cleaned after run (on
redis)

### DIFF
--- a/src/libs/Launcher.js
+++ b/src/libs/Launcher.js
@@ -387,10 +387,8 @@ export default class Launcher {
     const { data: timeoutTrigger } = await client.save({
       _type: 'io.cozy.triggers',
       worker: 'service',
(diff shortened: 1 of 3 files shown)


Ldoppea (approved):

Crash--:
@doubleface https://github.com/cozy/cozy-flagship-app/blob/1ea43cf2000ee81a4a4b7745d4f147184fa52643/src/libs/konnectors/stopKonnectorJob.ts#L56 should also be updated at least

doubleface:
Thanks @Crash-- Here is the fix : https://github.com/cozy/cozy-flagship-app/pull/663

PR #667, opened 2023-03-15
feat: Make WelcomeScreen neutral

Ldoppea:
> **Note**
> Due to its size, this PR is better to be read commit by commit

Now that we support OnboardingPartner scenario we want to have a neutral WelcomeScreen when no OnboardingPartner info is detected

The new neutral screen contains 2 buttons instead of one: `CRÉER`, and `J'AI DÉJÀ UN CLOUD PERSONNEL` .

This PR implements those two scenario by splitting the Cloudery WebView into 2 different ones, one for the Login scenario and the other for the Sign-in scenario

In order to keep the interface reactive, we want to keep the preloading feature that was implemented prior to that change. So both WebView are preloaded and the WelcomeScreen should give the expected scenario (`CLOUDERY_MODE_LOGIN` or `CLOUDERY_MODE_SIGNING`) to be displayed. Then we display the correct WebView by modifying their respective
z-index

If an OnboardingPartner context is detected, only the Login WebView is displayed with the OnboardingPartner url. Even if its not only about Login scenario, as the scenario choice is handled by the Cloudery, we made an arbitrary choice to simplify the code by not adding a third WebView  

The navigation behavior (back button management) has been kept as prior but adapted to react to the displayed WebView

The focus behavior has been kept as prior but adapted to react to the displayed WebView. Note that in the Sign-in scenario we expect to find a `postcode` field to focus

The overlay now waits for both WebView to be loaded before hiding itself

> **Warning**
> The Sign-in route is not implemented yet on the Cloudery, so clicking on the `CRÉER` button would display a 404 page

image


Crash-- (approved):
Look nice 👍 . Great refacto, nice commit.

I'm just wondering if we should merge this into master since it breaks a few things until we have it implemented.

PR #669, opened 2023-03-17
fix: use local fonts in html webviews

acezard:
Using fonts from device instead of user instance on unlogged (or potentially unlogged) views.

Inspired from this SO post https://stackoverflow.com/a/67409099/4817704

### DIFF
--- a/src/components/getLocalFonts.ts
+++ b/src/components/getLocalFonts.ts
@@ -1,0 +1,37 @@
+import { Platform } from 'react-native'
+
+export const getLocalFonts = (): string => {
+  const fontUrlRegular =
+    Platform.select({
+      ios: 'Lato-Regular.ttf',
+      android: 'file:///android_asset/fonts/Lato-Regular.ttf'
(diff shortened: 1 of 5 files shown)


src/components/getLocalFonts.ts
```diff
@@ -0,0 +1,35 @@
+import { Platform } from 'react-native'
+
+export const getLocalFonts = (): string => {
+  const fontUrlRegular = Platform.select({
+    ios: 'Lato-Regular.ttf',
+    android: 'file:///android_asset/fonts/Lato-Regular.ttf'
+  })
```
Ldoppea, line 7: nit: putting the nullish coalescing operator here instead of in the templated return would make it easier to read and would be more type safe.
 
```suggestion
  }) ?? ''
```

Ldoppea (approved):

src/components/getLocalFonts.ts
acezard: Agree

PR #680, opened 2023-03-20
fix(iOS): Download file from Intent

Crash--:
We can display a Drive Intent from Cozy-Banks
by clicking on the chip associated to the transactions.

This Drive intent shows the Viewer and this viewer has a "Download" button. This button was not working because for iOS we have to make some work in order to be download files, but we only did it for the
main frame. And since Intents are not in the main
frame, it was not working for them.

I fixed this issue by checking the link for every frames.

#### Checklist

* [X] Tested on iOS

### DIFF
--- a/src/components/webviews/ReloadInterceptorWebView.js
+++ b/src/components/webviews/ReloadInterceptorWebView.js
@@ -45,7 +45,14 @@ const interceptNavigation = ({
   client,
   setDownloadProgress
 }) => {
-  if (Platform.OS === 'ios' && !initialRequest.isTopFrame) {
+  const isPreviewableLink = checkIsPreviewableLink(initialRequest.url, client)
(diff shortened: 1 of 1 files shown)


src/components/webviews/ReloadInterceptorWebView.js
```diff
@@ -45,7 +45,14 @@ const interceptNavigation = ({
   client,
   setDownloadProgress
 }) => {
-  if (Platform.OS === 'ios' && !initialRequest.isTopFrame) {
+  const isPreviewableLink = checkIsPreviewableLink(initialRequest.url, client)
+  // We don't want to intecerpt iframe navigation excepts for iOS and the download
```
zatteo, line 49: nit: intercept

zatteo (approved):

PR #684, opened 2023-03-21
fix: Fixing metadata deduplication by upgrading cozy-clisk

LucsT:
Upgrading cozy-clisk only


doubleface (approved):

LucsT:
Adding 0.9.4 cozy-clisk too, Fixing worker errors

PR #687, opened 2023-03-22
Fix/handle default redirection error

zatteo:
Fix to avoid errors when a redirect link (default redirect, onboarding redirection) is invalid.

Now, it also allows to run app even if the stack has not been updated with the fix of invalid default redirection.

### DIFF
--- a/src/libs/functions/formatRedirectLink.spec.ts
+++ b/src/libs/functions/formatRedirectLink.spec.ts
@@ -10,15 +10,19 @@ describe('formatRedirectLink', () => {
     })
   } as CozyClient
 
-  it('should format onboarding redirection', () => {
+  it('should format redirect link with hash', () => {
     expect(formatRedirectLink('contacts/#/hash', client)).toStrictEqual(
       'http://contacts.mycozy.test/#/hash'
     )
   })
 
-  it('should format onboarding redirection', () => {
+  it('should format redirect link with path and hash', () => {
     expect(formatRedirectLink('contacts/path/#/hash', client)).toStrictEqual(
       'http://contacts.mycozy.test/path/#/hash'
     )
+  })
+
+  it('should return null when redirect link is invalid', () => {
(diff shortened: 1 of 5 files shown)


src/libs/functions/formatRedirectLink.ts
```diff
@@ -4,22 +4,26 @@ import CozyClient, {
...
+    return generateWebLink({
+      cozyUrl,
+      subDomainType,
+      slug,
+      pathname,
+      hash,
+      searchParams: []
+    })
+  } catch {
+    return null
```
Ldoppea, line 27: Would it be possible to `log.error()` here? So it would be sent to Sentry

Note that we use `getErrorMessage` in this project to ensure we send correct error message to sentry

Example: https://github.com/cozy/cozy-flagship-app/blob/d62f13e7377c40db68c4219be2657c0ff389ac87/src/hooks/useAppBootstrap.functions.ts#L77-L83

src/libs/functions/formatRedirectLink.spec.ts
```diff
@@ -10,13 +10,13 @@ describe('formatRedirectLink', () => {
...
 
-  it('should format onboarding redirection', () => {
+  it('should format redirect link', () => {
     expect(formatRedirectLink('contacts/#/hash', client)).toStrictEqual(
       'http://contacts.mycozy.test/#/hash'
     )
   })
 
-  it('should format onboarding redirection', () => {
+  it('should format redirect link', () => {
```
Ldoppea, line 19: nit: can you add a unit test with the "wrong format" that we currently receive from the stack? This would make a good documentation of the current bug

Ldoppea (approved):

src/libs/functions/formatRedirectLink.ts
zatteo: Sure!

PR #695, opened 2023-03-24
feat: Use redirectLink in notification instead of separate values

zatteo:
Before, we received slug, pathname and hash from notification data. But it is exactly the purpose of the redirectLink to carry thes values in one human readable value. So here we go.


zatteo:
I checked and I received emails with correct links. Except that they are always flat but it is not related to my PR. Should I create an issue or an card about this ?

Otherwise it seems good. Emails do not use notification data.

Example of an email : 
- https://github.com/cozy/cozy-banks/blob/09d7594564dce2863866834633e8a3e3e259afc6/src/ducks/notifications/BalanceGreater/template.hbs#L41
- https://github.com/cozy/cozy-banks/blob/09d7594564dce2863866834633e8a3e3e259afc6/src/ducks/notifications/BaseNotificationView.js#L18

Crash-- (approved):

PR #697, opened 2023-03-24
fix: worker reload

doubleface:
All clisk konnectors began to fail when waiting for a worker event (ex waitForElementInWorker).

It looks like when the worker changes url, the flagship app could not restart the worker execution anymore and the konnector execution would wait indefinitely.

I could link this fail with any flagship app update or any other environment change but when looking at the exchanges between the worker and the launcher, I could not know how it would work anyway.

The worker restart was conditioned with worker.workerReady launcherView state and this state could not be changed without the worker restart.

This is why I removed the workerReady state and all code related. We already have the NEW_WORKER_INITIALIZING event, launched by the worker (outside of post-me), which can trigger the reload of the worker and it works well when I test it on multiple konnectors.

Since the sequence of calls was not clear for us when trying to debug the problem, I also added a sequence diagram in the `docs` directory. 

#### Checklist

* [X] Tested on iOS
* [X] Tested on Android


src/libs/ReactNativeLauncher.js
```diff
@@ -4,6 +4,7 @@ import MicroEE from 'microee'
 
 import Launcher from './Launcher'
 import ContentScriptBridge from './bridge/ContentScriptBridge'
+import debounce from 'lodash/debounce'
```
Crash--, line 8: @doubleface can you fix these warnings 🙏

Crash-- (approved):

src/libs/ReactNativeLauncher.js
doubleface: Done

PR #700, opened 2023-03-26
Feat/OIDC

Ldoppea:
This PR implements OIDC support for the ClouderyView

___

Cozy Cloudery will implement OIDC login and onboarding using our
partners OIDC systems

ClouderyView now intercepts navigation to URLs containing `oidc=true`
and then starts an OIDC process to the given URL using an InAppBrowser

Two scenario are implemented following the InAppBrowser OIDC process:
one for login an existing user, and one for onboarding a non-existing
user

___

TODO:

- [x] find a way to test on Android (local cloudery seems not to support other than `localhost`)
- [x] remove hardcoded protocol in `startOidcOAuth()`
- [x] find a way to handle `mention légales` PDF link in Cloudery (download not triggered on iOS)
- [ ] ~~verify implementation of 2FA (password vs 2FA code)~~ to be done in another PR
- [x] upgrade react-native-webview after https://github.com/cozy/react-native-webview/pull/4 is merged
- [x] rebase this PR after https://github.com/cozy/cozy-flagship-app/pull/707 is merged
- [x] upgrade cozy-client after https://github.com/cozy/cozy-client/pull/1338 is merged


src/libs/client.js
```diff
@@ -468,3 +468,48 @@ export const removeNotificationDeviceToken = async client => {
 
   await saveClient(client)
 }
+
+export const getOidcAccessToken = async (client, oidcCode) => {
```
Crash--, line 472: this method does not do what it says it does. It doesn't return the `access_token`. Also this method is async and make a request to the stack, should be `fetchSomething`.

src/screens/login/LoginScreen.js
```diff
@@ -172,6 +175,55 @@ const LoginSteps = ({
...
+
+      const result = await getOidcAccessToken(client, code)
+
+      if (result.state === STATE_2FA_NEEDED) {
+        setState(oldState => ({
+          ...oldState,
+          step: TWO_FACTOR_AUTHENTICATION_STEP,
+          isOidc: true,
+          stepReadonly: false,
+          client: result.client,
```
Crash--, line 206: Do we really need to put the client in a state? It seems an anti-pattern here

src/screens/login/components/ClouderyView.js
```diff
@@ -34,7 +38,10 @@ import { rootCozyUrl } from 'cozy-client'
 export const ClouderyView = ({
   clouderyMode,
   setInstanceData,
-  disabledFocus
+  disabledFocus,
+  setError,
+  startOidcOAuth,
+  startOidcOnboarding
```
Crash--, line 45: can you update the jsdoc?

src/screens/login/components/ClouderyView.js
```diff
@@ -58,6 +65,25 @@ export const ClouderyView = ({
...
+            startOidcOnboarding(oidcResult.onboardUrl, oidcResult.code)
+          }
+          return
+        })
+        .catch(error => {
+          if (error !== 'USER_CANCELED') {
+            setError(error.message, error)
+          }
+        })
+      return false
```
Crash--, line 84: I assume you don't await because `handleNavigation` can't be async, right? But can't we have a race condition later in the code?

src/screens/login/components/OidcOnboardingView.tsx
```diff
@@ -0,0 +1,102 @@
+import Minilog from '@cozy/minilog'
+import React, { useEffect, useState } from 'react'
+import { KeyboardAvoidingView, Platform, StyleSheet, View } from 'react-native'
+
+import { SupervisedWebView } from '/components/webviews/SupervisedWebView'
+
+import { parseOidcOnboardingFinishedUrl } from './functions/oidc'
```
Crash--, line 18: alias

src/screens/login/components/OidcOnboardingView.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+import { NetService } from '/libs/services/NetService'
+import { routes } from '/constants/routes'
+
+const log = Minilog('OidcOnboardingView')
+
+interface ClouderyWebViewProps {
+  code: string
+  onboardUrl: string
```
Crash--, line 32: shouldn't be `onboardingUrl`?

src/screens/login/components/OidcOnboardingView.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  useEffect(() => {
+    // Wait 1ms before hiding overlay to prevent screen flickering
+    if (!loading) {
+      setTimeout(() => {
+        setDisplayOverlay(false)
+      }, 1)
+    }
+  }, [loading])
```
Crash--, line 70: 🤔 Between which screens? Can't the first screen send an event or else? 1ms is totally arbitrary, what about old fashion device?

src/screens/login/components/OidcOnboardingView.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const Wrapper = Platform.OS === 'ios' ? View : KeyboardAvoidingView
+
+  return (
+    <Wrapper style={styles.view} behavior="height">
+      <SupervisedWebView
+        source={{ uri: onboardUrl }}
+        onShouldStartLoadWithRequest={handleNavigation}
+        onLoadEnd={(): void => setLoading(false)}
+        style={{
+          backgroundColor: '#4b4b4b'
```
Crash--: theme/color?

src/screens/login/components/OidcOnboardingView.tsx
```diff
@@ -0,0 +1,102 @@
...
+        }}
+        onError={handleError}
+      />
+      {displayOverlay && (
+        <View
+          testID="overlay"
```
Crash--, line 113: should be more specific here

src/screens/login/components/functions/oidc.ts
```diff
@@ -0,0 +1,177 @@
...
+    const url = new URL(request.url)
+
+    const oidcParam = url.searchParams.get(OIDC_PARAM)
+    const isOIDC = oidcParam !== null && oidcParam !== 'false'
+
+    return isOIDC
+  } catch (error) {
+    const errorMessage = getErrorMessage(error)
+    log.error(
+      `Something went wrong while trying to extract OIDC info from URL: ${errorMessage}`
```
Crash--, line 55: you should also log the `url`

src/screens/login/components/functions/oidc.ts
```diff
@@ -0,0 +1,177 @@
...
+    onboardUrlWithRedirect.searchParams.append(
+      OIDC_ONBOARD_CALLBACK_URL_PARAM,
+      OIDC_ONBOARD_CALLBACK_URL
+    )
+
+    return { code, onboardUrl: onboardUrlWithRedirect.toString() }
+  } catch (error) {
+    const errorMessage = getErrorMessage(error)
+    log.error(
+      `Something went wrong while trying to extract OIDC onboarding info from URL: ${errorMessage}`
```
Crash--, line 84: same here

src/screens/login/components/functions/oidc.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    if (!code || !fqdn) {
+      return null
+    }
+
+    return { code, fqdn, defaultRedirection }
+  } catch (error) {
+    const errorMessage = getErrorMessage(error)
+    log.error(
+      `Something went wrong while trying to extract OIDC login info from URL: ${errorMessage}`
```
Crash--, line 106: same here. We must have error messages on which we can act directly

src/screens/login/components/functions/oidc.ts
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    if (!fqdn) {
+      return null
+    }
+
+    return { fqdn, onboardedRedirection }
+  } catch (error) {
+    const errorMessage = getErrorMessage(error)
+    log.error(
```
Crash--, line 172: same here

src/screens/login/components/functions/clouderyBackgroundFetcher.ts
```diff
@@ -0,0 +1,34 @@
...
+export const tryProcessClouderyBackgroundMessage = (
+  event: WebViewMessageEvent,
+  setBackgroundColor: SetBackgroundColorCallback
+): void => {
+  const message = JSON.parse(event.nativeEvent.data) as ClouderyMessage
+
+  if (message.message === 'setBackgroundColor' && message.color !== undefined) {
+    setBackgroundColor(message.color)
+  }
+}
```
Crash--, line 34: I really like this approach, but what are the performances of it? Waiting for "load" and the "getComputedStyle" that request a reflow, I assume it's pretty slow to get the information. Don't we have visually "glitch"?

src/screens/login/components/OidcOnboardingView.tsx
```diff
@@ -7,20 +7,30 @@ import {
...
+  WebViewNavigation
+} from 'react-native-webview'
 
 import { SupervisedWebView } from '/components/webviews/SupervisedWebView'
 
 import { parseOidcOnboardingFinishedUrl } from './functions/oidc'
+import {
+  fetchBackgroundOnLoad,
+  tryProcessClouderyBackgroundMessage
+} from './functions/clouderyBackgroundFetcher'
```
Crash--, line 22: alias

src/screens/login/components/ClouderyViewSwitch.tsx
```diff
@@ -17,7 +22,11 @@ import { showInAppBrowser } from '/libs/intents/InAppBrowser'
 import { setFocusOnWebviewField } from '/libs/functions/keyboardHelper'
 import { NetService } from '/libs/services/NetService'
 import { ClouderyUrls } from '/screens/login/cloudery-env/clouderyEnv'
-import { getColors } from '/ui/colors'
+
+import {
+  fetchBackgroundOnLoad,
+  tryProcessClouderyBackgroundMessage
+} from './functions/clouderyBackgroundFetcher'
```
Crash--, line 29: alias

src/screens/login/LoginScreen.js
```diff
@@ -172,6 +175,55 @@ const LoginSteps = ({
     }))
   }
 
+  const startOidcOAuth = async (fqdn, code) => {
+    if (await NetService.isOffline())
+      NetService.handleOffline(routes.authenticate)
+
+    try {
+      const instance = 'http://' + fqdn
```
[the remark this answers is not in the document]
Ldoppea: Yes this task is already in the TODO list 😉

src/screens/login/LoginScreen.js
Ldoppea: > but maybe we already do that and in that case this is not the responsibility of this PR to fix it 

Yes, the entire LoginScreen pattern is based on a local state that keep a `client` between steps and call `setClient` only when everything is done

We tried to switch this behavior for a `useClient()` based one some month ago. But we encountered some problematics with login/logout scenarios, so we delayed this task.

src/screens/login/components/ClouderyView.js
Ldoppea: I don't think so.

First we return `false` so the navigation is aborted.

Then `startOidcOAuth` go to another step in LoginScreen, so this will unmount the ClouderyView component.
And `startOidcOnboarding` opens the InAppBrowser, so nothing should happen in the react-native side until the InAppBrowser is closed

src/screens/login/components/OidcOnboardingView.tsx
Ldoppea: This was a temporary value and repalced in [feat: Adapt LoginScreen overlay and bar colors to Cloudery's background](https://github.com/cozy/cozy-flagship-app/pull/700/commits/ce05b46cbc939b974b0a22f3a1e9d6942ec54311) commit

src/screens/login/components/functions/oidc.ts
Ldoppea: I would agree for the `isOidcNavigationRequest()` method, but not for `parseOIDCOnboardingStartUrl()` and `parseOIDCLoginUrl()` as those URL should contain some sensible data (OAuth codes).

src/screens/login/components/functions/clouderyBackgroundFetcher.ts
Ldoppea: For now I didn't see anyglitch but I didn't test every scenario (also i tested only iOS).

The base scenario should not be problematic because those view are loaded behind the Welcome screen (which has its own colored bars).

But the color may change in foreground when we will implement the "postcode choice" scenario. I tried to simulate it and it seems ok on iOS. But I would prefer to test it in real condition

src/libs/client.js
Ldoppea: What about `connectOidcClient()` since the method is inspired from `connectClient()` one?

src/screens/login/components/OidcOnboardingView.tsx
Ldoppea: I'm not sure what exactly happens, I just copied the code from ClouderyView based on this PR: https://github.com/cozy/cozy-flagship-app/pull/308/commits/241a428d78fc0e887a47875b7d7a17a50be3637f

What I can tell you:
- `Can't the first screen send an event or else?` -> this is already what is happening. But Pierre had to add this timeout mechanism to prevent a flash.
- `1ms is totally arbitrary` -> in React and Angular worlds, when you see a timeout with 1ms you can assume that the timeout is here just for "delaying" the effect to the next render loop.

Crash--: > 1ms is totally arbitrary -> in React and Angular worlds, when you see a timeout with 1ms you can assume that the timeout is here just for "delaying" the effect to the next render loop.

If the goal is to wait the next tick, then setTimeout(0) do the work. It's clearer than 1. No?

src/screens/login/components/functions/oidc.ts
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/compare/c351d6a960edac9811994f000ba5a790dcd992bf..02727db31d8c486e4afd0241610cd83f87ebf22a

src/screens/login/components/functions/oidc.ts
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/compare/c351d6a960edac9811994f000ba5a790dcd992bf..02727db31d8c486e4afd0241610cd83f87ebf22a

src/screens/login/components/functions/oidc.ts
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/compare/c351d6a960edac9811994f000ba5a790dcd992bf..02727db31d8c486e4afd0241610cd83f87ebf22a

src/screens/login/components/functions/oidc.ts
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/compare/c351d6a960edac9811994f000ba5a790dcd992bf..02727db31d8c486e4afd0241610cd83f87ebf22a

Ldoppea:
I made some edits:
- Adapt URLs to Android's environment: https://github.com/cozy/cozy-flagship-app/compare/ce05b46cbc939b974b0a22f3a1e9d6942ec54311..8f92bdb40d95db634e1ba75910e22286fcffcbb2
- Trim the received colour from Cloudery background, as on Android the received colour would start with a space

Ldoppea:
I made some edits:
- Intercept `window.open` links that should be triggered by the Cloudery when clicking on `terms and condition` links (this is supposed to open a PDF on the same domain): https://github.com/cozy/cozy-flagship-app/compare/4204963c699be2ab172a3a75643e40b14ebcdb14..ec1c29724f224cd27c80cfbf4904b0dbadc6f553
- I also fixed the way I injected the background colour code, as previous implementation would override injected `run` code (used to intercept logs)

src/screens/login/components/ClouderyView.js
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/pull/700/commits/a3a53a99316ab01c0291d1ac15a8af6c044e3a17

src/libs/client.js
Crash--: 👌

Ldoppea:
I added those 3 commits to handle OnboardedRedirection in OIDC scenario
- [refactor: Convert HomeStateProvider to Typescript](https://github.com/cozy/cozy-flagship-app/pull/700/commits/c3f0f4b8dddaa8b9f140853321e8c7f5aeba339b)
- [refactor: Use React context instead of js var for OnboardingRedirection](https://github.com/cozy/cozy-flagship-app/pull/700/commits/a64df375223c95a3099d6cabde7c94f2392f328d)
- [feat: Intercept OnboardedRedirection on OIDC Onboarding scenario](https://github.com/cozy/cozy-flagship-app/pull/700/commits/29bf2955423266ac776f0cce0cee8bbcf87e0f2a)

To check tomorrow: I may need to implement this in the email scenario (start OIDC then receive an email)

src/libs/client.js
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/compare/c5471ef90602aabec1cb57301e0abcadedcfdadd..2542fadaa6f412b928aa438258e603e51a74aabd

src/screens/login/components/OidcOnboardingView.tsx
Ldoppea: Here I chose to fit to the queryParam name from the Cloudery result. Here is an example url I get as the IAB result:
https://links.mycozy.cloud/flagship/oidc_result?code=f38714d48d39916deb2e4577&onboard_url=http%3A%2F%2Flocalhost%3A3000%2Fv2%2Fgrdlyon%2Fstart%3Fonboarding_id%3Du3dF9tYnNqIG763W

Crash-- (approved):
nice

PR #704, opened 2023-03-29
feat: delete existing accounts if needed

acezard:
## What does this do?

- When connecting a Konnector account, we check if any account is linked to this connector. If this is the case, we delete all of them. This avoids multi-account on a CCC.

## Why did you do this?

- For now, we don't want CCC to use the multiple account feature.

## Who/what does this impact?

- Every CCC

## How did you test this?

- Tested manually with accounts created in local CouchDB
- Added unit test for the new function
- The new function logs what it does for debugging purposes

### DIFF
--- a/src/libs/Launcher.functions.ts
+++ b/src/libs/Launcher.functions.ts
@@ -1,0 +1,27 @@
+import CozyClient, { Q } from 'cozy-client'
+import type { CozyClientDocument } from 'cozy-client/types/types'
+
+export const cleanExistingAccountsForKonnector = async (
+  client: CozyClient,
+  konnectorSlug: string,
+  log?: MiniLogger
+): Promise<void> => {
+  const { data: accounts } = (await client.query(
+    Q('io.cozy.accounts').where({
+      accountType: konnectorSlug
+    })
+  )) as { data: CozyClientDocument[] }
+
+  if (accounts.length === 0) {
+    log?.info(`No existing accounts for "${konnectorSlug}"`)
+    return
+  }
+
+  log?.info(
+    `Deleting ${accounts.length} existing accounts for "${konnectorSlug}"`
+  )
+
+  for (const account of accounts) {
+    await client.destroy(account)
+  }
+}
--- a/src/libs/Launcher.functions.spec.ts
+++ b/src/libs/Launcher.functions.spec.ts
@@ -1,0 +1,55 @@
+/* eslint-disable @typescript-eslint/unbound-method */
+import CozyClient from 'cozy-client'
+
+import { cleanExistingAccountsForKonnector } from './Launcher.functions'
+
+const client = {
+  query: jest.fn(),
+  destroy: jest.fn()
+} as Partial<CozyClient> as jest.Mocked<CozyClient>
+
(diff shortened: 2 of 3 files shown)


src/libs/Launcher.functions.ts
```diff
@@ -0,0 +1,27 @@
+import CozyClient, { Q } from 'cozy-client'
+import { CozyClientDocument } from 'cozy-client/types/types'
+
+export const cleanExistingAccountsForKonnector = async (
+  client: CozyClient,
+  konnectorSlug: string,
+  log?: MiniLogger
```
Ldoppea, line 7: Why did you make `log` optional?

src/libs/Launcher.functions.ts
```diff
@@ -0,0 +1,27 @@
+import CozyClient, { Q } from 'cozy-client'
+import { CozyClientDocument } from 'cozy-client/types/types'
```
Ldoppea, line 2:
```suggestion
import type { CozyClientDocument } from 'cozy-client/types/types'
```

src/libs/Launcher.functions.ts
```diff
@@ -0,0 +1,27 @@
...
+    log?.info(`No existing accounts for "${konnectorSlug}"`)
+    return
+  }
+
+  log?.info(
+    `Deleting ${accounts.length} existing accounts for "${konnectorSlug}"`
+  )
+
+  for (const account of accounts) {
+    await client.destroy(account)
```
Ldoppea, line 25: Should we also delete triggers and jobs linked to those accounts?

Ldoppea (approved):
Some question I didn't have in mind during previous daily: Can we have multiple "legacy" accounts in a scenario where we migrated a ServerSideConnector to a ClientSideConnector? Is it expected that we delete those account?

src/libs/Launcher.functions.ts
acezard: as far as I understood they should be deleted by the stack (or if they aren't it's a stack issue). cc @Crash--

src/libs/Launcher.functions.ts
acezard: Didn't want to inject it during the test as I didn't care if it worked or not, so I figured might as well make it optional. The function works whether you inject log or not anyway and is not a logging function per se, it's not tested because TypeScript enforces the verification

acezard:
> Some question I didn't have in mind during previous daily: Can we have multiple "legacy" accounts in a scenario where we migrated a ServerSideConnector to a ClientSideConnector? Is it expected that we delete those account?

I'm not sure about that, the issue is how would I differentiate an account from node to client side? they would have the exact same shape up to the connector slug

Crash--:
>  Can we have multiple "legacy" accounts in a scenario where we migrated a ServerSideConnector to a ClientSideConnector?

For now the scenario doesn't exist. We will see later, but I'm more and more against it.

src/libs/Launcher.functions.ts
Crash--: I confirm.

PR #705, opened 2023-03-30
Feat/welcome cozy pass web

zatteo:
When we open `passwords`, if no extension is installed, we open cozy-pass-web inside the flagship app to show the onboarding.  If extension is installed, we keep calling openAppNative to open native app (or the store).

- [x] update cozy-client https://github.com/cozy/cozy-client/pull/1339
- [x] update cozy-intent https://github.com/cozy/cozy-libs/pull/2092


src/libs/functions/openApp.js
```diff
@@ -41,6 +41,35 @@ const slugFallbacks = {
...
+ * @param {string} href - The app web url
+ * @param {AppManifest} app - The app information
+ * @returns {Promise}
+ */
+const mobileDispatcher = async (client, navigation, href, app, iconParams) => {
+  if (app.slug === 'passwords') {
+    try {
+      const { extension_installed } = await client
+        .getStackClient()
+        .fetchJSON('GET', '/data/io.cozy.settings/io.cozy.settings.bitwarden')
```
zatteo, line 58: On cozy-pass-web, we use fetchJSON directly like this to fetch this value and cache is not wanted.

src/libs/httpserver/indexGenerator.js
```diff
@@ -67,10 +69,18 @@ const isSlugInAllowlist = currentSlug =>
       (slug === currentSlug && platform === Platform.OS)
   )
 
+const isSlugInBlocklist = currentSlug =>
+  slugBlockList.some(
+    ({ slug, platform }) =>
+      (slug === 'ALL' && platform === Platform.OS) ||
```
Ldoppea, line 75: Note that we implemented the "block" mechanism in the past: https://github.com/cozy/cozy-flagship-app/commit/9fcf6c0dc6a12c0e3758e41e8fd7fdc89e91e9a2

We chose to replace it by an "allow" list to fit the following logic: we should first fully test an app before allowing it in this mechanism
Put another way: we chose "opt-in" over "opt-out"

As today we have a new logic that need to implement the "opt-out" version back, then I would do the following:
- Add in the comments that we should priorize the use of the "allow" list over the "block" list
- I would simplify this condition as we should never have `slug === 'ALL'` in the "block" list (today we need at least `home` to be served on all platform and we don't want to disable it by mistake)

src/hooks/useIsNativePassInstalled.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const useIsNativePassInstalled = (): boolean => {
+  const [isNativePassInstalled, setIsNativePassInstalled] = useState(false)
+
+  useEffect(() => {
+    const checkNativePassInstallation = async (): Promise<void> => {
+      const canOpenApp = await Linking.canOpenURL('cozypass://')
+      setIsNativePassInstalled(canOpenApp)
+    }
+    void checkNativePassInstallation()
```
Ldoppea, line 12: I would not put this in a `componentDidMount` effect. I estimate it would be a common scenario to install CozyPass whild the Flagship app is still running. So `isNativePassInstalled` would be wrong after that event.

I think it would be better to execute this logic every time we try to open cozy-pass-web

src/libs/functions/openApp.js
```diff
@@ -41,6 +41,35 @@ const slugFallbacks = {
...
 }
 
+/**
+ * Manage custom behavior for mobile app
+ * Open the native mobile app if no custom behavior
+ * @param {import("cozy-client/dist/index").CozyClient} client - CozyClient instance
+ * @param {any} navigation - The React navigation context
+ * @param {string} href - The app web url
+ * @param {AppManifest} app - The app information
+ * @returns {Promise}
```
Crash--, line 51: miss the iconParams?

src/hooks/useIsNativePassInstalled.ts
Ldoppea: I didn't have in mind that you should also inject this in the webview. Maybe we should find a way to inject it only if the rendered slug is cozy-pass-web, so we would not have to do the check everytime we open a cozy-app

src/libs/functions/openApp.js
Crash--: Yeah, I don't think we should by pass cozy-client redux store. In fact, I'm pretty sure we already do this kind of request during the login phase, and that we'll need this information later when we'll deal with the vault. 

So we should use the "cache" of cozy-client here. 

And since you are in a non react world, I would suggest to not use client.query() with a fetchPolicy (see https://github.com/cozy/cozy-client/issues/931) but instead `fetchQueryAndGetFromState` (see for exemple https://github.com/cozy/cozy-settings/commit/24be85a43c185308075978ce38b81862f2b518f5)

src/libs/functions/openApp.js
```diff
@@ -41,6 +41,35 @@ const slugFallbacks = {
...
+  if (app.slug === 'passwords') {
+    try {
+      const { extension_installed } = await client
+        .getStackClient()
+        .fetchJSON('GET', '/data/io.cozy.settings/io.cozy.settings.bitwarden')
+
+      if (extension_installed) {
+        return openAppNative(app.mobile)
+      } else {
+        return navigateToApp({ navigation, href, slug: app?.slug, iconParams })
```
Crash--, line 63: no need to `app?.slug`  since we are in a test if(app.slug === password)

src/components/webviews/jsInteractions/jsCozyInjection.js
```diff
@@ -19,13 +19,18 @@ const makeMetadata = routeName => {
...
+  isSecureProtocol,
+  isNativePassInstalled
+) => `
   if (!window.cozy) window.cozy = {}
   window.cozy.isFlagshipApp = true
   window.cozy.ClientKonnectorLauncher = 'react-native'
   window.cozy.ClientConnectorLauncher = 'react-native' // deprecated
   window.cozy.flagship = ${makeMetadata(routeName)}
   window.cozy.isSecureProtocol = ${isSecureProtocol || 'false'}
+  window.cozy.isNativePassInstalled = ${isNativePassInstalled || 'false'}
```
Crash--, line 33: I would rename it `isNativePassInstalledOnTheDevice` since it can be confusing with "at least branched once on the cozy".

Crash--: Also, I don't like to give this information to all the apps. Also I think we should avoid relying on window.cozy stuff for this kind of information.

Wondering if the cozy-pass-web should not call an intent (webapp -> native -> webapp) like : `webview.call(isNativePassInstalledOnDevice)`

Ldoppea:
nit: For more clarity, instead of `extension is installed` we should talk about `vault is configured`. The `extension_installed` value in database is a legacy from the first CozyPass version.

At that time the only way to access CozyPass was to use the browser extension, so we used that value to check if the extension was installed or not.

Today the extension is not a requirement anymore, so we suggest to install it but the user can skip this step and log from the web interface instead (or from the mobile app).

We still use the `extension_installed` value in our scenario, but the Functional need is to check if the vault is configured, and not only if the extension is installed.

src/libs/httpserver/indexGenerator.js
zatteo: Thank you for the context 👍

zatteo:
> nit: For more clarity, instead of `extension is installed` we should talk about `vault is configured`. The `extension_installed` value in database is a legacy from the first CozyPass version.
> 
> At that time the only way to access CozyPass was to use the browser extension, so we used that value to check if the extension was installed or not.
> 
> Today the extension is not a requirement anymore, so we suggest to install it but the user can skip this step and log from the web interface instead (or from the mobile app). Login to the web is enough to have `extension_installed=true`
> 
> We still use the `extension_installed` value in our scenario, but the Functional need is to check if the vault is configured, and not only if the extension is installed.

src/libs/functions/openApp.js
zatteo: Will need https://github.com/cozy/cozy-client/pull/1339

src/components/webviews/jsInteractions/jsCozyInjection.js
zatteo: I replace it with an intent with `isNativePassInstalledOnDevice`.

src/hooks/useIsNativePassInstalled.ts
zatteo: Agree. It must be solved with what I did for https://github.com/cozy/cozy-flagship-app/pull/705#discussion_r1155089713

Crash-- (approved):

Ldoppea (approved):

PR #706, opened 2023-03-30
fix: Set appMetadata for LauncherClient

Crash--:
The goal is to have an instanciated cozy-client with the right metadata. Before we had the right token, but the appMetadata was not right.

The createdByApp was right for the io.cozy.files since the stack doesn't read the value from the client, but for triggers or bills, the createdByApp was not right.

With this PR, we should have the right metadata now.

CozyClient is also able to take a `sourceAccount` appMetadata, but I think we don't have a `sourceAccount` at the first launch. So we should update it later, but I don't think cozy-client API allow this ATM.

I also added the appMetadata to the flagship client.

### DIFF
--- a/src/screens/home/hooks/useLauncherClient.ts
+++ b/src/screens/home/hooks/useLauncherClient.ts
@@ -20,9 +20,14 @@ export const useLauncherClient = (
 
     const slug = launcherContextValue?.konnector.slug
 
-    if (slug) void getLauncherClient(client, slug, setLauncherClient)
+    if (slug)
+      void getLauncherClient(
+        client,
+        launcherContextValue.konnector,
+        setLauncherClient
+      )
     else setLauncherClient(undefined)
-  }, [client, launcherContextValue?.konnector.slug])
+  }, [client, launcherContextValue?.konnector])
 
   return { launcherClient }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -6317,16 +6317,16 @@ cozy-client@^34.11.0:
     sift "^6.0.0"
     url-search-params-polyfill "^8.0.0"
 
-cozy-client@^36.2.0:
-  version "36.2.0"
-  resolved "https://registry.yarnpkg.com/cozy-client/-/cozy-client-36.2.0.tgz#662a659415dd9c7d8ae0d69e6c445bcbdaf1a935"
-  integrity sha512-CLom7iJ8b3cCPvA537BNxe7n1DGE6vM48P/xtjLB+Hl7ZKFWMiWSWKHgfQqxKwrDxwnNDD9XJ7VCxwlMBdqpWw==
+cozy-client@^37.1.0:
+  version "37.1.0"
+  resolved "https://registry.yarnpkg.com/cozy-client/-/cozy-client-37.1.0.tgz#d8d583dd79a58c88de811649304763909799ee76"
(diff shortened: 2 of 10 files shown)


doubleface:
@Crash-- To my knowledge, it is not used anywhere. It is juste here for information

Crash--:
@doubleface At least MesPapiers use this information, and I'm wondering if banks or apps like that do the same. 

I'll add the setAppMetadata to cozy-client and use it here. 

@doubleface can you point me where we fetch the account when we run the konnector? In order to be able to call the launcherClient.setAppMetadata() please ?

doubleface:
@Crash-- [Here](https://github.com/cozy/cozy-flagship-app/blob/9ec554e6fc13a66ba2340175bbc8ec6ecd9fca7e/src/libs/Launcher.js#L259) you are sure to have an account.

Or just after ensureAccountAndTriggerAndJob, you are sure to have the account in startContext

Crash--:
Yeah I added it just after `ensureAccountAndTriggerAndJob` see ea15718

yarn.lock
```diff
@@ -6317,16 +6317,16 @@ cozy-client@^34.11.0:
     sift "^6.0.0"
     url-search-params-polyfill "^8.0.0"
 
-cozy-client@^36.2.0:
-  version "36.2.0"
-  resolved "https://registry.yarnpkg.com/cozy-client/-/cozy-client-36.2.0.tgz#662a659415dd9c7d8ae0d69e6c445bcbdaf1a935"
-  integrity sha512-CLom7iJ8b3cCPvA537BNxe7n1DGE6vM48P/xtjLB+Hl7ZKFWMiWSWKHgfQqxKwrDxwnNDD9XJ7VCxwlMBdqpWw==
+cozy-client@^37.1.0:
```
doubleface, line 6320: Why this update of cozy-client ?

doubleface: Oups, just saw the linked cozy-client PR :+1:

src/screens/home/hooks/useLauncherClient.ts
```diff
@@ -20,9 +20,14 @@ export const useLauncherClient = (
...
-    if (slug) void getLauncherClient(client, slug, setLauncherClient)
+    if (slug)
+      void getLauncherClient(
+        client,
+        launcherContextValue.konnector,
+        setLauncherClient
+      )
     else setLauncherClient(undefined)
-  }, [client, launcherContextValue?.konnector.slug])
+  }, [client, launcherContextValue?.konnector])
```
doubleface, line 30: Did you check it does not generate multiple rerenders ? Just in case

src/screens/home/hooks/useLauncherClient.ts
```diff
@@ -20,9 +20,14 @@ export const useLauncherClient = (
 
     const slug = launcherContextValue?.konnector.slug
```
acezard, line 21: FYI this will fail if we somehow get an odd object named `connector` instead of `konnector`, IDK if it's worth preventing the issue. As of now there is no safety net and I had multiple crashes where this error happened

acezard (approved):

src/screens/home/hooks/useLauncherClient.ts
Crash--: I don't know how we can have a `connector` object. If you still got this issue, we should dig in to fix the root cause.

PR #710, opened 2023-03-31
feat: Send LOGIN_SUCCESS event via postMessage to the home application

doubleface:
This depends on https://github.com/cozy/cozy-libs/pull/2091 to be
deployed on home application to work completely

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -222,7 +222,8 @@ class ReactNativeLauncher extends Launcher {
       if (ensureResult.createdJob) {
         this.emit('CREATED_JOB', ensureResult.createdJob)
       }
-      await this.sendLoginSuccess()
+
+      launcherEvent.emit('loginSuccess', ensureResult.createdAccount?._id)
 
       const pilotContext = []
       // FIXME not used at the moment since the fetched file will not have the proper "createdByApp"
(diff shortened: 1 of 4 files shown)


src/libs/ReactNativeLauncher.js
```diff
@@ -222,7 +222,8 @@ class ReactNativeLauncher extends Launcher {
       if (ensureResult.createdJob) {
         this.emit('CREATED_JOB', ensureResult.createdJob)
       }
-      await this.sendLoginSuccess()
+
+      launcherEvent.emit('loginSuccess', ensureResult.createdAccount?._id)
```
Crash--, line 226: why are we sending the `createdAccount._id`? It seems that we don't read it from the other side (cause we read it from the flow)? https://github.com/cozy/cozy-libs/blob/88ea4b202b6766db64f1cd22cd4c7d9412d7f8e0/packages/cozy-harvest-lib/src/components/FlowProvider.jsx#L132-L153

doubleface: Actually, not reading it on harvest side was my mistake, because sometimes, the "loginSuccess" event can get to harvest before the account and trigger creation realtime events.
harvest is now fixed and uses this account id

Crash-- (approved):

PR #715, opened 2023-04-01
fix: Intercept external links in OidcOnboardingView

Ldoppea:
Cloudery can display some external links in partners pages, so we should intercept them and display them in an InAppBrowser

Also some links would be open by the ClouderyView using `window.open()` method. This should be used for the "terms and condition" PDF that is stored on the Cloudery (and so we cannot detect URL.origin change)

By using `onOpenWindow` we would intercept those URLs

### DIFF
--- a/src/screens/login/components/ClouderyViewSwitch.tsx
+++ b/src/screens/login/components/ClouderyViewSwitch.tsx
@@ -13,7 +13,6 @@ import type {
   WebViewNavigation,
   WebViewProps
 } from 'react-native-webview'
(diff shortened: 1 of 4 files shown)


Crash-- (approved):
Approved, but console.log to remove 🙏

Ldoppea:
> Approved, but console.log to remove 🙏

Woups 😅 Fixed in https://github.com/cozy/cozy-flagship-app/compare/16854d31972cea00b618f339017cffc0ff7a56ae..a157ee9f9964c2fac4f16b30c61b9ffb119bafcc

Crash-- (approved):

PR #717, opened 2023-04-01
fix: Set status bar text color on Login screen based on background

Ldoppea:
Now that we adapt the status bar color to the Cloudery view's background we need to adapt the status bar foreground color to ensure it is still readable

___

I chose to make it dynamic by doing Math on background color, this should be future proof when we will add more partners. However the formula is not perfect as mentioned in the code comments. There is a more complex formula that we can implement in the future if we need to


Crash-- (approved):

zatteo (approved):

Ldoppea:
I added a try...catch and checked that the string is a correct regex here : https://github.com/cozy/cozy-flagship-app/compare/4dae40ce9924ecec612109a2ab192694b81f3531..3e5d1597de199540d97321a8e2e2e926fe5fc816

Crash-- (approved):

PR #722, opened 2023-04-04
feat: Handle new redirection URL type in ClouderyCreateInstanceView and OidcOnboardingView

Ldoppea:
This PR handle new OIDC redirection links from latest Cloudery changes

All commits contain their own description with context so feel free to review commit by commit

### DIFF
--- a/src/screens/login/components/OidcOnboardingView.tsx
+++ b/src/screens/login/components/OidcOnboardingView.tsx
@@ -22,11 +22,7 @@ import {
   fetchBackgroundOnLoad,
   tryProcessClouderyBackgroundMessage
 } from '/screens/login/components/functions/clouderyBackgroundFetcher'
-import {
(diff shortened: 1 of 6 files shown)


src/screens/login/components/OidcOnboardingView.tsx
Crash--, line 1: > But as we keep the `onOpenWindow` interception, we can still handle
external links, we agreed that ALL external links should be open using
a new tab on Cloudery pages

In order to not think about tab since my brain doesn't handle it pretty well, does it mean that all external links on the cloudery should use `target=_blank` or use `window.open`?

Crash-- (approved):

src/screens/login/components/OidcOnboardingView.tsx
Ldoppea: yes it does mean that. I edited the commit message to add precision.

PR #723, opened 2023-04-05
feat: Trigger quality_checks CI on all PRs

Ldoppea:
~~Current `quality_check` workflow is only triggered on PRs that target the `master` branch~~

~~We want to allow meta PRs to be checked by this workflow too~~

~~To make this happens, we should now name branches `feat/meta_pr_*`~~

Current `quality_check` workflow is only triggered on PRs that target
the `master` branch

We want all PRs to be checked by this workflow too

___

This PR is for https://github.com/cozy/cozy-flagship-app/pull/708 that is the target branch for all OIDC PRs and I noticed that all child PRs were not checked

### DIFF
--- a/.github/workflows/quality_checks.yml
+++ b/.github/workflows/quality_checks.yml
@@ -3,8 +3,6 @@ name: React-Native CI
 on:
   workflow_dispatch:
   pull_request:
-    branches:
-      - master
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


Crash--:
Shoudl we run it on EVERY pr? Not only master or meta_* ?

zatteo (approved):

Ldoppea:
> Shoudl we run it on EVERY pr? Not only master or meta_* ?

I don't know what would be the impact. At least this formalizes our branch naming convention : if a branch should receive other PRs, then name it `feat/meta_pr_something`.

Ldoppea:
What about now?

Crash-- (approved):

PR #728, opened 2023-04-06
feat: Make field autofocus more generic in login and onboarding screens

Ldoppea:
In previous implementations, we handled field autofocus from the react-native components by predicting which field would be displayed

Those fields were easily predictable as only a few login and onboarding scenario were implemented

Now that we implemented OIDC scenario, it is harder to predict those fields. For example, the link from email onboarding can now redirect to the Cloudery's FQDN choice page, or the ToS page

The new implementation allow to focus whatever DOM field that has `autofocus` prop. If no `autofocus` field is present, then no focus is triggered. This would prevent the keyboard to appear on the ToS page

Also as this mechanism is triggered from the HTML page and on page's load event, then the keyboard focus is handled on every navigation, which would make the autofocus experience more robust and homogenous

> **Note**
> OnboardingPasswordView, PasswordView and TwoFactorAuthenticationView all use the old autofocus mechanism. I didn't adapt them as those are locally handled pages (html is injected) so I don't feel the need to adapt them
> However I may change my mind based on how (and if) we will fix the missing autofocus feature on iOS

### DIFF
--- a/src/libs/functions/keyboardHelper.ts
+++ b/src/libs/functions/keyboardHelper.ts
@@ -1,5 +1,5 @@
 import { NativeModules, Platform } from 'react-native'
-import type { WebView } from 'react-native-webview'
+import type { WebView, WebViewMessageEvent } from 'react-native-webview'
 
 import log from 'cozy-logger'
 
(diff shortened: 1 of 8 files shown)


Crash-- (approved):
Nice!

PR #729, opened 2023-04-06
Change app icon

zatteo:
Goal : change native icon when cozy default app is mespapiers to a custom icon. Set to default icon for everything else. 

#### Checklist

* [x] ~~Faithful integration of the mockups at all screen sizes~~
* [x] Tested on iOS
* [x] Tested on Android
* [x] ~~Localized in English and French~~
* [x] All changes have test coverage
* [x] ~~Updated README & CHANGELOG, if necessary~~

- [x] Add the real mespapiers icon
- [x] Test with ~~OIDC process~~ app init


android/app/src/main/AndroidManifest.xml
```diff
@@ -59,16 +59,43 @@
...
+        <intent-filter>
+          <action android:name="android.intent.action.MAIN" />
+          <category android:name="android.intent.category.LAUNCHER" />
+        </intent-filter>
+      </activity-alias>
+
       <activity
         android:name="com.zoontek.rnbootsplash.RNBootSplashActivity"
         android:theme="@style/BootTheme"
+        android:enabled="false"
```
Crash--, line 90: Sure about this one? Can you explain why?

zatteo: Sorry, it is because it generates duplicate icons. See https://github.com/skb1129/react-native-change-icon/issues/44#issuecomment-974042627

Crash-- (approved):

src/libs/icon/icon.ts
```diff
@@ -0,0 +1,18 @@
+import { changeIcon as RNChangeIcon } from 'react-native-change-icon'
+
+const DEFAULT_ICON = 'cozy'
+const ALLOWED_ICONS = ['mespapiers']
```
acezard, line 4: I'm not sure this kind of data/config should be in functional files. Though I don't have a suggestion where you could put them. maybe src/libs/icon/config.json for now

PR #733, opened 2023-04-12
feat: Add Magic Link support to Login and Onboarding scenario

Ldoppea:
We want the user to be able to create and login to their Cozy using passwordless authentication

To make this possible, when the user onboard to a partner context by entering their email, then the sent email contains a `magic link` that allows to complete the onboarding without entering any password

Then with this account type, the user can login using the same principle. After entering their email or their instance url, the Cloudery send a `magic link` by email and this magic link would allow to login without the password step

Related PR: cozy/cozy-stack#3872

> **Note**
> This PR adds more complexity to client.js, so I added another PR to refactor it: #734

### DIFF
--- a/src/hooks/useAppBootstrap.js
+++ b/src/hooks/useAppBootstrap.js
@@ -11,6 +11,7 @@ import { navigate } from '/libs/RootNavigation'
 import { routes } from '/constants/routes'
 import {
   parseFallbackURL,
+  parseMagicLinkURL,
   parseOnboardingURL
 } from '/hooks/useAppBootstrap.functions'
 import { useSplashScreen } from '/hooks/useSplashScreen'
@@ -174,6 +175,17 @@ export const useAppBootstrap = client => {
           navigate(routes.authenticate, { fqdn })
           return
         }
+      }
+
+      const magicLink = parseMagicLinkURL(url)
+
+      if (magicLink) {
+        const { fqdn, magicCode } = magicLink
+        log.debug(
+          `🔗 Redirect to authenticate screen for ${fqdn} with magicCode`
+        )
+        navigate(routes.authenticate, { fqdn, magicCode })
+        return
       }
 
       const { mainAppFallbackURL, cozyAppFallbackURL } = parseFallbackURL(url)
--- a/src/libs/client.js
+++ b/src/libs/client.js
@@ -509,3 +509,93 @@ export const connectOidcClient = async (client, oidcCode) => {
     state: STATE_CONNECTED
   }
 }
+
+export const connectMagicLinkClient = async (client, magicCode) => {
+  const stackClient = client.getStackClient()
+
+  let oauthOptions = stackClient.oauthOptions
+  const data = {
+    magic_code: magicCode,
+    client_id: oauthOptions.clientID,
+    client_secret: oauthOptions.clientSecret,
+    scope: '*'
+  }
+
+  const {
+    two_factor_token: twoFactorToken,
+    session_code: sessionCode,
+    ...token
+  } = await stackClient.fetchJSON('POST', '/auth/magic_link/flagship', data)
+
+  const need2FA = twoFactorToken !== undefined
+
+  if (need2FA) {
+    return {
+      client,
+      state: STATE_2FA_NEEDED,
+      twoFactorToken: twoFactorToken
+    }
+  }
+
+  const needFlagshipVerification = sessionCode !== undefined
+
+  if (needFlagshipVerification) {
+    return {
+      client: client,
+      state: STATE_AUTHORIZE_NEEDED,
+      sessionCode: sessionCode
+    }
+  }
+
+  stackClient.setToken(token)
+
+  return {
+    client: client,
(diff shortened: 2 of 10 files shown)


src/hooks/useAppBootstrap.js
```diff
@@ -176,6 +177,17 @@ export const useAppBootstrap = client => {
         }
       }
 
+      const magicLink = parseMagicLinkURL(url)
+
+      if (magicLink) {
+        const { fqdn, magicCode } = magicLink
+        log.debug(
+          `🔗 Redirect to authenticate screen for ${fqdn} with magiCode`
```
Crash--, line 185:
```suggestion
          `🔗 Redirect to authenticate screen for ${fqdn} with magicCode`
```

src/libs/client.js
```diff
@@ -509,3 +509,93 @@ export const connectOidcClient = async (client, oidcCode) => {
...
+
+export const connectMagicLinkClient = async (client, magicCode) => {
+  const stackClient = client.getStackClient()
+
+  let oauthOptions = stackClient.oauthOptions
+  const data = {
+    magic_code: magicCode,
+    client_id: oauthOptions.clientID,
+    client_secret: oauthOptions.clientSecret,
+    scope: '*'
```
Crash--, line 521: why do we need to pass the `scope`?

Crash--: Seems that this is not documented on the stack's side https://github.com/cozy/cozy-stack/pull/3872/files#diff-d9cbd97d17d60b9a415f168c988e7af55ab4d6945e24b0fb570a0577eff6b884R294

src/libs/client.js
```diff
@@ -509,3 +509,93 @@ export const connectOidcClient = async (client, oidcCode) => {
...
+      client: client,
+      state: STATE_AUTHORIZE_NEEDED,
+      sessionCode: sessionCode
+    }
+  }
+
+  stackClient.setToken(token)
+
+  return {
+    client: client,
```
Crash--, line 553: I still don't like that, but you know it :p

src/libs/client.js
```diff
@@ -509,3 +509,93 @@ export const connectOidcClient = async (client, oidcCode) => {
...
+  const stackClient = client.getStackClient()
+
+  await client.certifyFlagship()
+
+  let oauthOptions = stackClient.oauthOptions
+  const data = {
+    magic_code: magicCode,
+    client_id: oauthOptions.clientID,
+    client_secret: oauthOptions.clientSecret,
+    scope: '*'
```
Crash--, line 572: same here?

src/screens/login/LoginScreen.js
```diff
@@ -231,6 +243,48 @@ const LoginSteps = ({
     }
   }
 
+  const startMagicLinkOAuth = useCallback(
+    async (fqdn, magicCode) => {
+      if (await NetService.isOffline())
+        NetService.handleOffline(routes.authenticate)
+
+      try {
+        const instance = 'https://' + fqdn
```
Crash--, line 252: This time we're not only working with a remote / local cloudery right. We can use this `startMagicLinkOAuth` with a localhost instance, right? If yes, we should add http if localhost or cozy.tools, no?

src/screens/login/OnboardingScreen.js
```diff
@@ -154,6 +203,16 @@ const OnboardingSteps = ({ setClient, route, navigation }) => {
...
 
+  if (state.step === MAGIC_LINK_STEP) {
+    hideSplashScreen()
+    const { fqdn } = state.onboardingData
+    return (
+      <View>
+        <Text>Magic Onboarding on {fqdn}</Text>
+      </View>
+    )
+  }
```
Crash--, line 214: is it really wanted this screen?

Ldoppea: No it is not wanted. I forgot to mention that, but this part is temporary. I didn't implement the final version yet as this is not in the critical path and it is discreet (displayed half a second before we get the authorization IAB or the home navigation)

src/libs/client.js
Ldoppea: Good catch, I assumed it was the same signature as for the OIDC endpoint, I should have double-checked that part.

As this code has been moved in another file in other PRs, I did the fix in this PR: https://github.com/cozy/cozy-flagship-app/pull/737/commits/3ba803f0e4bd002462957b8a79b68abb391bbdd3

src/hooks/useAppBootstrap.js
Ldoppea: Fixed: https://github.com/cozy/cozy-flagship-app/compare/8211342d36a056a96102d3a31410147d41f466d2..3587507e53d3cf6584feba3015f2eca9bafa0ba7

src/screens/login/LoginScreen.js
Ldoppea: We can try to guess it with localhost, cozy.tools and nip.io yes. Are you ok if I do this edit in another PR in the following days?

Crash-- (approved):

PR #735, opened 2023-04-12
feat: Rename app from `Cozy Cloud` to `Cloud Personnel`

Ldoppea:
We want to rename the app to better fit the "neutral" scenario we are implementing

Related PR: #708


Crash-- (approved):
Wondering if we'll have to do a few manipulations on the store side?

Ldoppea:
> Wondering if we'll have to do a few manipulations on the store side?

We will have to do the renaming to the store also, but my understanding is that we will have to push a new binary and a new release first to make the edit.

zatteo (approved):

PR #737, opened 2023-04-12
feat: Handle password based 2FA on Magic Link login scenario

Ldoppea:
On Magic Link scenario, if the user enabled 2FA, then then 2FA should be done using the user's password instead of the usual code sent by email

This choice has been made because the Magic Link scenario already uses the email channel, so a 2FA code sent through the same channel would add no benefits to security. So instead we ask to the user for their password

To make this possible, we will have to modify cozy-settings so the 2FA feature requires a password to be set on the Cozy

> **Note**
> I'm not happy with the complexity I added on the LoginScreen with this PR but also with the OIDC one. We now have a lot of similar methods (`startOAuth`, `continueOAuth`, `startOidcOAuth`, `startMagicLinkOAuth`, `continueMagicLinkOAuth`) but with enough difference to make it difficult to factorize them. I expect to rework on this after the delivery deadline.

### DIFF
--- a/src/libs/clientHelpers/magicLink.ts
+++ b/src/libs/clientHelpers/magicLink.ts
@@ -1,5 +1,4 @@
 import type CozyClient from 'cozy-client'
-import type { LoginFlagshipResult } from 'cozy-client'
 
 import { authorizeClient } from '/libs/clientHelpers/authorizeClient'
 import { createClient } from '/libs/clientHelpers/createClient'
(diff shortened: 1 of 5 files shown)


src/libs/clientHelpers/magicLink.ts
```diff
@@ -10,38 +9,93 @@ import {
...
+  const stackClient = client.getStackClient()
+
+  try {
+    const oauthOptions = stackClient.oauthOptions
+    const data = {
+      magic_code: magicCode,
+      passphrase: passwordHash,
+      client_id: oauthOptions.clientID,
+      client_secret: oauthOptions.clientSecret,
+      scope: '*'
```
Crash--: No need?

PR #750, opened 2023-04-20
fix: Disable ClouderyView tests that timeout

Ldoppea:
Those are flacky tests that often timeout on Github Action CI

We didn't find a way to stabilise them yet, so until a solution is found we want them to be disabled

Related commit: 3cb78493449dc5795ebb152c512502b702820440


Crash-- (approved):
🤷‍♂️

Ldoppea:
I created an issue to track the problem: https://github.com/cozy/cozy-flagship-app/issues/753

PR #758, opened 2023-04-25
Fix/deep linking change icon

zatteo:
Android only deep link + change icon fixes

fix: Move launchMode in .MainActivity
fix: App can not be opened anymore when switching icon 2x

### DIFF
--- a/patches/react-native-change-icon+4.0.0.patch
+++ b/patches/react-native-change-icon+4.0.0.patch
@@ -1,0 +1,12 @@
+diff --git a/node_modules/react-native-change-icon/android/src/main/java/com/reactnativechangeicon/ChangeIconModule.java b/node_modules/react-native-change-icon/android/src/main/java/com/reactnativechangeicon/ChangeIconModule.java
+index 07402bc..d44f29d 100644
+--- a/node_modules/react-native-change-icon/android/src/main/java/com/reactnativechangeicon/ChangeIconModule.java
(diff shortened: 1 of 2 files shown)


patches/react-native-change-icon+4.0.0.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/react-native-change-icon/android/src/main/java/com/reactnativechangeicon/ChangeIconModule.java b/node_modules/react-native-change-icon/android/src/main/java/com/reactnativechangeicon/ChangeIconModule.java
+index 07402bc..d44f29d 100644
+--- a/node_modules/react-native-change-icon/android/src/main/java/com/reactnativechangeicon/ChangeIconModule.java
++++ b/node_modules/react-native-change-icon/android/src/main/java/com/reactnativechangeicon/ChangeIconModule.java
+@@ -82,6 +82,7 @@ public class ChangeIconModule extends ReactContextBaseJavaModule implements Appl
+             promise.reject("ICON_INVALID");
+             return;
+         }
++        this.classesToKill.clear();
```
Crash--, line 9: do you have created an upstream PR? If yes, can you link it to the commit message?

Crash-- (approved):

patches/react-native-change-icon+4.0.0.patch
zatteo: Not for the moment. It may have consequences depending on how you want to use the plugin. But I will create it and start a discussion with the owner.

PR #765, opened 2023-04-26
feat: Handle Blocked Cozy on Login screen

Ldoppea:
When trying to log in a "blocked" Cozy then the app should display a custom error page explaining to the user what is happening and what to do

This can happen on multiple client's calls (i.e. `client.register()`) so to be sure to intercept the error we chose to handle it in the ErrorView which would redirect to the correct screen if detected. This should handle classic login and OIDC, MagicLink ones

The classic login scenario depends on cozy/cozy-client#1352, otherwise the displayed screen will be CozyNotFound

The new CozyBlockedPage displays a forbidden logo that is a copy of the one from cozy-stack (`assets/templates/instance_blocked.html`)

Related PR: cozy/cozy-client#1352
Related PR: cozy/cozy-client#1343

____

TODO:

- [ ] Upgrade cozy-client when cozy/cozy-client#1352 and cozy/cozy-client#1343 merged


Ldoppea:
> **Note**
> In this commit I still use hardcoded value for the gray theme, we discussed about this with @nono, I still have to do some tests to make it dynamic, it's on my todo list

Ldoppea:
I close this PR in favor of https://github.com/cozy/cozy-flagship-app/pull/876

PR #773, opened 2023-05-04
fix: Run Debug on real Device

Crash--:
I was not able to run the debug build on my real iphone because the profil & cert were not for debug purpose.

So we added a new provisioning profile "amiral-dev-profile". You can add a new device to this profile by going to the developper apple website.

And since we added a profile for dev purpose we needed to create a dev certificat. This certificate is available in the password store


Crash--:
@Ldoppea can you review it?

Crash--:
@Ldoppea ping pong

PR #777, opened 2023-05-10
fix: Resolve setWorkerState only when webview reference is ready

doubleface:
In some cases, we still had the following message :
"nodeHandle expected to be non-null"

This was a regression after #751
In the case a webview reload was forced :

Ex : when the clisk konnector does :

```javascript
await this.goto(url)
await this.runInWorker('any  command')
```

The goto command resolved when the worker webview was not ready yet and
the launcher did not have a reference to the new webview yet.

Now, the goto command waits for the webview reference to be ready to be
resolved.

Also solve some minor issues I found during the debug

- fix: Resolve setWorkerState only when webview reference is ready
- fix: Do not create a new binded function on each message
- fix: Display the number of logs we try to send to the cozy stack
- feat: Format messenger debug message

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -18,6 +18,8 @@ import { wrapTimerFactory } from 'cozy-clisk'
 
 const log = Minilog('ReactNativeLauncher')
 
+const SET_WORKER_STATE_TIMEOUT_MS = 30 * 1000
+
 Minilog.enable()
 
 function LauncherEvent() {}
@@ -288,11 +290,21 @@ class ReactNativeLauncher extends Launcher {
    *
    * @param {SetWorkerStateOptions} options
    */
-  async setWorkerState(options) {
-    this.emit('SET_WORKER_STATE', options)
-    if (options?.visible === true) {
-      this.emit('worker:visible')
-    }
+  setWorkerState(options) {
+    return new Promise((resolve, reject) => {
+      let timerId = null
+      this.emit('SET_WORKER_STATE', options)
+      this.once('worker:webview:ready', () => {
+        clearTimeout(timerId)
+        if (options?.visible === true) {
+          this.emit('worker:visible')
+        }
+        resolve()
+      })
+      timerId = setTimeout(() => {
+        reject('ReactNativeLauncher.setWorkerState took more than 30000 ms')
+      }, options?.timeout || SET_WORKER_STATE_TIMEOUT_MS)
+    })
   }
 
   /**
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -20,6 +20,11 @@ const log = Minilog('ReactNativeLauncher')
 
 const SET_WORKER_STATE_TIMEOUT_MS = 30 * 1000
 
+export const ERRORS = {
+  SET_WORKER_STATE_TOO_LONG_TO_INIT:
+    'ReactNativeLauncher.setWorkerState took more than 30000 ms'
+}
+
 Minilog.enable()
 
 function LauncherEvent() {}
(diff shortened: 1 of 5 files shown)


src/libs/ReactNativeLauncher.js
```diff
@@ -269,11 +271,21 @@ class ReactNativeLauncher extends Launcher {
...
+      this.emit('SET_WORKER_STATE', options)
+      this.once('worker:webview:ready', () => {
+        clearTimeout(timerId)
+        if (options?.visible === true) {
+          this.emit('worker:visible')
+        }
+        resolve()
+      })
+      timerId = setTimeout(() => {
+        reject('ReactNativeLauncher.setWorkerState took more than 30000 ms')
```
Crash--, line 286: nit: Shouldn't we start using const to have some kind of errors constant to match them easily within the app? Matching string is error prone. 

```

reject(ERRORS.SET_WORKER_STATE_TOO_LONG_TO_INIT)

...
catch(e){
if(e.message === ERRORS.SET_WORKER_STATE_TOO_LONG_TO_INIT) {
...
}
}

```

Like that, we can also have a quick views at all the errors we can have.

const ERRORS = [...]

src/screens/konnectors/LauncherView.js
```diff
@@ -259,7 +261,10 @@ class LauncherView extends Component {
               <WebView
                 key={this.state.workerKey}
                 mediaPlaybackRequiresUserAction={true}
-                ref={ref => (this.workerWebview = ref)}
+                ref={ref => {
+                  this.workerWebview = ref
+                  this.launcher.emit('worker:webview:ready')
+                }}
```
Crash--, line 267: I can understand this change, but I'm not understanding the change made here https://github.com/cozy/cozy-flagship-app/pull/777/commits/69410ca7c301e72e034dd89abcde1db416c9e735#diff-83374be02352a4c209865e6cab112ff6d487b98a8e9000cc58e5e42e7898a993R173-R174

Can you explain it please? 

Also, you should test that `ref` is not null. Ref callback can be called with ref === null and I think that in that case you should not emit. See https://github.com/facebook/react/issues/4533 & https://github.com/facebook/react/pull/8333/files

src/constants/dev-config.ts
```diff
@@ -4,11 +4,11 @@
  * Please do not use this object directly in other files, use the exported functions instead.
  */
 export const devConfig = {
-  disableGetIndex: false, // Bypass HTTPServer HTML generation which will fallback to local stack, useful for webpack-dev-server
```
Crash--, line 7: should not be commited

src/libs/bridge/ReactNativeLauncherMessenger.js
```diff
@@ -38,9 +41,21 @@ export default class ReactNativeLauncherMessenger extends MessengerInterface {
...
 }
+
+function formatIds(message) {
+  const { sessionId, requestId, ...rest } = message
+  label+= 'sessionId=' + sessionId
+  if (requestId) {
+    label+= ' requestId=' + requestId
+  }
+  return {label, rest }
+}
```
Crash--, line 61: You don't have the linter installed on your machine? It doesn't work?

src/screens/konnectors/LauncherView.js
doubleface: This other event emit is done when the url change does not need any webview reload. In this case, we know that the webview reference won't change and we can resolve the `setWorkerState` promise directly

Crash--: Maybe adding this in a comment can be useful? 🙏

Crash-- (approved):

PR #781, opened 2023-05-12
fix: Resolve goto when the bridge is ready

doubleface:
Before we resolved the goto when the webview was reload but but it fails
when running goto twice in a row for example (and also in other cases on
slow devices).

Also added some enhancements on post-me logs display
- feat: Do not log non post-me messages in messenger
- fix: Call the listener when non null
- fix: Show requestId even when = 0
- feat: Better display of handshake logs
- feat: Hide log events

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -298,7 +298,7 @@ class ReactNativeLauncher extends Launcher {
   setWorkerState(options) {
     return new Promise((resolve, reject) => {
       let timerId = null
-      this.emit('SET_WORKER_STATE', options)
+
       this.once('worker:webview:ready', () => {
         clearTimeout(timerId)
         if (options?.visible === true) {
@@ -309,6 +309,7 @@ class ReactNativeLauncher extends Launcher {
       timerId = setTimeout(() => {
         reject(ERRORS.SET_WORKER_STATE_TOO_LONG_TO_INIT)
       }, options?.timeout || SET_WORKER_STATE_TIMEOUT_MS)
+      this.emit('SET_WORKER_STATE', options)
     })
   }
 
@@ -410,6 +411,7 @@ class ReactNativeLauncher extends Launcher {
       throw new Error(`worker bridge restart init error: ${err.message}`)
     }
     this.emit('WORKER_RELOADED')
+    this.emit('worker:webview:ready')
   }
 
   /**
--- a/src/libs/bridge/ReactNativeLauncherMessenger.js
+++ b/src/libs/bridge/ReactNativeLauncherMessenger.js
@@ -60,7 +60,7 @@ function formatIds(message) {
   // eslint-disable-next-line no-unused-vars
   const { sessionId, requestId, type, ...rest } = message
   let label = 'sessionId=' + sessionId
-  if (requestId) {
+  if (requestId !== undefined) {
     label += ' requestId=' + requestId
   }
   return { label, rest }
--- a/src/libs/bridge/ReactNativeLauncherMessenger.js
+++ b/src/libs/bridge/ReactNativeLauncherMessenger.js
@@ -41,8 +41,13 @@ export default class ReactNativeLauncherMessenger extends MessengerInterface {
    */
   onMessage(event) {
     const data = JSON.parse(event.nativeEvent.data)
-    this.listener({ data })
-    if (this.debug && data.type === '@post-me') {
+    if (this.listener) {
+      // in scenarios where we watch the flagship app with hot reload I found this.listener to be null sometimes.
+      // As addMessageListener is directly called by post-me, I don't know why, it may be null but this avoids some headaches
+      // when developing on flagship app.
+      this.listener({ data })
+    }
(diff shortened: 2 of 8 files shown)


src/libs/ReactNativeLauncher.js
```diff
@@ -410,6 +411,7 @@ class ReactNativeLauncher extends Launcher {
       throw new Error(`worker bridge restart init error: ${err.message}`)
     }
     this.emit('WORKER_RELOADED')
+    this.emit('worker:webview:ready')
   }
```
Crash--, line 415: > but it fails when running goto twice in a row for example 

Can you add a test for that please?

src/libs/ReactNativeLauncher.js
```diff
@@ -309,6 +309,7 @@ class ReactNativeLauncher extends Launcher {
       timerId = setTimeout(() => {
         reject(ERRORS.SET_WORKER_STATE_TOO_LONG_TO_INIT)
       }, options?.timeout || SET_WORKER_STATE_TIMEOUT_MS)
+      this.emit('SET_WORKER_STATE', options)
```
Crash--, line 312: I don't understand why this change is necessary in this commit? Can you explain?

src/libs/bridge/ReactNativeLauncherMessenger.js
```diff
@@ -15,7 +15,7 @@ export default class ReactNativeLauncherMessenger extends MessengerInterface {
   }
```
Crash--, line 15: can you add the `why` in the commit message? What kind of messages are logged that we don't want anymore?

src/libs/bridge/ReactNativeLauncherMessenger.js
```diff
@@ -41,7 +41,7 @@ export default class ReactNativeLauncherMessenger extends MessengerInterface {
    */
   onMessage(event) {
     const data = JSON.parse(event.nativeEvent.data)
-    this.listener({ data })
+    if (this.listener) this.listener({ data })
```
Crash--, line 48: I don't like this kind of code. 

`this.listener` is added during the `addMessageListener`. So we should not be able to pass within the `onMessage` function without calling `addMessageListener` first. 

So it's the argument we pass to the `addMessageListener` that is null. So we should not call the `addMessageListener` if the `listener` is null, no? Where are we calling this method?

src/libs/bridge/ReactNativeLauncherMessenger.js
```diff
@@ -46,7 +46,7 @@ export default class ReactNativeLauncherMessenger extends MessengerInterface {
   onMessage(event) {
     const data = JSON.parse(event.nativeEvent.data)
     if (this.listener) this.listener({ data })
-    if (this.debug && data.type === '@post-me') {
+    if (this.debug && data.type === '@post-me' && data.eventName !== 'log') {
```
Crash--, line 49: This commit should be squashed with the one adding the `post-me` check. But I still don't understand the need.

src/libs/ReactNativeLauncher.js
Crash--: Since this is at least the second time in a row we have an issue on this case, we should add a lost of tests!

src/libs/ReactNativeLauncher.js
doubleface: Je vais mieux y arriver en français : l'évènement `SET_WORKER_STATE` est ce qui provoque potentiellement le rechargement de la webview et donc l'évènement `worker:webview:ready`. Avant ce changement, on écoutait l'évènement `worker:webview:ready` après avoir déclenché `SET_WORKER_STATE`, il y avait donc un risque (minime) de rater `worker:webview:ready` et de rester bloqué.

src/libs/bridge/ReactNativeLauncherMessenger.js
doubleface: These are different : 
 * this one hides post-me message related to the transmission of console.logs in the webviews
 * the one adding post-me check hides postMessage messages not related to post-me

I don't think we should merge them.

Globally, the need here is to make log messages related to post-me easier to scan visually and easier to link, making it easier to find messages related to a given session or a given request.

src/libs/bridge/ReactNativeLauncherMessenger.js
doubleface: As I say in this commit message, I only have this problem in scenarios when I watch the flagship app and with hot reload.

`addMessageListener` is called directly by post-me itself

src/libs/ReactNativeLauncher.js
Crash--: You can just test the ReactNativeLauncher, without the LauncherView. We can test that the events are well emitted. 

But OK to do that later.

src/libs/bridge/ReactNativeLauncherMessenger.js
Crash--: You should add this comment to the code directly. 👍

PR #787, opened 2023-05-15
fix: Use `trim()` on `tryProcessClouderyThemeMessage`

Ldoppea:
In some environments `getComputedStyle().getPropertyValue()` would return a result with a leading space

In this scenario we retrieve ` normal` instead of `normal`

So we want to `trim()` the result

Note that `getComputedStyle().getPropertyValue()` always return a string even if the property does not exist. In that cas the string is empty. So we don't need to check for nullity

We lost this security on feature rework when fetching Theme instead of Background


Ldoppea:
This has been merged without review so I can quickly push the fix on production, feel free to do a post-review

PR #788, opened 2023-05-16
feat: secure application if phone is unsecured

acezard:
Known issues:
- When going back with Android from a securitizing screen until you reach Android home, the application stays stuck on the loading screen until it gets killed. Redirection from HomeView to SecureView does not happen, even if the function is called to the router level.
- At the moment a user can be asked to set a new PIN code even if he already has one (no auto-lock set, PIN code set). What will happen at the time of this commit: the new PIN code will be ignored and the old one will be kept, the user will be redirected to the home page. Afterwards, his new PIN code will not work on the Lock screen that is now active, but the old one will still work. This is not good but there are multiple ways to solve this, we need to decide at the business-side level.

image


src/app/domain/authentication/models/User.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  return result.data.attributes.password_defined
+}
+
+const isInstanceSettingsData = (obj: unknown): obj is InstanceSettingsData => {
+  return typeof obj === 'object' && obj !== null && 'attributes' in obj
+}
+
+const isInstanceSettings = (item: unknown): item is InstanceSettings => {
```
Ldoppea, line 78: nit: the entire `isInstanceSettings` algorithm here seems to be generic to all `client.query()` responses isn't it? If yes we should keep that in mind and try to make a generic check method the next time we want to check a `client.query()` response

src/app/domain/authentication/models/User.ts
```diff
@@ -0,0 +1,137 @@
...
+  )
+}
+
+const IsVaultInstalled = async (client: CozyClient): Promise<boolean> => {
+  try {
+    const vault = await client
+      .getStackClient()
+      .fetchJSON<{ extension_installed?: boolean }>(
+        'GET',
+        '/data/io.cozy.settings/io.cozy.settings.bitwarden',
```
Ldoppea, line 29: Here we can client.query `Q('io.cozy.settings').getById('io.cozy.settings.bitwarden')` instead

Ldoppea: You can find an exemple here: https://github.com/cozy/cozy-flagship-app/blob/3381ab6ffb4c4bd80d2a00e983f21fd4e98a82d1/src/libs/functions/openApp.js#L59-L67

src/app/domain/authentication/models/User.ts
```diff
@@ -0,0 +1,137 @@
...
+import { ClientCapabilities } from 'cozy-client/types/types'
+import flag from 'cozy-flags'
+
+import { devlog } from '/core/tools/env'
+
+export const canAuthWithOIDC = (client: CozyClient): boolean | undefined => {
+  return isOidcAuth(client)
+}
+
+export const shouldCreatePassword = async (
```
Ldoppea, line 11: nit: As we are in a generic `User` model, I think the `shouldCreatePassword` is too specific to your onboarding prompt. I would prefer a `hasDefinedPassword` that only describe the user's state

Note that this would reverse the return statement:
```js
const passwordDefined = await isPasswordDefined(client)
const vaultInstalled = await IsVaultInstalled(client)

return (
    vaultInstalled || passwordDefined || (!isOidcAuth(client) && !isMagicLinkAuth(client))
)
```

This can be simplified with:
```js
return (
    vaultInstalled || passwordDefined || (!isPasswordlessCozy(client))
)

//...

const isPasswordLessCozy = (client) => {
    return isOidcAuth(client) || isMagicLinkAuth(client)
}
```

Ldoppea (commented):
nit: Some comments about naming:

- `view/Secure` is too generic name that can contain everything related to security
  - Here we wan't to have components responsible to advise the use to increase their security
  - Maybe we should rename this with something like `SecurityPrompt` or `SecuriseOnboarding` or `SecurityCTA` etc?

- Based on our previous architecture reflexions, I'm a bit lost about what should be in a model and what should be in a service

- Also `domain/authorization` seems to be more about `authentication` that `authorization`. If you want to keep the distinction from `domain/authentication` maybe we should rename this `domain/deviceAuthorization`?

- `devMode.ts` seems pretty generic, shouldn't we move it outside of `authorization` domain?

src/libs/localStore/storage.ts
```diff
@@ -37,6 +40,9 @@ export const getData = async <T>(name: StorageKeys): Promise<T | null> => {
   try {
     const value = await getItem(name)
 
+    if (value === 'true') return true as unknown as T
+    if (value === 'false') return false as unknown as T
```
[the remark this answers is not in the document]
acezard: I was having issues with the test suite for some reason, might look into it a bit more when I have time

src/libs/localStore/storage.ts
```diff
@@ -27,7 +27,10 @@ export const storeData = async (
   value: StorageItems[keyof StorageItems]
 ): Promise<void> => {
   try {
-    await setItem(name, JSON.stringify(value))
+    // Convert value to a string before storing it
+    const stringValue =
+      typeof value === 'boolean' ? String(value) : JSON.stringify(value)
```
Crash--, line 32: What is the benefit to do that? `JSON.stringify` already convert boolean to string 🤔

src/app/view/Lock/useLockScreen.ts
```diff
@@ -1,27 +1,33 @@
+import { AppState } from 'react-native'
 import { BiometryType } from 'react-native-biometrics'
+import { Route } from '@react-navigation/native'
```
Ldoppea, line 3:
```suggestion
import type { Route } from '@react-navigation/native'
```

src/app/domain/authorization/constants/index.ts
```diff
@@ -0,0 +1,3 @@
+export const authConstants = {
+  kdfIterations: 65e4 // 65*10^4 = 650000
```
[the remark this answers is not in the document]
acezard: Just a habit I started to take in personal projects. I'll modify it

- [ ] Updated

src/app/domain/authentication/models/User.ts
```diff
@@ -0,0 +1,137 @@
...
+  return (
+    !(await IsVaultInstalled(client)) &&
+    (isOidcAuth(client) || isMagicLinkAuth(client)) &&
+    !passwordDefined
+  )
+}
+
+const IsVaultInstalled = async (client: CozyClient): Promise<boolean> => {
+  try {
+    const vault = await client
```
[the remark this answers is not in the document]
acezard: This is hard to say as for me it could be in both. As I see it as a getter, I did put it in a model, it doesn't really have business logic of its own

src/app/domain/authentication/models/User.ts
acezard: - [ ] Updated

src/app/domain/authentication/models/User.ts
acezard: Sorry not sure to follow, I admit I used this code from cozy pass as is

src/app/domain/authorization/services/SecurityService.ts
```diff
@@ -0,0 +1,180 @@
...
+)
+
+export const determineSecurityFlow = async (
+  client: CozyClient
+): Promise<void> => {
+  if (await fns.isAutoLockEnabled()) {
+    devlog('🔒', 'Application has autolock activated')
+    devlog('🔒', 'Device should be secured or autolock would not work')
+
+    showLockView()
```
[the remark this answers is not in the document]
acezard: There was more logic initially. I guess it can be simplified now.

- [ ] Updated

src/libs/client.js
```diff
@@ -120,6 +120,26 @@ export const fetchCozyDataForSlug = async (slug, client, cookie) => {
   return result
 }
 
+/**
+ * @typedef {Object} getFqdnFromClient
+ * @property {string} fqdn - The fqdn is the raw host of the cozy instance (usually you don't want to use it)
+ * @property {string} normalizedFqdn - The normalizedFqdn is the fqdn without the port or other parts of the url (usually you want to use it)
```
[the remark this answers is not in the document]
acezard: - [ ] Updated

src/libs/client.js
```diff
@@ -128,7 +148,8 @@ export const getFqdnFromClient = client => {
 
   return {
     fqdn,
-    normalizedFqdn
+    normalizedFqdn,
+    uri: rootURL
```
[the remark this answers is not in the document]
acezard: Since we have rootUrl here, I think it's logical to export it at the same time. But then let's rename sure

- [ ] Updated

src/libs/localStore/storage.ts
acezard: It was tied to an issue I had during testing with the AsyncStorage mock. I should look into it again

- [ ] Investigate

src/libs/localStore/storage.ts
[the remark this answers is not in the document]
acezard: https://github.com/cozy/cozy-flagship-app/pull/788/files#r1199606935

src/utils/safePromise.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @returns A function that will call the async function and ignore any errors
+ *
+ * @example
+ * const asyncFn = async (arg1, arg2) => {
+ *  // do something
+ * }
+ * const safeAsyncFn = safePromise(asyncFn)
+ * safeAsyncFn(arg1, arg2) // will call asyncFn(arg1, arg2) and ignore any errors
+ */
+export const safePromise = <T extends unknown[]>(
```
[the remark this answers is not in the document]
acezard: - [ ] Updated

src/app/domain/authentication/models/User.ts
acezard: - [ ] Updated

Ldoppea (approved):

PR #789, opened 2023-05-16
Add userPersistedStorage

zatteo:
Add a wrapper to AsyncStorage supporting storage by user. So for a UserPersistedStorageKey, we can store a value for different user.

Every keys are prefixed with `@user_<user uri>_` so we can easily find every keys for an user.

It will be used in the backup feature to keep backup config when the user log out.

**What do you think about this feature ?**

**What do you think about this implementation ?**

I find it annoying to pass each time the client but I am not convinced by other solutions I found to store the client : 
- creating a hook like `const [getUserPersistedData, storeUserPersistedData] = useUserPersisedData(client)`
- initializing at log in an non React JS value in _src/libs/localStore/userPersistedStorage_
- checking each time the current user in AsyncStorage
- ???


src/libs/localStore/userPersistedStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+
+interface UserPersistedStorageItems {
+  backupConfig: object
+}
+
+const formatKey = (
+  client: CozyClient,
+  name: UserPersistedStorageKeys
+): string => {
+  return `@user_${client.getStackClient().uri}_${name}`
```
Ldoppea, line 17: I would prefer to use the "normalised" fqdn here. It is already used in some other places (i.e. bundle config) and it is easier to read (we skip the protocol and remove special chars)

https://github.com/cozy/cozy-flagship-app/blob/718f7be87fc5508e15b473819e6ce6912fa9f559/src/libs/client.js#L123-L124

src/libs/localStore/userPersistedStorage.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+) => Promise<unknown>
+const storeUserPersistedData = initialStoreUserPersistedData as (
+  client: CozyClient,
+  name: string,
+  value: unknown
+) => Promise<unknown>
+
+describe('userPersistedStorage', () => {
+  const aliceClient = {
+    getStackClient: () => ({ uri: 'alice.mycozy.cloud' })
```
Ldoppea, line 22: this is not representative of a real life uri, you should have the protocol here
```suggestion
    getStackClient: () => ({ uri: 'https://alice.mycozy.cloud' })
```

Also you can add a test case with a port
```
    getStackClient: () => ({ uri: 'https://alice.cozy.tools:8080' })
```

src/libs/localStore/userPersistedStorage.ts
zatteo: Much better thank you

Ldoppea (approved):

PR #792, opened 2023-05-17
fix: Target `<main>` instead of `<body>` for WebViews paddings injection

Ldoppea:
If we apply a padding to `<body>`, the padding will be applied only if the screen's content is smaller than the screen's size

When the content takes nearly the same size as the screen, then the padding would disappear and UI would exceed the screen's safe area without any scroll bar displayed (which is the problem we initially tried to resolve)

When the content is longer than the screen's size, then the scroll bar appears, but when scrolling to the bottom then the UI would still be outside of the safe area

To prevent this, we can target the `<main>` container instead. By chance this container is present on all our Cloudery screens. But in some case it already contains a CSS class that targets paddings (i.e the 2FA view), so we want to enforce the safe area padding

### DIFF
--- a/src/components/makeHTML.js
+++ b/src/components/makeHTML.js
@@ -23,8 +23,8 @@ const makeDOM = ({ body, isInverted }) => {
       <style type="text/css">${cirrusCss}</style>
     </head>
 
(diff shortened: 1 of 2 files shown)


Crash-- (approved):
If you say so, trusting you on this one.

PR #795, opened 2023-05-19
fix: Build error due to react-native-document-scanner-plugin

zatteo:
react-native-document-scanner-plugin uses a compileSdkVersion of 30 and we use 31 leading to build error.


Crash-- (approved):
@Ldoppea can you test it? Should fix our CI issue :)

Ldoppea (approved):

PR #796, opened 2023-05-19
Minimum viable backup

zatteo:
**Included** 
- manage permissions
- upload images and videos to a backup folder with your device name
- support incremental backup by comparing file name (comparison will be strengthened)
- store backup data in local storage
- do not lose backup data after local storage clean, log out and reinstall if another Cozy app is installed

Do not hesitate to challenge everything, especially code architecture, models, communication betweend front and back, ... There are surely bugs and edge cases not managed. I am mainly interested by the big picture to see if the chosen path is good.. 

Front end if you want to try : https://github.com/cozy/cozy-drive/pull/2922

**Known bugs** 
- uploaded video file is corrupted on Android

**Advices/notes** 
- better to read it commit by commit


src/app/domain/backup/services/getMedias.ts
```diff
@@ -0,0 +1,55 @@
+import { CameraRoll } from '@react-native-camera-roll/camera-roll'
+
+import { Media, BackupedMedia } from '/app/domain/backup/models/Media'
+import { getBackupConfig } from '/app/domain/backup/services/manageBackupConfig'
+
+import type CozyClient from 'cozy-client'
+
+export const getAllMedias = async (): Promise<Media[]> => {
+  const cameraRollPage = await CameraRoll.getPhotos({
+    first: 10000,
```
Crash--, line 10: 10k: you don't backup all my photos ^^'

zatteo:
I added pagination. If possible it would be nice to merge this PR in the backup master before starting to develop other backup features.

zatteo:
I reordered files and commits to offer a more explicit journey and be able to continue here without loosing in readability.

ios/CozyReactNative/Info.plist
```diff
@@ -144,5 +144,9 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>Pour faire un backup de vos photos</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
```
Ldoppea, line 149: The `Add` version seems to be redundant here, isn't it?

Cf https://developer.apple.com/documentation/bundleresources/information_property_list/nsphotolibraryusagedescription
and https://developer.apple.com/documentation/bundleresources/information_property_list/nsphotolibraryaddusagedescription

src/libs/localStore/userPersistedStorage.ts
```diff
@@ -5,7 +5,7 @@ import { normalizeFqdn } from '/libs/functions/stringHelpers'
 import { storeData, getData } from './storage'
 
 export enum UserPersistedStorageKeys {
-  BackupConfig = 'AmiralAppBackupConfig'
+  LocalBackupConfig = 'AmiralAppLocalBackupConfig'
```
Ldoppea, line 8: Why did you chance this? Won't we will lose some data on users' devices if we rename this without any migration methods? Or maybe this was not used yet?

src/app/domain/backup/services/manageBackup.ts
```diff
@@ -0,0 +1,124 @@
...
+export const prepareBackup = async (
+  client: CozyClient
+): Promise<BackupInfo> => {
+  log.debug('Backup preparation started')
+
+  await managePermissions()
+
+  const backupConfig = await initializeBackup(client)
+
+  if (backupConfig.currentBackup.status === 'running') {
```
Ldoppea, line 43: initializeBackup returns no `currentBackup` if the method throws and creates a new backupconfig, whill this throw?

src/app/domain/backup/services/managePermissions.ts
```diff
@@ -0,0 +1,40 @@
...
+    ? ('android.permission.READ_MEDIA_IMAGES' as Permission) // PermissionsAndroid.PERMISSIONS.READ_MEDIA_IMAGES is only available starting RN 0.70
+    : PermissionsAndroid.PERMISSIONS.READ_EXTERNAL_STORAGE
+
+export const hasMediaPermissions = async (): Promise<boolean> => {
+  if (Platform.OS === 'android') {
+    const hasPermission = await PermissionsAndroid.check(ANDROID_PERMISSION)
+    return hasPermission
+  }
+
+  return true
```
Ldoppea, line 19: Does this mean that iOS has no permission mechanism for media items?

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -0,0 +1,196 @@
...
+): Promise<boolean> => {
+  for (const mediaToUpload of mediasToUpload) {
+    try {
+      const uploadUrl = getDriveUrl(client, backupFolderId, mediaToUpload)
+
+      const { success } = await uploadMedia(client, uploadUrl, mediaToUpload)
+
+      if (success) {
+        log.debug(`✅ ${mediaToUpload.name} uploaded`)
+        await setMediaAsBackuped(client, mediaToUpload)
```
Ldoppea, line 35: nit: Adding another log after this line would be helpfull as I already imagine some looped synchro bug to be investigated :D

src/libs/localStore/userPersistedStorage.ts
zatteo: This was not used yet.

android/app/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,11 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="com.google.android.finsky.permission.BIND_GET_INSTALL_REFERRER_SERVICE" />
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
```
Crash--, line 9: Can we remove this write perm?

src/@types/cozy-client.d.ts
```diff
@@ -91,10 +91,34 @@ declare module 'cozy-client' {
...
+    findReferencedBy: (
+      params: object
+    ) => Promise<{ included: { attributes: unknown }[] }>
+    createDirectoryByPath: (path: string) => Promise<FileCollectionGetResult>
+    createDirectory: (attributes: object) => Promise<FileCollectionGetResult>
+    addReferencesTo: (references: object, dirs: object[]) => Promise<void>
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    launch: (trigger: any) => any
+    get: (id: string) => Promise<FileCollectionGetResult>
+  }
```
Crash--, line 115: Did you add these types to cozy-client itself?

src/app/domain/backup/models/Backup.ts
```diff
@@ -0,0 +1,46 @@
+import { Media, BackupedMedia } from '/app/domain/backup/models'
+
+type BackupStatus = 'to do' | 'initializing' | 'ready' | 'running' | 'done'
```
Crash--, line 3: I love the confidence that comes from not being afraid to use space!

src/app/domain/backup/models/Backup.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A local backup config representing the current backup of the device
+ * @member {string} remotePath
+ * @member {number} lastBackupDate
+ * @member {BackupedMedia[]} backupedMedias
+ */
```
Crash--, line 10: the JSDoc is not aligned with the content?

Crash--: Maybe `//@ts-check` can help?

src/app/domain/backup/models/Backup.ts
```diff
@@ -0,0 +1,46 @@
...
+  currentBackup: {
+    status: BackupStatus
+    mediasToBackup: Media[]
+  }
+}
+
+/**
+ * A remote backup config representing a backup in Cozy Drive
+ * @member {object} backupFolder
+ */
```
Crash--, line 24: same

src/app/domain/backup/models/Media.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @member {string} name
+ * @member {number} path
+ * @member {string} type
+ * @member {number} creationDate
+ */
+export interface Media {
+  name: string
+  path: string
+  type: 'image' | 'video'
+  creationDate: number
```
Crash--, line 12: no updateDate?

src/app/domain/backup/queries/index.ts
```diff
@@ -0,0 +1,21 @@
...
+const DOCTYPE_FILES = 'io.cozy.files'
+
+export const buildFilesQuery = (dirId: string): QueryDefinition => {
+  return Q(DOCTYPE_FILES)
+    .where({
+      dir_id: dirId,
+      type: 'file'
+    })
+    .indexFields(['dir_id', 'type'])
+}
```
Crash--, line 12: Files is a pretty big doctype. You should `.select()` only the fields you want 🙏

src/app/domain/backup/services/getMedias.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const backupConfig = await getLocalBackupConfig(client)
+
+  if (backupConfig === null) {
+    return []
+  }
+
+  const mediasToBackup = allMedias.filter(
+    allMedia => !isMediaAlreadyBackuped(allMedia, backupConfig.backupedMedias)
+  )
```
Crash--, line 78: I don't know, but having filter and some makes me think we're looping several times on potentially big arrays.

src/app/domain/backup/services/manageRemoteBackupConfig.ts
```diff
@@ -0,0 +1,148 @@
...
+  client: CozyClient
+): Promise<RemoteBackupConfig> => {
+  const deviceName = await getDeviceName()
+  const deviceId = await getDeviceId()
+
+  const {
+    data: { _id: backupRootId }
+  } = await client
+    .collection(DOCTYPE_FILES)
+    .createDirectoryByPath(BACKUP_ROOT_PATH)
```
Crash--, line 83: Can't you use `client.save()` directly? You should avoid calling `collection()`

src/app/domain/backup/services/uploadMedias.ts
Crash--, line 35: There is a lot of stuff done in `setMediaAsBackuped`: 
- We read from the localStorage 
- We map on all the backupedMedia array 
- We add the file to an array 
- We map on another array to remove the file to sync 
- We finally save the data to the localStorage

And we do that for every media we upload. Do we have an idea of the time taken just by this method during the first backup? 

At least the code and the path is readable and understandable, I just want to be sure that it'll not slow down too much the process

src/app/domain/backup/services/manageRemoteBackupConfig.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export const fetchBackupedMedias = async (
+  client: CozyClient,
+  deviceRemoteBackupConfig: RemoteBackupConfig
+): Promise<BackupedMedia[]> => {
+  const { backupFolder } = deviceRemoteBackupConfig
+
+  const filesQuery = buildFilesQuery(backupFolder.id)
+
+  const { data } = (await client.query(filesQuery)) as FilesQueryResult
```
Crash--, line 133: We'll only have the first 100 items with this query. If you really want all the files, you should use queryAll() or fetchMore()

Crash--: also, if you need only the filename, you should `select(['name'])`

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -0,0 +1,196 @@
...
+export const uploadMedias = async (
+  client: CozyClient,
+  backupFolderId: string,
+  mediasToUpload: Media[]
+): Promise<boolean> => {
+  for (const mediaToUpload of mediasToUpload) {
+    try {
+      const uploadUrl = getDriveUrl(client, backupFolderId, mediaToUpload)
+
+      const { success } = await uploadMedia(client, uploadUrl, mediaToUpload)
```
Crash--, line 31: There is no error you can get from this `uploadMedia` ?

Crash--: I find the API pretty strange. Does `uploadMedia` throws if an error occurs? If yes, then why are we checking the success? 

If not, how can we have the real error logged somewhere?

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -0,0 +1,196 @@
...
+const uploadMedia = async (
+  client: CozyClient,
+  uploadUrl: string,
+  media: Media
+): Promise<UploadMediaResult> => {
+  if (Platform.OS === 'android') {
+    return await uploadMediaAndroid(client, uploadUrl, media)
+  } else if (Platform.OS === 'ios') {
+    return await uploadMediaIOS(client, uploadUrl, media)
+  }
```
Crash--, line 57: I don't know how much size it adds to install the 2 plugins and whether they are treeshaked or not, but I would have done it this way: 
- create an uploadMedia.ios.js file 
- creation of an uploadMedia.android.js file 

And then "import uploadMedias from ./uploadMedia" and let the bundler take care of itself.

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -0,0 +1,196 @@
...
+        }`
+      },
+      notification: {
+        enabled: false
+      }
+    } as UploadOptions
+
+    RNBackgroundUpload.startUpload(options)
+      .then(uploadId => {
+        RNBackgroundUpload.addListener('error', uploadId, data => {
```
Crash--, line 90: would have called `data` `error`

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -0,0 +1,196 @@
...
+          if (data.responseCode === 201) {
+            resolve({ success: true, data })
+          } else {
+            reject({ success: false, data })
+          }
+        })
+      })
+      .catch(e => {
+        log.debug(e)
+        reject({ status: false, data: {} })
```
Crash--, line 106: You don't send the `e`

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -0,0 +1,196 @@
...
+        }
+      })
+      .catch(e => {
+        log.debug(e)
+        reject({ status: false, data: {} })
+      })
+  })
+}
+
+const getDriveUrl = (
```
Crash--, line 177: I would not call that `getDriveUrl`. `getUploadUrl` `getCozyUploadUrl` or what you want, but there is nothing related to the `Drive` app in this method

android/app/src/main/AndroidManifest.xml
zatteo: Done

src/app/domain/backup/models/Backup.ts
zatteo: 😅

Changed to 'to_do' just in case 😄

src/app/domain/backup/models/Media.ts
zatteo: Not for the moment but it will come when I will improve file identification. I choose to add field only when needed.

src/app/domain/backup/services/uploadMedias.ts
zatteo: I just discovered platform specific extensions 😱 Very nice!

src/app/domain/backup/services/managePermissions.ts
zatteo: Permissions is asked automatically at first backup on iOS. On Android, we need to manage it manually. That's why we have this difference in the code. Anyway I do not manage when the user answers no so I will ask what is the expected behavior when the user say no and modify this part.

And if we want to manage permission manually on iOS we will need to add something like https://github.com/zoontek/react-native-permissions

src/app/domain/backup/services/uploadMedias.ts
zatteo: **Quick test**

Around 10ms per call on an iPhone SE 2020 with 10 files.
Around 20ms per call on an iPhone SE 2020 with 200 files.

So localStorage is not an issue but the multiple map could be with a lot of medias.

src/@types/cozy-client.d.ts
Crash--: As long as it's done ;)

src/app/domain/backup/services/manageRemoteBackupConfig.ts
Crash--: nit: Oh yeah didn't see that it was `createDirectoryByPath`. Maybe we can update CozyClient to accept that one: 

```
 async create(attributes) {
    if (attributes.type === 'directory') {
      return this.createDirectory(attributes)
    } else {
      const { data, ...createFileOptions } = attributes
      return this.createFile(data, createFileOptions)
    }
  }

async createDirectory(...){
if(params.path){
  createDirectoryByPath()
}
...
}
```

src/app/domain/backup/services/uploadMedias.ts
Crash--: Great. Let's go that way and see how it performs later?

src/app/domain/backup/services/manageRemoteBackupConfig.ts
```diff
@@ -0,0 +1,151 @@
...
+    [
+      {
+        _id: backupFolderId
+      }
+    ]
+  )
+
+  const { data: backupFolder } = await client
+    .collection(DOCTYPE_FILES)
+    .get(backupFolderId)
```
Crash--, line 114: here! You can use client.get() :D

src/app/domain/backup/services/manageRemoteBackupConfig.ts
```diff
@@ -0,0 +1,151 @@
...
+    name: deviceName,
+    dirId: backupRootId,
+    metadata: { backupDeviceIds: [deviceId] }
+  }
+
+  const {
+    data: { _id: backupFolderId }
+  } = await client
+    .collection(DOCTYPE_FILES)
+    .createDirectory(backupFolderAttributes)
```
Crash--, line 98: and here, you can use client.save()

src/app/domain/backup/services/manageRemoteBackupConfig.ts
zatteo: Ok! I prefer to keep it that way for the moment. Take time to modify cozy-client and createDirectoryByPath is kind of very explicit.

src/app/domain/backup/services/uploadMedias.ts
zatteo: Yes! I will create a task to check performance with 1000+ and 5000+ medias and stay vigilant about big arrays. I'm also relieved about local storage now because I use it extensively...

src/app/domain/backup/services/manageRemoteBackupConfig.ts
zatteo: How? There is logic inside createDirectory so I will bypass it no?

src/app/domain/backup/services/manageRemoteBackupConfig.ts
zatteo: Updated with ` Q(DOCTYPE_FILES).getById(id)` and client.query

src/app/domain/backup/services/manageRemoteBackupConfig.ts
zatteo: Answered : _client.save({_type: 'io.cozy.files', type:'directory', ...})_ calls _create_ which call _createDirectory_.

Crash-- (approved):

PR #799, opened 2023-05-22
feat: securize phone interface adjustments

acezard:
## UI Fixes
- [x] Prompt views for Password and PinCode have been harmonized
- [x] Prompt views for Password and PinCode have been revised to implement correct text styles (boldness, size, margin, max-width)
- [x] "Securize" icon replaced the placeholder CozyIcon on those views
- [x] Primary and Secondary Button styles have been updated to correspond to Figma specs
- [x] Added `KeyboardAvoidingView` in SetPin page

## Flow fix
- [x] Instead of activating autolock if the user accepts to define a PinCode, we will activate it as soon as he reaches the Pin Suggestion (prompt) view (https://github.com/cozy/cozy-flagship-app/pull/799/files#diff-5df3919ccc35b89183a1f7b1813f9205d2e5018d85a09edc93902d1a0c35b489R19)

## Screenshots
image
image


src/ui/Button/styles.ts
```diff
@@ -11,10 +11,12 @@ export const styles = StyleSheet.create({
   },
   primary: {
     backgroundColor: palette.Common.white,
-    borderColor: 'rgba(29, 33, 42, 0.16)',
+    borderColor: 'rgba(255, 255, 255, 1)',
```
Ldoppea, line 14: Someday we will have to transfert this color into the palette or into a theme dedicated file

Ldoppea (approved):

src/ui/Button/styles.ts
acezard: @Ldoppea Absolutely, and this applies to many other values. It was intended that the current feature with "dynamic theming" would rework these values and erase any magic number used in the app.

PR #806, opened 2023-05-24
fix: add hideSplashScreen fallbacks

acezard:
Calling hideSplashScreen everywhere since it seems the HomeView fails to do it in some unknown scenarios.

Not a good long-term solution, but we need it now. Especially as it can't create side effects even if the home view hides the splash screen.

Some additional minor refactor/improvements that do not touch user/data flow have been added

### DIFF
--- a/src/app/domain/authorization/services/SecurityService.ts
+++ b/src/app/domain/authorization/services/SecurityService.ts
@@ -27,7 +27,6 @@ import { safePromise } from '/utils/safePromise'
 import { navigateToApp } from '/libs/functions/openApp'
 
 // Can use mock functions in dev environment
-// async (): Promise<boolean> => Promise.resolve(false)
 const fns = getDevModeFunctions(
   {
     isDeviceSecured,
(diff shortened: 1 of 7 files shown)


src/app/view/Lock/useLockScreen.ts
```diff
@@ -111,6 +112,13 @@ export const useLockScreenProps = (props: LockScreenProps): LockViewProps => {
     [biometryEnabled]
   )
 
+  // The HomeView should have called hideSplashScreen() already,
+  // but in case it didn't, we do it here as a fallback as it is critical.
```
Ldoppea, line 116: I think those case are when a cozy-app is awaited by the home (cozyAppFallback), in that case no hideSplashScreen is called until the cozy-app is rendered. And because you stopped the navigation to the cozy-app to display the secure prompt, then you have to do it by yourself

Ldoppea (approved):

src/app/domain/authorization/services/SecurityService.ts
```diff
@@ -27,17 +27,16 @@ import { safePromise } from '/utils/safePromise'
...
     isAutoLockEnabled,
     hasDefinedPassword
   },
   {
-    isDeviceSecured: undefined,
-    isAutoLockEnabled: undefined,
-    hasDefinedPassword: undefined
+    isDeviceSecured: undefined, // async (): Promise<boolean> => Promise.resolve(false),
+    isAutoLockEnabled: undefined, // async (): Promise<boolean> => Promise.resolve(false),
+    hasDefinedPassword: undefined // async (): Promise<boolean> => Promise.resolve(false)
```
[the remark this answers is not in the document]
acezard: I let them to show how to use them

src/app/view/Lock/useLockScreen.ts
acezard: Ok, then these calls seem needed

PR #812, opened 2023-05-31
fix: move "hasRenderedOnce" into an useRef()

acezard:
It should handle better login/unlogin bad persistence
in "hasRenderedOnce" state.

There was a bug on onboarding where the home would never show up (seemingly),
or at least the phoneSecurity check was never called.
It is still unsure how this scenario arises but it's suspected it happens when creating
accounts in a row without leaving the application.
In that case, the scoped variable would stay to true, which is technically correct,
but functionally we prefer the home to think the app has rebooted since the 
last login.
By using this in an useRef(), we're hoping the blue screen issue will be fixed
since the home state should refresh itself more naturally than with
a hardcoded let value in the homeview file.

### DIFF
--- a/src/app/view/Lock/useLockScreen.ts
+++ b/src/app/view/Lock/useLockScreen.ts
@@ -21,6 +21,7 @@ import {
   tryUnlockWithPassword,
   validatePin
 } from '/app/domain/authorization/services/LockScreenService'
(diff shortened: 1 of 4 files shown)


acezard:
closing since the commit has been rebased into https://github.com/cozy/cozy-flagship-app/pull/815 for consistency

PR #816, opened 2023-06-01
Feat/new icon and popup

zatteo:
- Add a new icon (mabulle.png)
- Add a modal displayed only on Android to inform the user that the icon has changed
- Use flags to allow icon change and set default icon remotely

image

#### Checklist

* [x] Tested on iOS
* [x] Tested on Android
* [x] All changes have test coverage
* [x] Updated README & CHANGELOG, if necessary


src/libs/icon/icon.spec.ts
```diff
@@ -1,21 +1,50 @@
...
+  'flagship.icon.defaultIcon': 'cozy'
+}
+
+const ICON_CHANGE_NOT_ALLOWED_FLAGS = {
+  'flagship.icon.changeAllowed': false
+}
+
+const ICON_CHANGE_UNDEFINED_FLAGS = {
+  'flagship.icon.changeAllowed': false
+}
```
Crash--, line 27: are they documented in the wiki?

patches/react-native-change-icon+4.0.0.patch
```diff
@@ -1,12 +1,12 @@
...
-             promise.reject("ICON_INVALID");
+@@ -83,6 +83,7 @@ public class ChangeIconModule extends ReactContextBaseJavaModule implements Appl
              return;
          }
-+        this.classesToKill.clear();
          this.classesToKill.add(this.componentClass);
++        this.classesToKill.remove(activeClass); // remove the new icon from the list of classes to kill to avoid being unable to start the app
          this.componentClass = activeClass;
          activity.getApplication().registerActivityLifecycleCallbacks(this);
+         iconChanged = true;
```
Crash--, line 12: is is necessary this var?

Crash-- (approved):

patches/react-native-change-icon+4.0.0.patch
zatteo: This var is not added in "my" patch. I updated the patch package 1 line below the previous patch package, so it took one line more at then end.

src/libs/icon/icon.spec.ts
zatteo: Now, yes 😄

PR #818, opened 2023-06-01
fix: Do not change the user agent on pilot webview 

doubleface:
When a clisk konnector called the `setUserAgent` function, it changed
the user agent on worker and pilot webviews.

It seems that this change of user agent on the pilot webview is causing
the pilot to loose the course of it's execution, making it wait forever.

Since now we do not download any file from the pilot itself, I think
removing the user agent from the pilot webview has no impact.

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/screens/konnectors/LauncherView.js
+++ b/src/screens/konnectors/LauncherView.js
@@ -255,7 +255,6 @@ class LauncherView extends Component {
                 }}
                 useWebKit={true}
                 javaScriptEnabled={true}
-                userAgent={this.state.userAgent}
                 sharedCookiesEnabled={true}
                 onMessage={this.onPilotMessage}
                 onError={this.onPilotError}


Crash--:
> It seems that this change of user agent on the pilot webview is causing
the pilot to loose the course of it's execution, making it wait forever.

This is because the props change so the webview is re-rendered, right?

doubleface:
@Crash-- 

> This is because the props change so the webview is re-rendered, right?

No, I forgot to mention it, but there is no trace of a webview reload in this case. I actually have no good explanation on why the pilot looses the course of the execution. It just stops showing logs, sometimes, after running the "setUserAgent" command. And I could not reproduce this behavior 10 times in a row when removing the userAgent state property from the pilot webview.

doubleface:
@Crash-- Rerender, but not reload actually

PR #824, opened 2023-06-06
Improve App Security with Back Button and AppState Listeners

acezard:
In this PR, we've added additional layers of security to our React Native app to handle Android back button presses and App state changes more effectively. 

Specifically, we've:

- Created a `SecurityNavigationService` that listens for Android back button presses and exits the app if the user is currently on one of the following routes: lock screen, PIN prompt, or password prompt. This prevents users from navigating back to the previous route without successful authentication. 

- Added an AppState listener to the `SecurityNavigationService`. This ensures that we only listen for back button presses when the app is active (i.e., in the foreground). When the app goes to the background, the listener is removed to prevent unnecessary usage of resources.

- Added comprehensive unit tests to ensure the new service behaves as expected.

This change enhances the security of our app by ensuring that sensitive information cannot be accessed without proper authentication and improves resource usage by only listening for back button presses when necessary.

## Testing

Tests have been added for the `SecurityNavigationService`. You can run these tests using the command `yarn test`.

Please also manually test this feature by trying to use the Android back button to navigate from the lock, PIN prompt, and password prompt screens.

## Impact

This change impacts any screens that use the lock, PIN prompt, and password prompt screens. The back button on Android will now behave differently on these screens, improving the security of the app.

### DIFF
--- a/src/app/domain/authorization/services/SecurityService.ts
+++ b/src/app/domain/authorization/services/SecurityService.ts
@@ -26,6 +26,7 @@ import { authConstants } from '/app/domain/authorization/constants'
 import { safePromise } from '/utils/safePromise'
 import { navigateToApp } from '/libs/functions/openApp'
 import { hideSplashScreen } from '/libs/services/SplashScreenService'
+import { SecurityNavigationService } from '/app/domain/authorization/services/SecurityNavigationService'
 
 // Can use mock functions in dev environment
 const fns = getDevModeFunctions(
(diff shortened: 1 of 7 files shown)


src/app/domain/authorization/services/SecurityService.ts
```diff
@@ -35,9 +36,9 @@ const fns = getDevModeFunctions(
     hasDefinedPassword
   },
   {
-    isDeviceSecured: undefined, // async (): Promise<boolean> => Promise.resolve(false),
-    isAutoLockEnabled: undefined, // async (): Promise<boolean> => Promise.resolve(false),
-    hasDefinedPassword: undefined // async (): Promise<boolean> => Promise.resolve(false)
+    isDeviceSecured: async (): Promise<boolean> => Promise.resolve(false),
+    isAutoLockEnabled: async (): Promise<boolean> => Promise.resolve(false),
+    hasDefinedPassword: async (): Promise<boolean> => Promise.resolve(true)
```
zatteo, line 41: This should be let to undefined no ?

acezard: fixed

PR #826, opened 2023-06-07
Fix: Handle Offline on CozyApp Open

acezard:
This PR introduces several improvements to our offline handling strategy. The main changes include:

- **Redirect to Home instead of Stack:** Previously, when the device was offline, the app would redirect the user to the Stack when online again. To provide an updated user experience, we've changed this functionality to redirect users to the Home screen instead.

- **Graceful Unsubscribe Event Handling:** This update ensures that the `unsubscribe` event is handled more gracefully, preventing any potential issues that could arise from calling it while undefined.

- **Removal of Navigation Reset:** We've removed the use of the navigation reset function (`reset()`) because it was causing destructive effects. Instead, we've opted for less disruptive methods of navigation. Destructive effects were linked to the unmounting of the HomeView, which provokes a lot of issues with post-me. Navigating avoid these issues entirely and the home stays rendered. Better performance and fewer issues.

- **Addressing Potential Backpress Issues:** We have started investigating how to handle scenarios where a user triggers a back press from the Offline Screen, especially as we have removed the navigation reset function. This is a complex area that we will continue to explore and improve upon in future updates.

- **Migration to TypeScript:** All files related to `CozyAppScreen` have been migrated from JavaScript to TypeScript. This change not only provides more robust type-checking and tooling support but also makes the codebase more consistent and maintainable. All types have been fully defined to ensure strict type checking. 

This PR provides a safer and more user-friendly approach to handling offline scenarios, enhancing the overall performance and reliability of our app.

### DIFF
--- a/src/libs/services/NetService.spec.ts
+++ b/src/libs/services/NetService.spec.ts
@@ -1,12 +1,10 @@
 import NetInfo, { NetInfoState } from '@react-native-community/netinfo'
 
 import strings from '/constants/strings.json'
-
-import { NetService } from './NetService'
-
+import { NetService } from '/libs/services/NetService'
 import { routes } from '/constants/routes'
 
-const mockReset = jest.fn<jest.Mock, unknown[]>()
+const mockNavigate = jest.fn<jest.Mock, unknown[]>()
 const callbackRoute = 'foo'
 const mockedNetInfo = NetInfo as jest.Mocked<typeof NetInfo>
 
@@ -18,7 +16,7 @@ const mockedListener = mockedNetInfo.addEventListener as unknown as jest.Mock<
 const mockedFetch = NetInfo.fetch as jest.Mock
 
 jest.mock('../RootNavigation', () => ({
-  reset: (...args: unknown[]): jest.Mock => mockReset(...args)
+  navigate: (...args: unknown[]): jest.Mock => mockNavigate(...args)
 }))
 
 afterEach(() => {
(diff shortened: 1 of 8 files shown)


src/libs/services/NetService.ts
```diff
@@ -38,23 +40,26 @@ export const useNetService = (client?: CozyClient): void =>
...
     if (state.isConnected) {
       log.debug('Removing NetInfo listener')
 
       callbackRoute === routes.stack && showSplashScreen().catch(log.error)
 
       try {
-        reset(callbackRoute)
+        navigate(callbackRoute)
       } catch (error) {
         log.error(error)
```
acezard, line 52: handle error by navigating to safe route

src/libs/services/NetService.spec.ts
```diff
@@ -1,12 +1,10 @@
 import NetInfo, { NetInfoState } from '@react-native-community/netinfo'
 
 import strings from '/constants/strings.json'
-
-import { NetService } from './NetService'
-
+import { NetService } from '/libs/services//NetService'
```
zatteo, line 4: Two `/`?

zatteo:
What do you mean by `unsure what happens in some scenarios when user backpress,
since we have removed reset navigation`? Do we need to do something about this ?

src/libs/services/NetService.spec.ts
acezard: good catch, weird that I didn't have a failing test, or the CI

acezard:
> What do you mean by `unsure what happens in some scenarios when user backpress, since we have removed reset navigation`? Do we need to do something about this ?

well that's the thing, I'm not sure if all scenarios are handled, even if I think I covered them. Only thing to do here since no functional test is to manually test every scenario

src/libs/services/NetService.spec.ts
acezard: updated

zatteo (approved):

zatteo:
Just my opinion : first commit would be easier to read if split in two, one for ts conversion and on for moving functions in new files.

src/libs/services/NetService.ts
```diff
@@ -38,33 +39,36 @@ export const useNetService = (client?: CozyClient): void =>
...
       }
 
+      log.debug('NetService redirect done, removing NetInfo listener')
+      // Unsubscribe from the NetInfo event listener, as we're now online
```
Ldoppea, line 56: nit: seems like this comment is redundant since we already have a log explaining what's happening?

Ldoppea (approved):

acezard:
> Just my opinion : first commit would be easier to read if split in two, one for ts conversion and one for moving functions in new files.

ok I'll see how to manage that next time I refactor

PR #829, opened 2023-06-12
fix: forcing the light theme for the app

acezard:
In order to avoid issues with unhandled dark modes,
we disable the DayNight compat of the app
and replace it with a Light compat only.
This will disable algorithmic darkening

Tested with forceDarkMode on WebView on a Android phone in dark theme with a dark themed chrome: the homepage is not dark despite not having a meta color-scheme

### DIFF
--- a/android/app/src/main/res/values/styles.xml
+++ b/android/app/src/main/res/values/styles.xml
@@ -1,8 +1,10 @@
 <resources>
     <!-- Base application theme. -->
-    <style name="AppTheme" parent="Theme.AppCompat.DayNight.NoActionBar">
-        <!-- Customize your theme here. -->
-        <item name="android:enforceNavigationBarContrast">false</item>
+    <!-- Set the base theme to light for the entire app -->
+    <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
+        <!-- Additional customizations can be added here -->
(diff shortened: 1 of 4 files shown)


android/app/src/main/res/values/styles.xml
```diff
@@ -1,6 +1,7 @@
 <resources>
     <!-- Base application theme. -->
-    <style name="AppTheme" parent="Theme.AppCompat.DayNight.NoActionBar">
+    <!-- Forcing the light theme for the entire app -->
+    <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
```
Crash--, line 4: Do you know the difference between that and        ` <item name="android:forceDarkAllowed">false</item>` ?https://github.com/facebook/react-native/issues/27876

acezard: @Crash-- from what I understand we should use both to be sure. They seem to do the same thing but in a different way (to me). It's possible some devices may work or not in one mode or another.

acezard: updated

Crash-- (approved):

acezard:
> @acezard Let's merge it. Once it's merged, can you send me a built APK please? Like the one we send to the play store please

ok

PR #833, opened 2023-06-14
feat: Download files progressively only when needed 

doubleface:
Now call cozy-clisk saveFiles directly for all the given files and pass
a callback which saveFiles will call when the download of a file is
needed. This downloadAndFormatFile callback downloads the given file in
the worker webview and format it to ArrayBuffer as expected by the cozy
stack.

See https://github.com/konnectors/libs/pull/949 for the related
cozy-clisk PR

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/screens/login/components/ClouderyViewSwitch.tsx
+++ b/src/screens/login/components/ClouderyViewSwitch.tsx
@@ -226,7 +226,7 @@ const ClouderyWebView = forwardRef(
         onOpenWindow={openWindowWithInAppBrowser}
         injectedJavaScriptBeforeContentLoaded={run}
         style={{
-          backgroundColor: clouderyTheme?.backgroundColor
+          backgroundColor: clouderyTheme.backgroundColor
         }}
         onError={handleError}
         {...other}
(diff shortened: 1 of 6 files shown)


src/screens/login/components/ClouderyViewSwitch.tsx
```diff
@@ -226,7 +226,7 @@ const ClouderyWebView = forwardRef(
         onOpenWindow={openWindowWithInAppBrowser}
         injectedJavaScriptBeforeContentLoaded={run}
         style={{
-          backgroundColor: clouderyTheme?.backgroundColor
+          backgroundColor: clouderyTheme.backgroundColor
```
Crash--, line 229: can you remove this change? (or make in a new commit) cc @Ldoppea

Crash-- (approved):
Split change before commit

src/screens/login/components/ClouderyViewSwitch.tsx
doubleface: Done

PR #834, opened 2023-06-15
feat: Intercept Manager related UniversalLinks

Ldoppea:
> **Note**
> All commits contain their own description with context so feel free to review commit by commit

With current Onboarding and Login implementation, the user receives an email containing a link to the Cloudery Manager. Then when clicked, this link is opened into the device's Browser. Then the opened Manager page redirects into a Universal Link that should trigger the Flagship app with the corresponding scenario

In some environments, this scenario may not work. This is the case when the email's link is opened into the Mailer InAppBrowser or when the OS Browser is Firefox (by default this browser won't open external apps)

To mitigate this, we choose to convert the email's link into an UniversalLink that has the original link into a `fallback` parameter

With this implementation, the Flagship app should be triggered whatever is the environment, as we don't rely in a Browser anymore

However when opened by the Flagship app, we now need to open the `fallback` parameter into a webview so the Manager can generate a new UniversalLink (which is the same one we got in the original scenario). Then we intercept this UniversalLink and continue the Onboarding/Login scenario the same way we did before


src/app/domain/deeplinks/services/DeeplinksParserService.ts
```diff
@@ -5,6 +5,10 @@ import { MagicLinkUrl } from '/app/domain/deeplinks/models/MagicLinkUrl'
 import { OnboardingParams } from '/app/domain/deeplinks/models/OnboardingParams'
 import { getErrorMessage } from '/libs/functions/getErrorMessage'
 
+import { BootstrapAction } from '../../bootsrap/model/BootstrapAction'
```
Crash--: nit: import

src/app/view/Manager/ManagerScreen.tsx
```diff
@@ -20,6 +20,7 @@ export const ManagerScreen = (props: ManagerScreenProps): React.ReactNode => {
             managerProps.onShouldStartLoadWithRequest
           }
           onError={managerProps.handleError}
+          onHttpError={managerProps.handleHttpError}
```
Crash--, line 23: Note Android API minimum level 23.

Crash--: Et pour l'instant on target 22.

src/components/makeErrorPage.js
```diff
@@ -65,6 +72,15 @@ export const makeErrorPage = ({ icon, title, body, footer, header }) => {
             window.ReactNativeWebView.postMessage(JSON.stringify({
               message: 'mailto'
             }))
+
+          const resetButton = document.getElementById('resetButton')
+          if (resetButton) {
+            resetButton.onclick = (e) => {
+              console.log('HEHEOL')
```
Crash--, line 86: to remove

Crash-- (approved):

src/components/makeErrorPage.js
Ldoppea: Removed

PR #838, opened 2023-06-19
Add keep awake API

zatteo:
Add react-native-keep-awake and wrap it in a service. Tested on Android simulator and iOS device.

This API **works** when app is **in foreground** and **until app is quit**.

#### How to use it 

```
import {
  activateKeepAwake,
  deactivateKeepAwake
} from '/app/domain/sleep/services/sleep'


...

activateKeepAwake()

// long task where you do not want to go to sleep

deactivateKeepAwake()

...
```

#### Checklist

* [ ] ~~Faithful integration of the mockups at all screen sizes~~
* [x] Tested on iOS
* [x] Tested on Android
* [x] Localized in English and French
* [ ] ~~All changes have test coverage~~
* [ ] ~~Updated README & CHANGELOG, if necessary~~

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -366,6 +366,8 @@ PODS:
   - react-native-gzip (2.0.0):
     - NVHTarGzip
     - React
+  - react-native-keep-awake (1.1.0):
+    - React-Core
   - react-native-netinfo (9.3.7):
     - React-Core
   - react-native-safe-area-context (4.5.0):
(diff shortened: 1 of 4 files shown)


Crash--:
Is it supported by all our targeted versions? No need of a new permissions?

zatteo:
> Is it supported by all our targeted versions? No need of a new permissions?

Good question : 
- iOS > 8
- android minSdk = 16
- android compileSdk = 27

So everything seems fine.

No permissions needed.

Crash--:
Another question @zatteo on this subject: 

Does this change is persisted accros reboot? If, for example, we disable sleep mode and forget / or not able to restore it (because of an app crash, for example), won't that cause the app to always block sleep mode on restart?

zatteo:
If you go to background and foreground -> change is persisted, sleep timer is restarted when you go to foreground
If you quit and restart -> change is **not** persisted


For Android and iOS

Crash--:
Thanks for the info.

> sleep timer is restarted when you go to foreground

What does that mean?

zatteo:
If your iPhone is set up to go to sleep after 30s : 
- you start the app
- you do nothing for 30s
- then it goes to sleep

- you start the app
- you do nothing for 15s
- app goes to background
- app goes to foreground
- you do nothing for 30s 
- then it goes to sleep (whatever the 15s before which is common sense)

PR #841, opened 2023-06-22
Handle Non-HTTP or HTTPS URL Navigation

acezard:
**Changes:**
Introduced _isNotHttpOrHttps helper function to check if a URL does not start with HTTP or HTTPS. This function is useful when we need to decide whether we should use the standard HTTP/HTTPS navigation or use a different navigation strategy for other protocols (such as FTP, mailto, etc.). This function will return true if the URL is not HTTP or HTTPS, and false otherwise.

Created _handleNonHttpOrHttps function to handle URL navigation for non-HTTP/HTTPS URLs. If a URL does not start with HTTP or HTTPS, we will try to open the URL using React Native's Linking.openURL function. If the URL opening fails, we will log the error using the devlog utility.

**Testing:**
Unit tests have been added for the _isNotHttpOrHttps function to ensure it correctly identifies HTTP, HTTPS, and non-HTTP/HTTPS URLs.

In the unit tests for the _handleNonHttpOrHttps function, we've ensured that React Native's Linking.openURL is called when appropriate, and any errors during URL opening are correctly caught and logged.

**Impact:**
These changes will improve our ability to handle non-HTTP/HTTPS URLs in our app, and also improve error logging in cases where URL opening fails. This will improve the overall user experience when interacting with URLs within our app.


src/app/domain/navigation/webviews/UrlModels.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface IWebView {
+  injectJavaScript: (script: string) => void
+}
+
+export interface InterceptNavigationProps {
+  initialRequest: { url: string; isTopFrame: boolean }
+  targetUri: string
+  subdomainType: 'flat' | 'nested'
+  navigation: NavigationProp<Record<string, object | undefined>>
+  onShouldStartLoadWithRequest: (request: {
```
Ldoppea, line 19:
```ts
import type {  WebViewNavigation } from 'react-native-webview/lib/WebViewTypes'

onShouldStartLoadWithRequest (request: WebViewNavigation) => boolean
```

src/app/domain/navigation/webviews/UrlService.ts
```diff
@@ -0,0 +1,180 @@
...
+        slug: newSlug
+      })
+    } catch (error) {
+      webviewUrlLog.warn('Could not navigate to app', error)
+    }
+
+    return
+  }
+
+  if (_isNotHttpOrHttps(destinationUrl)) {
```
Ldoppea, line 167: Why are you using the `_XXX` notation?

src/app/domain/navigation/webviews/UrlUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Handle Non HTTP or HTTPS urls.
+ * Be careful, this function expects a protocol that is not HTTP or HTTPS.
+ * @param url - The URL to handle
+ */
+export const _handleNonHttpOrHttps = async (url: string): Promise<void> => {
+  try {
+    await Linking.openURL(url)
+  } catch (e: unknown) {
+    if (e instanceof Error) {
+      devlog(`Could not open url ${url}`, e)
```
Ldoppea, line 24: We should use `Minilog.error()` here, otherwise the error won't be logged in Sentry

src/app/domain/navigation/webviews/UrlModels.ts
```diff
@@ -0,0 +1,36 @@
+import { NavigationProp } from '@react-navigation/native'
+import { RefObject } from 'react'
+
+import CozyClient from 'cozy-client'
+
+/**
+ * Interface to define the structure of WebView
+ * Developers can add new functions as required by the WebView
+ */
+export interface IWebView {
```
Ldoppea, line 10:
```js
import type { WebView } from 'react-native-webview'
```

Ldoppea (approved):
Looks good to me

All my comment are nit except for the minilog comment

src/app/domain/navigation/webviews/UrlService.ts
Ldoppea: If you rename the method `openUrlWithOS()` or similar, then it is meant to be exported and can be easily used by other files 😛

src/app/domain/navigation/webviews/UrlModels.ts
acezard: 👍

src/app/domain/navigation/webviews/UrlUtils.ts
acezard: 👍

src/app/domain/navigation/webviews/UrlModels.ts
acezard: 👍

src/app/domain/navigation/webviews/UrlService.ts
acezard: 👍

PR #844, opened 2023-06-22
feat: Add logs to ease debugging

Ldoppea:
This PR adds some logs to ease debugging. Those are basically logs on functions entry but also logs on mount/unmount for CozyAppScreen and HomeView

### DIFF
--- a/src/components/webviews/CozyProxyWebView.js
+++ b/src/components/webviews/CozyProxyWebView.js
@@ -1,3 +1,4 @@
+import Minilog from '@cozy/minilog'
 import React, { useState, useEffect } from 'react'
 import { Platform, View } from 'react-native'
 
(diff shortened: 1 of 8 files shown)


Crash--:
How are we going to access these logs?

Ldoppea:
They would be visible in the extracted file if we re-use this PR someday: https://github.com/cozy/cozy-flagship-app/pull/325

Until then, today the only way to access them is from the dev console, and maybe from Sentry if they are followed by an error (however I'm not sure about the exact rule here).

Example of a Sentry log that contains an exception and show all previous `log.debug()` calls:
image

Crash--:
Yep but in our case, we don't have any raised error, so the debug log will not be sent. So with this, ATM, it's not useful, no?

What about creating a flag to activate on the cozy that sent these debug logs for this feature?

Ldoppea:
I'm not sure sending all debug logs to Sentry would be a good idea (this would mean a lot of new Sentry issues that would be difficult to classify)

However maybe we can log those on a file (only if the flag exists as it may reduce performances) and then add an option to download/share the file from cozy settings?

If we do this we should improve the file logging a bit to make sure that the file won't grow too big

Crash--:
So as for today, this PR is not useful. Right? If yes, should we close it?

Crash--:
ping @Ldoppea

Crash--:
ping @Ldoppea

Issue #411, opened 2022-09-06
Can we have better WebView composition?


Ldoppea:
By reading [this PR](https://github.com/cozy/cozy-react-native/pull/402/files#diff-94ee99771b8d7d089bb046a665823be8eedf5c5a317115debff9a97a69638ce4) I'm wondering how we can handle all features we add to our WebViews (here `CozyWebView`, but also `ReloadInterceptorWebView`) in a more composable way.

Today we use to handle this in 2 different ways:
- We add more complexity to an existing WebView
- We create a new feature-specific WebView and add it somewhere into our CozyWebView component tree

Today we have:
```jsx
<HomeView>
  <CozyProxyWebView>
    <CozyWebView>
      <ReloadInterceptorWebView>
        <SupervisedWebView>
          <WebView>
        </SupervisedWebView>
      </ReloadInterceptorWebView>
    </CozyWebView>
  </CozyProxyWebView>
</HomeView>
```

In the [PR mentioned above](https://github.com/cozy/cozy-react-native/pull/402/files#diff-94ee99771b8d7d089bb046a665823be8eedf5c5a317115debff9a97a69638ce4), we chose to add complexity to the existing CozyWebView.

But if we chose the second option (feature-specific WebView), then we would have to inject it somewhere in the component tree. For example:
```diff
<HomeView>
  <CozyProxyWebView>
    <CozyWebView>
+     <BiometryWebView>
        <ReloadInterceptorWebView>
          <SupervisedWebView>
            <WebView>
          </SupervisedWebView>
        </ReloadInterceptorWebView>
+     </BiometryWebView>
    </CozyWebView>
  </CozyProxyWebView>
</HomeView>
```

I'm wondering if we can create a composable WebView to ease this process. Something that can be used like:
```jsx
<>
  <CozyWebView
    compose={[
      CryptoProxyWebView,
      NavigatorShareProxyWebView,
      CozyProxyWebView,
      BiometryWebView,
      ReloadInterceptorWebView,
      DownloadInterceptorWebView,
      SupervisedWebView,
      WebView,
    ]},
    someProps={someProps}
  />
</>
```

The idea would be to have a Container WebView able to take all component from the `compose` prop and to recreate a component tree from them.
The array order may be important. In the example `CryptoProxyWebView` would be parent of `NavigatorShareProxyWebView` etc.
Each component would take the props from its parent, manipulate them as needed then pass them to the next component (rendered as child)

The challenge would be:
[...]

PR #845, opened 2023-06-22
Temp: Proposal for ComposedWebView

Ldoppea:
This PR is a proposal for a refactoring on our WebView management.

What I'm waiting from you are some feedback on the proposal, if you think is a good idea or not. I'm currently stuck on the implementation and I don't want to invest more time until I'm sure it is the good direction.

### Problem

Today our codebase structure is the following:
```
HomeView
└ CozyProxyWebView
   └ CozyWebView
      └ ReloadInterceptorWebView
         └ SupervisedWebView
```

Each WebView component has a hard-coded reference to its WebView child (i.e `CozyWebView` defines that its child is a `ReloadInterceptorWebView`)

This makes difficult to add a layer to the WebView's tree, or to invert some of them in the hierarchy.

Imagine that we want to add a `ReloadableWebView` that embed ALL of our webviews and allows to unmount/remount everything when we want a reload (so we enforce the proxywebview to generate a new HTML)

To make this we would have to edit `HomeView` to use a `ReloadableWebView` instead of `CozyProxyWebView`, to hard code `ReloadableWebView` with `CozyProxyWebView` as child, and then we should remember to search for all other `CozyProxyWebView` references in the code and replace `CozyProxyWebView` by `ReloadableWebView` like we did in `HomeView`

Here would be the new codebase structure
```
HomeView
└ ReloadableWebView
  └ CozyProxyWebView
     └ CozyWebView
        └ ReloadInterceptorWebView
           └ SupervisedWebView
```

So my question is `How can we make this more flexible?`

### Proposal

My proposal is to create a `ComposedWebView` that would allow developers to "compose" their WebView like they want.

The API would be like this:
```js
return (
  <ComposedWebView
      {...props}
      childrenTypes={[
        CozyProxyWebView,
        CozyWebView,
        ReloadInterceptorWebView,
        SupervisedWebView
      ]}
    />
)
```

Which would be translated on runtime as the following
```
CozyProxyWebView
└ CozyWebView
   └ ReloadInterceptorWebView
      └ SupervisedWebView
```

This would allow to easily add/remove parts. If we want to remove the `ReloadInterceptorWebView` from the generated tree then we can do:
```js
return (
  <ComposedWebView
      {...props}
      childrenTypes={[
        CozyProxyWebView,
        CozyWebView,
        //ReloadInterceptorWebView,
        SupervisedWebView
      ]}
    />
)
```

If we want to add a `ReloadableWebView` on the top level then we can do:
```js
return (
  <ComposedWebView
      {...props}
      childrenTypes={[
        ReloadableWebView,
        CozyProxyWebView,
        CozyWebView,
        ReloadInterceptorWebView,
        SupervisedWebView
      ]}
    />
)
```

### Work in progress

This PR is a draft implementation for this concept.

`src/components/webviews/ComposedWebView.js` contains the logic that translates the flat declaration into a nested WebView tree on runtime.

This would have some repercussions on the way we create WebViews:
- WebView components must be wrapped in a `forwardRef`
- WebView components must take a `ChildWebview` prop defaulted to `WebView` and use it as their child
- All WebView methods like `onShouldStartLoadWithRequest` should consider they can have a parent and call it if exists
  - Example:
```js
<ChildWebView
  onShouldStartLoadWithRequest={(initialRequest) => {
    if (/* some condition */) {
      // do stuff
      return false
    }

    if (props.onShouldStartLoadWithRequest) {
      return props.onShouldStartLoadWithRequest(initialRequest)
    }  else {
      return true
    }
  }}
/>
```

⚠️ Current implementation is not complete and has some issues:
- ReloadInterceptorWebView gets nodeHandle errors
  - but if we remove it from the tree then everything seems to work
- CozyAppScreen will do an infinite loop
  - however HomeView seems to work correctly

### DIFF
--- a/src/components/webviews/ComposedWebView.js
+++ b/src/components/webviews/ComposedWebView.js
@@ -1,0 +1,78 @@
+import Minilog from '@cozy/minilog'
+import React, { useEffect, useRef } from 'react'
+
+import { CozyProxyWebView } from '/components/webviews/CozyProxyWebView'
+import { CozyWebView } from '/components/webviews/CozyWebView'
+import ReloadInterceptorWebView from '/components/webviews/ReloadInterceptorWebView'
+import { SupervisedWebView } from '/components/webviews/SupervisedWebView'
+
+const log = Minilog('🌍 ComposedWebView')
+
+const ComposedWebViewItem = React.forwardRef(
+  ({ ItemType, childrenTypes, currentLevel, ...otherProps }, ref) => {
+    const [firstChild, ...otherChildren] = childrenTypes
+
+    console.log('1️⃣ currentLevel', currentLevel)
+    const ChildWebview = React.forwardRef((subOtherProps, subRef) => {
+
+      useEffect(() => {
+        log.debug('ComposedWebView mount')
+
+        return () => {
+          log.debug('ComposedWebView unmount')
+        }
+      }, [])
+
+      return (
+        <ComposedWebViewItem
+          ItemType={firstChild}
+          childrenTypes={otherChildren}
+          currentLevel={currentLevel + 1}
+          ref={subRef}
+          {...subOtherProps}
+        />
+      )
+    })
+    ChildWebview.displayName = 'ChildWebviewRef'
+
+    return (
+      <>
+        {childrenTypes?.length > 0 ? (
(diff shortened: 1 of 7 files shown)


Crash--:
Maybe linked to https://github.com/cozy/cozy-flagship-app/issues/411 ?

PR #847, opened 2023-06-23
Fix icon feat splashscreen issues

zatteo:
**fix: Flags were not fetched at login with oidc and magic link**

Solve icon not changing at first login with oidc and magic link.

 **fix: Splash screen does not hide when icon has changed on iOS**

Changing icon and using splash screen do not work well together. If
we call hideSplashScreen when iOS native popup "Icon has changed" is
displayed, hideSplashScreen hides iOS native popup instead of splash
screen. So here we avoid to call changeIcon when splash screen is
displayed.

### DIFF
--- a/src/libs/icon/icon.spec.ts
+++ b/src/libs/icon/icon.spec.ts
@@ -1,3 +1,4 @@
+import { getVisibilityStatus } from 'react-native-bootsplash'
 import { changeIcon as RNChangeIcon } from 'react-native-change-icon'
 
 import flag from 'cozy-flags'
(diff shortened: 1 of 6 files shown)


src/libs/icon/icon.spec.ts
```diff
@@ -59,6 +70,7 @@ describe('icon', () => {
   })
 
   it('should throw if other error', async () => {
+    mockFlags(ICON_CHANGE_ALLOWED_FLAGS)
```
zatteo, line 73: By resetting every mock each time, I found that some tests very buggy.

PR #851, opened 2023-06-28
feat: save oauthName on app bootstrap if needed

acezard:
## Description

This PR introduces a new function `checkClientName` and its corresponding Jest tests. The function checks if a client's name starts with a specific string, defined as `translation.software.name`, and updates the client name on the server if it doesn't. In case of any error during the update, the error message is logged. 

This PR also includes TypeScript interfaces to represent the client response object, and Jest tests to ensure that `checkClientName` works as expected.

## Changes

- Created a new function `checkClientName` that checks a client's name and updates it on the server if necessary.
- Added TypeScript interfaces for the client response object.
- Wrote Jest tests for the `checkClientName` function, including tests for when the client name should be updated, should not be updated, and when an error occurs during update.
- Modified the logger to use `spyOn` for `info` and `error` methods, allowing for better isolation and control in testing.

## Testing

Run `yarn test` to execute the new Jest tests for the `checkClientName` function. All tests should pass, demonstrating that the function works as expected in different scenarios.

## Impact

This new function will help to ensure that all client names adhere to our desired naming convention. The added tests will help prevent regressions in this function in the future.

### DIFF
--- a/src/app/domain/authentication/services/SynchronizeService.ts
+++ b/src/app/domain/authentication/services/SynchronizeService.ts
@@ -1,14 +1,55 @@
 import Minilog from '@cozy/minilog'
+import { getDeviceName } from 'react-native-device-info'
 
 import type CozyClient from 'cozy-client'
+import { getErrorMessage } from 'cozy-intent'
 
-const log = Minilog('SynchronizeService')
+import { SOFTWARE_NAME } from '/libs/constants'
+import { saveClient } from '/libs/clientHelpers/persistClient'
+
+export const syncLog = Minilog('📱 SynchronizeService')
 
 export const synchronizeDevice = async (client: CozyClient): Promise<void> => {
   try {
     await client.getStackClient().fetchJSON('POST', '/settings/synchronized')
-    log.info('📱 Device synchronized')
+    syncLog.info('Device synchronized')
   } catch (error) {
-    log.warn('Error while synchronizing device', error)
+    syncLog.warn('Error while synchronizing device', error)
   }
+}
+
+export const getClientName = async (): Promise<string> =>
+  `${SOFTWARE_NAME} (${await getDeviceName()})`
+
+export const checkClientName = async (client: CozyClient): Promise<void> => {
+  try {
+    const { clientName, ...oauthOptions } = client.getStackClient().oauthOptions
+    const newClientName = await getClientName()
+
+    if (!clientName.startsWith(newClientName)) {
+      syncLog.info(`Updating client name...`, {
+        oldClientName: clientName
+      })
+
+      const response = await client.getStackClient().updateInformation({
+        ...oauthOptions,
+        clientName: newClientName
+      })
+
+      await saveClient(client)
+
+      syncLog.info(`Client name updated with "${newClientName}"`, {
+        OAuthOptions: response
+      })
+    } else {
+      syncLog.info(`Client name is up to date: "${clientName}"`)
+    }
+  } catch (error) {
+    syncLog.error('Failed to update clientName', getErrorMessage(error))
+  }
+}
+
+export const synchronizeOnInit = async (client: CozyClient): Promise<void> => {
+  await checkClientName(client)
+  await synchronizeDevice(client)
 }
--- a/src/hooks/useGlobalAppState.ts
+++ b/src/hooks/useGlobalAppState.ts
@@ -153,8 +153,6 @@ export const useGlobalAppState = (): void => {
         'useGlobalAppState: subscribing to AppState changes, synchronizing device'
       )
 
-      safePromise(synchronizeDevice)(client)
-
       subscription = AppState.addEventListener('change', e =>
         onStateChange(e, client)
       )
--- a/src/@types/cozy-client.d.ts
+++ b/src/@types/cozy-client.d.ts
@@ -23,6 +23,7 @@ declare module 'cozy-client' {
 
   interface OAuthOptions {
     clientID: string
+    clientName: string
     clientSecret: string
   }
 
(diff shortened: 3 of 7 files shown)


src/app/domain/authentication/services/SynchronizeService.ts
```diff
@@ -1,14 +1,35 @@
...
+}
+
+export const checkClientName = async (client: CozyClient): Promise<void> => {
+  const clientName = client.getStackClient().oauthOptions.clientName
+
+  if (!clientName.startsWith(translation.software.name)) {
+    try {
+      const { data } = (await client.save({
+        _type: 'io.cozy.oauth.client',
+        name: translation.software.name
```
acezard, line 27: mistake here, the devicename should be added back

Ldoppea: To prevent this we should extract a method that generates the client's name and use this method everywhere we set the client's name

src/app/domain/authentication/services/SynchronizeService.ts
```diff
@@ -1,14 +1,45 @@
...
+  }
+}
+
+export const getClientName = async (): Promise<string> =>
+  `${SOFTWARE_NAME} (${await getDeviceName()})`
+
+export const checkClientName = async (client: CozyClient): Promise<void> => {
+  const { clientName, ...oauthOptions } = client.getStackClient().oauthOptions
+
+  if (!clientName.startsWith(SOFTWARE_NAME)) {
```
Ldoppea, line 26: Here we can compare `clientName` to the entire `getClientName`, so it will also update if the user changes their device's name

src/hooks/useCheckClientName.ts
```diff
@@ -0,0 +1,10 @@
+import type CozyClient from 'cozy-client'
+
+import { useEffect } from 'react'
+
+import { checkClientName } from '/app/domain/authentication/services/SynchronizeService'
+import { safePromise } from '/utils/safePromise'
+
+export const useCheckClientName = (client: CozyClient): void => {
+  useEffect(() => safePromise(checkClientName)(client), [client])
```
Ldoppea, line 9: Should we use `useClient` here instead of passing it in parameters?

Ldoppea (approved):

src/hooks/useCheckClientName.ts
acezard: yes I think you're right it would look more streamlined

- [ ]

src/app/domain/authentication/services/SynchronizeService.ts
acezard: indeed, didn't think of that. 

- [ ] updated

src/app/domain/authentication/services/SynchronizeService.ts
```diff
@@ -1,14 +1,45 @@
...
+    try {
+      const newClientName = await getClientName()
+      const response = await client.getStackClient().updateInformation({
+        ...oauthOptions,
+        clientName: newClientName
+      })
+
+      syncLog.info(`Client name updated with "${newClientName}"`, {
+        OAuthOptions: response
+      })
```
acezard, line 40: add save local on success

- [ ] updated

src/app/domain/authentication/models/Device.ts
```diff
@@ -22,3 +22,27 @@ export const getBiometricType = async (): Promise<IsSensorAvailableResult> => {
 export const isAutoLockEnabled = async (): Promise<boolean> => {
   return Boolean(await getData(StorageKeys.AutoLockEnabled))
 }
+
+export interface UpdatedByApp {
```
acezard, line 26: n'est plus nécessaire

src/app/domain/authentication/services/SynchronizeService.ts
```diff
@@ -1,14 +1,51 @@
...
+    }
   } catch (error) {
-    log.warn('Error while synchronizing device', error)
+    syncLog.error('Failed to update clientName', getErrorMessage(error))
+  } finally {
+    safePromise(synchronizeDevice)(client)
```
Ldoppea, line 49: nit: I would pref that we call synchronizeDevice on app.js then getClientName is called by synchronizeDevice.

As a developper I don't expect that calling `checkClientName` would produce a synchronization. The reverse seems more logic.

src/hooks/useGlobalAppState.ts
```diff
@@ -153,8 +153,6 @@ export const useGlobalAppState = (): void => {
         'useGlobalAppState: subscribing to AppState changes, synchronizing device'
       )
 
-      safePromise(synchronizeDevice)(client)
```
Ldoppea, line 156: Care, by remove this from here, the synchronizeDevice is not called on every app focus anymore (which was the requirement as I remember it)

acezard: it is, a bit earlier in the file, this one was the first one. For some reason if it's called before the name update, there will be an error from the stack

src/app/domain/authentication/services/SynchronizeService.ts
acezard: I agree. Do you think I should investigate with back-end team why running sync before update triggers an error in the stack? But that means we can't deploy this PR today

Ldoppea: Isn't it because both call are async and fire&forget so the stack will receive both call at the same time?

src/hooks/useGlobalAppState.ts
Ldoppea: woups, didn't see the other one. So I guess this one was for the first render. So it's ok for me

src/app/domain/authentication/services/SynchronizeService.ts
acezard: https://github.com/cozy/cozy-stack/pull/4029 update on stack and I also refactored a bit so we always check if the name is correct before syncing when app starts

Ldoppea (approved):

PR #852, opened 2023-06-28
fix: Clean IconsTable on logout

Crash--:
We can have different icons for the same slug depending on the context of your cozy.

So instead of serving the wrong one, let's clean this data on logout.

fix: Display the right Mes Papiers icons when switching between cozies.

### DIFF
--- a/src/libs/localStore/storage.ts
+++ b/src/libs/localStore/storage.ts
@@ -2,24 +2,26 @@ import AsyncStorage from '@react-native-async-storage/async-storage'
 import { BiometryType } from 'react-native-biometrics'
 
 import { logger } from '/libs/functions/logger'
-
 const log = logger('storage.ts')
 
 const { setItem, getItem, removeItem } = AsyncStorage
-
 export enum StorageKeys {
   AutoLockEnabled = '@cozy_AmiralApp_autoLockEnabled',
   BiometryActivated = '@cozy_AmiralApp_biometryActivated',
   Capabilities = '@cozy_AmiralApp_Capabilities',
   LastActivity = '@cozy_AmiralApp_lastActivity',
   DefaultRedirectionUrl = '@cozy_AmiralAppDefaultRedirectionUrl',
-  SessionCreatedFlag = 'SESSION_CREATED_FLAG'
+  SessionCreatedFlag = 'SESSION_CREATED_FLAG',
+  IconsTable = '@cozy_AmiralAppIcons'
 }
 
-interface StorageItems {
+export type IconsCache = Record<string, { version: string; xml: string }>
+
+export interface StorageItems {
   biometryActivated: boolean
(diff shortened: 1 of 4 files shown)


Crash--:
@zatteo How can I save to the local storage but prefixed by the cozy's url. Is it ready to use?

zatteo:
I added this lib here https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/localStore/userPersistedStorage.ts

Exemple here https://github.com/cozy/cozy-flagship-app/blob/3f80ad221b44fdb6bafb219de1851fa8c4ef2b12/src/app/domain/backup/services/manageLocalBackupConfig.ts#L36C37-L36C37

It allows to store data by user fqdn **persisted** between login/logout.

Crash--:
Thanks for the info. Next time, I'll use it :)

Crash--:
@Ldoppea @acezard ready to review ;)

Ldoppea:
Ok now I understand why you wanted to re-use the string from `strings.json` into the `StorageKeys` enum

As you added the `IconsTable` key to the `StorageKeys` enum, then maybe we can homogenise the `iconTable.ts` api?

I guess it would consist of replacing all
```js
await AsyncStorage.getItem(strings.APPS_ICONS)
//...
await AsyncStorage.setItem(strings.APPS_ICONS, JSON.stringify(table))
```

by
```js
import { getData, StorageKeys, storeData } from '/libs/localStore/storage'

await getData(StorageKeys.IconsTable)
//...
await storeData(StorageKeys.IconsTable, table) // IconsCache type should be added to StorageItems interface in storage.ts
```

And then remove the entry from `strings.json`

src/libs/localStore/storage.ts
```diff
@@ -6,14 +6,14 @@ import { logger } from '/libs/functions/logger'
...
-
 export enum StorageKeys {
   AutoLockEnabled = '@cozy_AmiralApp_autoLockEnabled',
   BiometryActivated = '@cozy_AmiralApp_biometryActivated',
   Capabilities = '@cozy_AmiralApp_Capabilities',
   LastActivity = '@cozy_AmiralApp_lastActivity',
   DefaultRedirectionUrl = '@cozy_AmiralAppDefaultRedirectionUrl',
-  SessionCreatedFlag = 'SESSION_CREATED_FLAG'
+  SessionCreatedFlag = 'SESSION_CREATED_FLAG',
+  IconsTable = '@cozy_AmiralAppIcons' // yeah, I can't use strings.APP_ICONS here.
```
Ldoppea, line 16: Is that comment really needed? 😅

Ldoppea (approved):

Crash--:
@Ldoppea @acezard Ready for review. I know that I should have split the commit :(

Ldoppea (approved):

PR #854, opened 2023-06-29
fix: Don't capitalize password input

Crash--:
We don't want to autocapitalize input password field


LucsT:
Tested on android. Working Well.

Fixing the know 'bug'

Ldoppea (approved):

PR #858, opened 2023-07-03
fix: inject applicationNameForUserAgent

acezard:
By examining the existing code,
it is now clear that we weren't injecting userAgent props
but applicationNameForUserAgent props.

This was certainly done with intent, as to let the possibility
to overwrite the default value with the userAgent prop,
since this prop will override applicationNameForUA.

All webviews have been updated to incorporate this change

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android


src/components/webviews/ReloadInterceptorWebView.js
```diff
@@ -32,8 +32,9 @@ const ReloadInterceptorWebView = React.forwardRef((props, ref) => {
       <ProgressContainer progress={progress}>
         <SupervisedWebView
           {...props}
+          {...(userAgent ? { userAgent } : {})}
```
Ldoppea, line 35: you can remove this line and the `userAgent` declaration on l25 as we don't do any transformation anymore, if provided by the parent it will be injected with the `{...props}` instruction.

Ldoppea: Also we never inject it from parents on our codebase, so there is no need to make it visible.

Only place userAgent is used is from LauncherView, but it uses the classic WebView.

Ldoppea (approved):

PR #860, opened 2023-07-04
fix: Call login, saveClient and listenTokenRefresh on magicLink and OIDC

Ldoppea:
Those methods were missing from the MagicLink and OIDC login processes

Without them:
- the `client.emit('login', ...)` would never be called and so CozyClient's plugins would not be initialized, which would produce infinite await on `client.plugins.flags.initializing` on LoginScreen
- the client would not be persisted
- the TokenRefresh event wouldn't be handled

This happens only on production environment and on Flagship certified devices. Otherwise, the manual certification workflow would correctly call those methods

### DIFF
--- a/src/libs/client.spec.js
+++ b/src/libs/client.spec.js
@@ -23,7 +23,8 @@ const mockCozyClient = {
     setUri: mockSetUri,
     register: mockRegister,
     updateInformation: mockUpdateInformation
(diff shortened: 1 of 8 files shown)


Crash-- (approved):

zatteo:
Nice new function!

PR #862, opened 2023-07-05
feat: handle revoked tokens

acezard:
This PR introduces a new feature to handle user token revocation in a more seamless and user-friendly manner. When a user's token is no longer valid (e.g., due to expiration or a revocation by the server), the user is redirected to the welcome page, and a modal dialog is displayed to inform them about the issue.

## Changes

1. **AuthService Update:** The authentication service was updated to include a new state, `userRevoked`, which is set to `true` when a token error is detected, indicating that the user has been revoked due to an invalid token.

2. **Event-Based Status Updates:** The AuthService now uses an event-based approach to communicate status updates to other parts of the app. This ensures a clear separation of concerns and improves code maintainability.

3. **User Revocation Modal:** A new modal dialog has been created to inform the user when they have been revoked due to an invalid token. This modal is displayed on the welcome screen and can be dismissed by the user.

4. **Custom Hook:** A new custom hook, `useRevokedStatus`, has been introduced to encapsulate the logic of handling the user revoked status. This makes the logic reusable and helps keep the component code clean and focused on rendering.

5. **WelcomeScreen Update:** The WelcomeScreen component has been updated to use the `useRevokedStatus` hook and show the user revocation modal when necessary.

## Testing

to be added

## Screenshots

image


src/app/domain/authentication/services/AuthService.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+// Start listening to cozy-client and AuthService event
+export const startListening = (client: CozyClient): void => {
+  authLogger.info('Start listening to cozy-client events')
+  clientInstance = client
+
+  clientInstance.on('revoked', () => {
+    authLogger.info('Token revoked')
```
Ldoppea, line 54: This log seems to be useless as you have the same one in `handleTokenError`

src/app/domain/authentication/services/AuthService.ts
```diff
@@ -0,0 +1,64 @@
...
+    emitter.emit('userLoggedOut', false)
+  }
+}
+
+// Start listening to cozy-client and AuthService event
+export const startListening = (client: CozyClient): void => {
+  authLogger.info('Start listening to cozy-client events')
+  clientInstance = client
+
+  clientInstance.on('revoked', () => {
```
Ldoppea, line 53: You pass no params, so I guess you can just
```suggestion
  clientInstance.on('revoked', handleTokenError)
```

src/libs/client.js
```diff
@@ -56,6 +57,9 @@ export const getClient = async () => {
   })
   listenTokenRefresh(client)
   client.getStackClient().setOAuthOptions(oauthOptions)
+
+  startListening(client)
```
Ldoppea, line 61: Here you are in `getClient` which is the method called when retrieving the client from asyncStorage on app re-start

On login this method is not called, instead we use `createClient` and `finalizeClientCreation`. So you should add this to one of them

src/app/domain/authentication/services/AuthService.ts
Ldoppea, line 53: We never call `removeListener` on this? However I'm not sure it is necessary as we recreate a client on next login, but someday in the future we expect to re-use the same client and only work with login/logout methods. So maybe we should handle this `removeListener` now?

src/app/domain/authentication/services/AuthService.ts
```diff
@@ -0,0 +1,64 @@
...
+  authLogger.warn('Token revoked')
+  setUserRevoked(true)
+}
+
+// Handle user confirmation of the error
+const handleOnUnrevoked = async (): Promise<void> => {
+  try {
+    if (clientInstance === null) throw new Error('No client instance set')
+    authLogger.info('User confirmed the error, logging out')
+    await asyncLogout(clientInstance)
```
Ldoppea, line 39: `asyncLogout` will call 
```
await sendKonnectorsLogs(client)
await client.logout()
```

Both those method will do HTTP requests to the stack and will fail if client is revoked.

The process may correctly catch those exceptions, but we will still get the `log.error` calls into sentry. Maybe we should just clean the asyncstorage ?

src/app/view/Auth/ErrorTokenModal.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+describe('ErrorTokenModal and useRevokedStatus', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('renders ErrorTokenModal when revoked status is true', () => {
```
Ldoppea, line 55: Those tests are not testing `ErrorTokenModal` but `ComponentUsingHook`. As the conditional rendering is in this mocked component.

src/app/domain/authentication/services/AuthService.ts
```diff
@@ -0,0 +1,64 @@
...
+import { safePromise } from '/utils/safePromise'
+
+export const authLogger = Minilog('[AuthService]')
+const emitter = new EventEmitter()
+
+let userRevoked = false
+let clientInstance: CozyClient | null = null
+
+// Setter
+export const setUserRevoked = (status: boolean): void => {
```
Ldoppea, line 17: Same here, `setUserRevoked` goal is to react to local state, but it seems like we can just react to `handleTokenError` and `userConfirmation` calls. My comprehension is that this service would be easier to understand if stateless without any feature loss. I'm I right?

Ldoppea (changes requested):
Sorry I made a lot of comments but I'm a bit worried by the over-complexity of this PR.

The only business logic seems to be `handleOnUnrevoked` that effectively clean the local state. The rest seems to be unnecessary layers between `client.on('revoked')` and the dialog's validation button. So there seems overkill to have 2 states (one in `useRevokedStatus` and one in `AuthService`) where only one seems enough and 2 new events with a new listener where the `client.on('revoked')` seems enough.

But maybe I'm missing something so feel free to explain why you did this.

acezard:
> Sorry I made a lot of comments but I'm a bit worried by the over-complexity of this PR.
> 
> The only business logic seems to be `handleOnUnrevoked` that effectively clean the local state. The rest seems to be unnecessary layers between `client.on('revoked')` and the dialog's validation button. So there seems overkill to have 2 states (one in `useRevokedStatus` and one in `AuthService`) where only one seems enough and 2 new events with a new listener where the `client.on('revoked')` seems enough.
> 
> But maybe I'm missing something so feel free to explain why you did this.

Well, my goal was to make this feature as explicit and compartmentalized as possible, separating concerns for clean and testable code. But it's possible that I disregarded too many practical considerations while doing so. 

I guess it could be simplified a lot by eliminating both the mirrored state and using a more declarative approach in the handling of errors and hiding dialogs. But since I didn't want to use a Context, it's somehow complex to send information to UI from a service. Maybe I could use react-navigation params when redirecting, which will show the modal, and remove these params when the user confirms.

src/app/domain/authentication/services/AuthService.ts
acezard: - [ ] Updated

src/app/domain/authentication/services/AuthService.ts
acezard: - [ ] Updated

src/libs/client.js
acezard: - [ ] Updated

src/app/domain/authentication/services/AuthService.ts
acezard: That's correct, I shall look into the API to unsubscribe
- [ ] Updated

src/app/domain/authentication/services/AuthService.ts
```diff
@@ -0,0 +1,64 @@
...
+export const isUserRevoked = (): boolean => userRevoked
+
+// Handle token error
+const handleTokenError = (): void => {
+  authLogger.warn('Token revoked')
+  setUserRevoked(true)
+}
+
+// Handle user confirmation of the error
+const handleOnUnrevoked = async (): Promise<void> => {
```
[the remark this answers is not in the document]
acezard: Yes, it's more `handleUserConfirmation` 
- [ ] Updated

src/app/domain/authentication/services/AuthService.ts
acezard: Agree, I think we should call everything that doesn't rely on a client, will look into it
- [ ] Updated

src/app/domain/authentication/services/AuthService.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import type CozyClient from 'cozy-client'
+
+import { asyncLogout } from '/libs/intents/localMethods'
+import { safePromise } from '/utils/safePromise'
+
+export const authLogger = Minilog('[AuthService]')
+const emitter = new EventEmitter()
+
+let userRevoked = false
```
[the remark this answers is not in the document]
acezard: I think it could be handled in a functional stateless way yes. That way I can remove both this state and the hook state

src/app/domain/authentication/services/AuthService.ts
acezard: Yeah, I guess it can expose only one or two of these methods depending on how the modal display/hide is handled. Overall I think there is room to simplify the feature and remove almost all stateful logic

src/locales/fr-FR/translation.json
```diff
@@ -70,6 +70,11 @@
   "modals": {
     "IconChangedModal": {
       "description": "L'icône de l'application «Cozy» a été modifiée."
+    },
+    "ErrorToken": {
+      "title": "Accès révoqué",
+      "body": "Vous semblez avoir révoqué cet appareil depuis votre Cozy. Pour accéder à nouveau à vos données, veuillez vous reconnecter.Si cette action ne vient pas de vous, n’hésitez pas à nous contacter à support@cozycloud.cc.",
```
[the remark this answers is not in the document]
Ldoppea: Should `support@cozycloud.cc` be clickable in the UI?

src/app/domain/authentication/services/AuthService.ts
```diff
@@ -0,0 +1,49 @@
...
+import { deleteKeychain } from '/libs/keychain'
+import { clearCookies } from '/libs/httpserver/httpCookieManager'
+import { clearData } from '/libs/localStore/storage'
+import { routes } from '/constants/routes'
+import { navigationRef } from '/libs/RootNavigation'
+
+export const authLogger = Minilog('[AuthService]')
+let clientInstance: CozyClient | null = null
+
+const asyncLogoutNoClient = async (): Promise<void> => {
```
Ldoppea, line 16: nit: it would be great to move this code near `asyncLogout` so we can easily see that both version exist and we don't forget to edit both if we add an item to clear on logout.

Or event better: find a way to mutualize them.

src/app/view/Auth/ErrorTokenModal.spec.tsx
acezard: This test is removed as I don't see any benefit in testing a simple modal now that the hook is removed

src/screens/welcome/WelcomeScreen.jsx
```diff
@@ -71,6 +72,13 @@ export const WelcomeScreen = ({ navigation, route, setClient }) => {
         setClient={setClient}
         goBack={handleBackPress}
       />
+
+      {route.params?.options === 'showTokenError' && (
+        <ErrorTokenModal
+          onClose={() => navigation.setParams({ options: '' })}
```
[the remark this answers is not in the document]
acezard: yes unfortunately it seems to trigger a rerender from what I know. Maybe it doesn't always do that? But I remember fixing a huge bug linked to that

src/app/domain/authentication/services/AuthService.ts
acezard: yeah but asyncLogout isn't in the greatest place right now. I have to check if it's easy to move it now

PR #866, opened 2023-07-10
refactor: replace @cozy/minilog with cozy-minilog

acezard:
This will use the latest version of the library.
Fixing some formatting bugs when running inside Hermes

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "web:test": "react-app-rewired test --env=jsdom"
   },
   "dependencies": {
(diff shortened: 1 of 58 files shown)


Crash--:
> Cleaner than aliasing modules

Sure! But takes a lot of time. IMHO, you should also alias

acezard:
> > Cleaner than aliasing modules
> 
> Sure! But takes a lot of time. IMHO, you should also alias

apparently only cozy-client is concerned for us

Ldoppea (approved):

PR #873, opened 2023-07-13
Backup : create folder structure for Android, fix various bugs

zatteo:
- Create same folder structure for Android as on device
- Stopping backup return ready state so that we can see number of medias remaining to backup
- Fix various bugs

### DIFF
--- a/src/app/domain/backup/queries/index.ts
+++ b/src/app/domain/backup/queries/index.ts
@@ -1,6 +1,7 @@
 import { Q, QueryDefinition } from 'cozy-client'
 
 const DOCTYPE_FILES = 'io.cozy.files'
+const DOCTYPE_ALBUMS = 'io.cozy.photos.albums'
 
 export const buildFilesQuery = (deviceId: string): QueryDefinition => {
   return Q(DOCTYPE_FILES)
@@ -25,3 +26,21 @@ export interface File {
 }
 
 export type FilesQueryAllResult = File[]
+
+export const buildAlbumsQuery = (deviceId: string): QueryDefinition => {
+  return Q(DOCTYPE_ALBUMS)
+    .where({
+      backupDeviceIds: {
+        $elemMatch: { $eq: deviceId }
+      }
+    })
+    .indexFields(['backupDeviceIds'])
+    .select(['backupDeviceIds', 'name'])
+}
+
+export interface AlbumDocument {
+  id: string
+  name: string
+}
+
+export type AlbumsQueryAllResult = AlbumDocument[]
--- a/src/app/domain/backup/queries/index.ts
+++ b/src/app/domain/backup/queries/index.ts
@@ -2,23 +2,26 @@ import { Q, QueryDefinition } from 'cozy-client'
 
 const DOCTYPE_FILES = 'io.cozy.files'
 
-export const buildFilesQuery = (dirId: string): QueryDefinition => {
+export const buildFilesQuery = (deviceId: string): QueryDefinition => {
   return Q(DOCTYPE_FILES)
     .where({
-      dir_id: dirId,
-      type: 'file'
+      type: 'file',
+      'metadata.backupDeviceIds': {
+        $elemMatch: { $eq: deviceId }
+      }
     })
-    .indexFields(['dir_id', 'type'])
-    .select(['name', 'dir_id', 'type'])
+    .indexFields(['metadata.backupDeviceIds', 'type'])
+    .select(['metadata.backupDeviceIds', 'type', 'name', 'path'])
 }
 
 export const buildFileQuery = (id: string): QueryDefinition => {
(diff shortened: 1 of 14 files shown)


src/app/domain/backup/queries/index.ts
```diff
@@ -2,23 +2,26 @@ import { Q, QueryDefinition } from 'cozy-client'
...
-      dir_id: dirId,
-      type: 'file'
+      type: 'file',
+      'metadata.backupDeviceIds': {
+        $elemMatch: { $eq: deviceId }
+      }
     })
-    .indexFields(['dir_id', 'type'])
-    .select(['name', 'dir_id', 'type'])
+    .indexFields(['type', 'metadata.backupDeviceIds'])
```
Crash--, line 14: You should invert the order. Always the most restrictive before. I think this is documented somewhere here https://docs.cozy.io/en/tutorials/data/advanced/#indexes-performances-and-design

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -51,7 +51,18 @@ export const uploadMedias = async (
     }
 
     try {
-      const uploadUrl = getUploadUrl(client, backupFolderId, mediaToUpload)
+      const uploadFolderId = await getUploadFolderId(
+        client,
+        localBackupConfig,
+        mediaToUpload
+      )
```
Crash--, line 60: An async call inside a loop, I don't like it. 

nit: Also this method should be rename to something like getOrCreateFolder

src/app/domain/backup/services/getMedias.ts
```diff
@@ -27,6 +27,36 @@ const getPhotoIdentifiersPage = async (
...
+        return `image/${extension}`
+    }
+  }
+
+  if (media.type === 'video') {
+    switch (extension) {
+      case 'mp4':
+        return 'video/mpeg'
+      default:
+        return `video/${extension}`
```
Crash--, line 53: Are we sure to want that? Having video/ogx will be weird (instead of video/ogg). Do we really need to pass the extension?

Crash--: How do Drive handle that?

src/app/domain/backup/queries/index.ts
```diff
@@ -25,3 +26,21 @@ export interface File {
...
+
+export const buildAlbumsQuery = (deviceId: string): QueryDefinition => {
+  return Q(DOCTYPE_ALBUMS)
+    .where({
+      backupDeviceIds: {
+        $elemMatch: { $eq: deviceId }
+      }
+    })
+    .indexFields(['backupDeviceIds'])
+    .select(['backupDeviceIds', 'name'])
```
Crash--, line 38: From: https://docs.cozy.io/en/tutorials/data/queries/#selectors 

> Thus, when an index is required, any field involved in a selector should be indexed.

Crash-- (approved):

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -146,7 +227,9 @@ const getUploadUrl = (
     '&Type=file&Tags=library&Executable=false&CreatedAt=' +
     createdAt +
     '&UpdatedAt=' +
-    createdAt
+    createdAt +
+    '&MetadataID=' +
+    metadataId
```
Ldoppea, line 232:
```ts
const toURL = new URL(client.getStackClient().uri)
toURL.pathname = '/files/' + backupFolderId
toURL.searchParams.append('Name', media.name)
// etc
```

src/app/domain/backup/services/uploadMedia.ios.ts
```diff
@@ -82,3 +94,7 @@ export const uploadMedia = async (
       })
   })
 }
+
+export const cancelUpload = (uploadId: string): Promise<void> => {
+  return RNFileSystem.stopUpload(parseInt(uploadId)) as unknown as Promise<void>
```
Ldoppea, line 99: `stopUpload` returns `void` cf https://github.com/itinance/react-native-fs/blob/64aa755cc1d37f59fa205bf2d52dd71a7d691504/RNFSManager.m#L676-L677

src/app/domain/backup/services/uploadMedias.ts
zatteo: I keep this review for a next PR about optimization.

src/app/domain/backup/services/getMedias.ts
zatteo: I add 'mime/lite' to have the best mime guessing. It is 2.5KB. It will be possible to remove it one day if cameraroll return mime type for iOS. 

Package is also used by Drive for some purposes.

Ldoppea (approved):

PR #875, opened 2023-07-17
feat: add i18next and transifex into the project

acezard:
### Implementations

- Added i18next, intl-pluralrules (polyfill), react-i18next, react-native-localize (native helpers)
- Implemented i18n setup file
- Created auto generated locales files for en and es (uploaded on tx)
- Implemented TX pull & locale files validation
- Export functions to use i18n in non-react files
- Implemented i18n everywhere in the app and removed old system
- Set default language to "en"
- Handles language at app creation, app start, and cozy-settings lang change

image


src/libs/clientHelpers/createClient.ts
```diff
@@ -41,6 +43,7 @@ export const createClient = async (instance: string): Promise<CozyClient> => {
   const stackClient = client.getStackClient()
   stackClient.setUri(instance)
   await stackClient.register(instance)
+  await changeLanguage(client.getInstanceOptions().locale)
```
acezard, line 46: changeLanguage() called here

src/libs/client.js
```diff
@@ -63,6 +64,7 @@ export const getClient = async () => {
 
   await client.registerPlugin(flag.plugin)
   await client.plugins.flags.initializing
+  await changeLanguage(client.getInstanceOptions().locale)
```
acezard, line 67: changeLanguage() called here

src/components/webviews/CozyWebView.js
```diff
@@ -179,9 +180,11 @@ export const CozyWebView = ({
         tryCrypto(m, log, logId, onAnswer)
         tryNavigatorShare(m, log, logId, onAnswer)
         tryConsole(m, log, logId)
-
         nativeIntent.tryEmit(m)
 
+        if (m.nativeEvent.data.includes('setLang'))
+          setLang(...JSON.parse(m.nativeEvent.data).args)
```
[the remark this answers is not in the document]
acezard: this is not needed anymore it will go through nativeIntent

package.json
```diff
@@ -10,15 +10,18 @@
     "home:build": "(cd cozy-home && yarn build:mobile)",
     "home:watch": "(cd cozy-home && yarn watch:mobile)",
     "install:home:local": "bash ./scripts/install-home.sh",
-    "install:home:registry": "yarn tsc --project scripts && node scripts/dist/install-home-registry.cmd.js",
+    "install:home:registry": "node scripts/dist/install-home-registry.cmd.js",
+    "install:scripts": "yarn tsc --project scripts && yarn install:home:registry",
```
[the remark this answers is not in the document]
acezard: yes but now it also compiles the validation script that is in ts, so I wanted to compile them only once and not twice. By doing it in the postinstall this is ensured

src/app/view/Lock/LockScreen.tsx
```diff
@@ -46,123 +46,126 @@ const LockView = ({
...
+          <Button onPress={tryUnlock}>
+            <Typography color="primary" variant="button">
+              {t('ui.buttons.unlock')}
+            </Typography>
+          </Button>
+        </Grid>
+      </Grid>
+    </Container>
+  )
+}
```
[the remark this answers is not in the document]
acezard: - [ ] Updated

src/libs/clientHelpers/magicLink.ts
```diff
@@ -60,7 +60,8 @@ const loginMagicLink = async (
           twoFactorPasswordNeeded: true
         }
       } else if (e.reason?.error === ERROR_INVALID_MAGIC_CODE) {
-        throw new Error(translation.screens.login.invalidMagicCode, e)
+        // Why is this translated but not the other error message?
```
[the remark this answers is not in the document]
acezard: - [ ] Updated

src/locales/i18n.ts
```diff
@@ -0,0 +1,63 @@
...
+export type UseTranslationResult = ReturnType<typeof useTranslation>
+export type ChangeLanguageFunction = (
+  ...args: Parameters<I18nInstance['changeLanguage']>
+) => ReturnType<I18nInstance['changeLanguage']>
+
+export const i18n = (): I18nInstance => _i18n
+export const t: TranslationFunction = (...args) => _i18n.t(...args)
+export const useI18n = (): UseTranslationResult => useTranslation()
+export const changeLanguage: ChangeLanguageFunction = languageCode =>
+  _i18n.changeLanguage(languageCode)
```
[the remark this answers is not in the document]
acezard: yes, what should we do if it's not supported? switch to english or stay on es/fr?

- [ ] Updated

src/libs/clientHelpers/magicLink.ts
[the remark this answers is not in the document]
acezard: I think it's not translated because the user doesn't get to see it unlike the other message from what I understand

src/libs/clientHelpers/createClient.ts
acezard: > Is `client.getInstanceOptions()` supposed to work when not logged yet?

it doesn't work indeed. I'm not sure where to put it, there are a lot of options. Can't we use a single call for both login and creation?

acezard: getClient, finalizeClient, check if lang can be used before login, add a fallback in nav to check if lang is correct

src/locales/i18n.ts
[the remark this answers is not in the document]
Ldoppea: If the language is available then take this language
Else try to take the phone's language
Else if phone's language not available, then take english

src/libs/clientHelpers/createClient.ts
acezard: setLang to phone lang on logout

acezard: set phone lang on welcome screen

src/locales/i18n.ts
```diff
@@ -0,0 +1,101 @@
...
+  // Otherwise, fallback to the device's language (if supported)
+  const deviceLanguage = getDeviceLanguage()
+
+  if (supportedLanguages.includes(deviceLanguage)) {
+    await _i18n.changeLanguage(deviceLanguage)
+    return
+  }
+
+  // If the device's language is also not supported, default to English
+  await _i18n.changeLanguage(defaultLanguage)
```
Ldoppea, line 89: Isn't this equivalent to `await _i18n.changeLanguage(language)`? My understanding is that `language` is already the device's language if supported, otherwise the default language

src/screens/login/LoginScreen.js
```diff
@@ -147,7 +149,10 @@ const LoginSteps = ({
       try {
         const client = await createClient(instance)
 
-        const { kdfIterations, name } = await fetchPublicData(client)
+        const { kdfIterations, name, locale } = await fetchPublicData(client)
+
+        // If we are able to get a locale from the public data, we update the language asap
+        if (locale) void changeLanguage(locale)
```
Ldoppea, line 155: `setInstanceData` is called only for classic login, in case of magicLink based auth then `startMagicLinkOAuth` is called and in case of OIDC oauth then `startOidcOAuth` is called. During those scenario we can have the 2FA screen displayed that should be translated too.

Also I'm wondering if we have enough data to do the same on the OnboardingScreen?

src/locales/i18n.ts
acezard: yeah that's true but I guess it's more explicit that way?

src/screens/login/LoginScreen.js
acezard: I suggest doing this in a following PR as getting locale info before full login wasn't really in the scope of this feature at the start

Ldoppea (approved):

src/screens/login/LoginScreen.js
Ldoppea: I'm ok with that 👍

PR #876, opened 2023-07-17
feat: Handle Blocked and Not Onboarded Cozy on Login screen

Ldoppea:
### feat: Handle Blocked Cozy on Login screen

When trying to log in a "blocked" Cozy then the app should display a custom error page explaining to the user what is happening and what to do

This can happen on multiple client's calls (i.e. `client.register()`) so to be sure to intercept the error we chose to handle it in the ErrorView which would redirect to the correct screen if detected. This should handle classic login and OIDC, MagicLink ones

The classic login scenario depends on cozy/cozy-client#1352, otherwise the displayed screen will be CozyNotFound

The error interception is done in two places. First on ClouderyView as we can intercept the dedicated exception, this should be the most common scenario. Second place is on ErrorView as a Cozy can be locked when entering a wrong password, so we should be able to intercept this exception anywhere in the login flow

The new CozyBlockedPage displays a forbidden logo that is a copy of the one from cozy-stack (`assets/templates/instance_blocked.html`)

Related PR: cozy/cozy-client#1352
Related PR: cozy/cozy-client#1343

___

### feat: Handle Not Onboarded Cozy on Login screen

When trying to log in a "not onboarded" Cozy then the app should display a custom error page explaining to the user what is happening and what to do

The new CozyBlockedPage displays an email logo that is a copy of the one from cozy-stack (`assets/templates/need_onboarding.html`)

Related PR: cozy/cozy-stack#4052

___

TODO:

- [x] Update cozy-client after https://github.com/cozy/cozy-client/pull/1352 is merged


src/screens/error/ErrorScreen.tsx
```diff
@@ -75,13 +83,17 @@ const handleMessage = async (
   )
 
 const makeSource = (route: ErrorScreenProps['route']): Source => {
```
acezard, line 89: maybe it's time to decouple this function from the view file if we keep this pattern. It's starting to grow big

src/locales/fr-FR/translation.json
```diff
@@ -24,6 +24,10 @@
       "body": "Nous n'avons trouvé aucun Cozy à cette adresse. Il s'agit peut-être d'une erreur de frappe. Pour le savoir, vérifiez l'adresse de votre Cozy dans l'e-mail que nous vous avons envoyé au moment de sa création.",
       "title": "L'adresse de ce Cozy n'existe pas."
     },
+    "cozyNotOnboarded": {
```
acezard, line 31: as for the other PRs we'll have to be careful and handle these new strings in the i18n

src/screens/login/components/ErrorView.js
acezard, line 1: for me, this component is getting too complex and handles too much logic. I think it would be safer to decouple it, allowing better testing and scalability

Ldoppea: This has been remove in the followin PR: https://github.com/cozy/cozy-flagship-app/pull/877

src/locales/fr-FR/translation.json
Ldoppea: I applied latest translation changes

acezard (approved):

acezard (approved):

PR #882, opened 2023-07-24
Fix/android backup upload

zatteo:
**Small improvements for Android backup**

Avoid notification flood in Android when uploading

patch-package react-native-background-upload to return server response code when server error to allow custom behavior depending on server error

**Custom behavior when quota exceeded**

When quota exceeded error, return status code to front end (see https://github.com/cozy/cozy-drive/pull/2941/files)


patches/react-native-background-upload+6.6.0.patch
```diff
@@ -0,0 +1,36 @@
...
+diff --git a/node_modules/react-native-background-upload/index.d.ts b/node_modules/react-native-background-upload/index.d.ts
+index 8b2a07c..80cff93 100644
+--- a/node_modules/react-native-background-upload/index.d.ts
++++ b/node_modules/react-native-background-upload/index.d.ts
+@@ -11,6 +11,8 @@ declare module "react-native-background-upload" {
+ 
+     export interface ErrorData extends EventData {
+         error: string
++        responseCode: number
++        responseBody: string
```
Crash--, line 33: Did you create an upstream PR? can you link it?

src/app/domain/backup/helpers/error.ts
```diff
@@ -0,0 +1,18 @@
...
+    })
+
+    super(stringifiedMessage)
+    this.name = 'BackupError'
+    this.statusCode = statusCode
+  }
+}
+
+export const STACK_ERRORS_INTERRUPTING_BACKUP = [
+  413 // quota exceeded
```
Crash--, line 17: This is not true all the time. You can have a 413 error when the file is too big (5 GB) for our storage service (swift). 

If you want to handle that correctly, see desktop: 

https://github.com/cozy-labs/cozy-desktop/blob/eb1d431d382b9e7dcb93c49421e0555c60085951/core/remote/errors.js#L274-L287

Crash--: can be done later, not a blocker ;)

src/app/domain/backup/services/uploadMedia.ios.ts
```diff
@@ -4,13 +4,24 @@ import RNFileSystem from 'react-native-fs'
...
+export const getVideoPathFromLivePhoto = (photoPath: string): string => {
+  const extension = getPathExtension(photoPath)
+
+  if (extension === extension.toUpperCase()) {
+    // Path is something like IMG_001.MOV
+    return getPathWithoutExtension(photoPath) + '.MOV'
+  } else {
+    // When Live Photo has been modified, path is something like FullSizeRender.mov
+    return getPathWithoutExtension(photoPath) + '.mov'
+  }
```
Crash--, line 24: seriously? 😮‍💨

Crash--: Really? New version of camera-roll handle that? Why can't we upgrade?

src/app/domain/backup/helpers/error.ts
zatteo: Did not know about this! I will update with this new PR : https://github.com/cozy/cozy-stack/pull/4069/files

patches/react-native-background-upload+6.6.0.patch
zatteo: Added in the commit https://github.com/Vydia/react-native-background-upload/pull/337

src/app/domain/backup/services/uploadMedia.ios.ts
zatteo: No no I wanted to say "we will make a PR to add the feature directly on react-native-camera-roll"

PR #884, opened 2023-07-26
feat: Intercept window.open in worker

doubleface:
And convert it to a change of url in the worker.

I suppose this is the expected behavior for every konnector we have at the moment or else we may make it optional.

This is needed for the sncf konnector login but we still have to check if this does not affect other konnectors regarding the download of files and this is why this PR is still draft

We also will have to check if this works as expected on iOS

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android


Crash--:
Nice use of this API ;). Can you just add a comment if it's a bug fix? Like that, the changelog and the tests after are more easier to read / do.

Crash--:
@doubleface Is it still required?

doubleface:
Yes it is still required for the sncf konnector. I was waiting for Y. to review it and forgot it. @Crash-- Can you review it please?

Crash-- (approved):

PR #886, opened 2023-07-28
fix: Upgrade cozy-client

Crash--:
In order to not destroy indexes automatically


Ldoppea:
Related commit: https://github.com/cozy/cozy-client/commit/fd6d6a487f845718e8a709fdd9e4cf74ea47a04c

PR #896, opened 2023-08-10
feat: More logs in konnectors

doubleface:
- Log error result of a konnector execution, to see the error in logs even if the error occurs before authentication
- Send logs to stack even when no job is created (theses logs are useful and will be sent by the flagship app later anyway)

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -97,14 +97,19 @@ class ReactNativeLauncher extends Launcher {
    *
    * @param  {ContentScriptLogMessage} message - log message
    */
-  log(logContent) {
+  log({ timestamp, ...logContent }) {
     const context = this.getStartContext()
-    const slug = context.konnector.slug // konnector attribut is available before manifest one
+    const slug = context.konnector.slug // konnector attribute is available before manifest one
     let jobId
     if (context.job) {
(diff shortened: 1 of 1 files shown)


LucsT (approved):

Crash-- (approved):

src/libs/ReactNativeLauncher.js
```diff
@@ -57,7 +58,8 @@ class ReactNativeLauncher extends Launcher {
     this.controller = new AbortController()
 
     const wrapTimer = wrapTimerFactory({
-      logFn: msg => this.log({ level: 'info', msg })
+      logFn: msg =>
+        this.log({ level: 'info', msg, timestamp: new Date().toISOString() })
```
Ldoppea, line 62: nit pour l'avenir: est-ce qu'on ne devrait pas mettre ce timestamp en fallback value dans `this.log()` pour être sûr de tout le temps en avoir un même si le développeur oublie d'en mettre un?

Ldoppea (approved):

src/libs/ReactNativeLauncher.js
doubleface: Done

PR #897, opened 2023-08-14
feat: implement sharing files navigation (android + iOS)

acezard:
### Data flow:
```mermaid
graph TD
    subgraph State Layer
        SProvider[Sharing Provider]
    end

    subgraph View Layer
        SS[SharingScreen]
    end

    subgraph User Flow Layer
        UGA[useGlobalAppState]
    end

    subgraph Business Layer
        SH[SharingStatus]
        SService[SharingData]
    end

    subgraph Native Layer
        SIM[SharingIntentModule\nAndroid]
        RNRSI[RNReceiveSharingIntent]
    end

    SH -->|"SharingIntentStatus\nenum"| SProvider
    SService -->|"filesToUpload\nReceivedFile[]"| SProvider
    SProvider -->|"filesToUpload\nReceivedFile[]"| SS
    SProvider -->|"sharingIntentStatus\nenum"| UGA
    SIM -->|"wasOpenedViaSharing()\nboolean"| SH
    RNRSI -->|"Linking.openUrl()\niOS"| SH
    RNRSI -->|"ReceiveSharingIntent\ngetReceivedFiles()"| SService

    style SS fill:slateblue
    style UGA fill:darkslateblue
    style SH fill:darkslategray
    style SService fill:darkslategray
    style SIM fill:indianred
    style SProvider fill:forestgreen
    style RNRSI fill:indianred
```

### User flow:
```mermaid
graph TD
    A[User Opens App] --> |Without File Sharing| B
    A --> |With File Sharing| E
    B --> |No Lock Screen| C[Home Screen]
    B --> |Has Lock Screen| D
    D --> |Unlocks Device| C
    D --> |Stuck| D1[Stuck on Lock Screen]
    E --> |No Lock Screen| F[File Sharing Screen]
    E --> |Has Lock Screen| G
    F --> H[Home Screen After Upload Confirmation]
    G --> |Unlocks Device| F
    G --> |Stuck| G1[Stuck on Lock Screen]
```

### End to end manual testing 

- [x] **Cold Start with Sharing Intent**:
    - Ensure the app is completely closed (not in the background).
    - Share content to your app from another app.
    - Verify that the app opens and detects the sharing intent (`sharingIntentStatus` should be `undetermined` then `openedViaSharing`).

- [x] **Cold Start without Sharing Intent**:
    - Ensure the app is completely closed.
    - Open the app normally (without sharing).
    - Verify that the app opens and detects the sharing intent (`sharingIntentStatus` should be `undetermined` then `notOpenedViaSharing`).

- [x] **Resume from Background with Sharing Intent**:
    - Open the app normally.
    - Put the app in the background.
    - Share content to your app from another app.
    - Verify that the app comes to the foreground and detects the sharing intent (`sharingIntentStatus` should be `openedViaSharing`).

- [x] **Resume from Background without Sharing Intent**:
    - Open the app normally.
    - Put the app in the background.
    - Bring the app to the foreground without a sharing intent (e.g., from the recent apps list).
    - Verify that the app does not detect a sharing intent (`sharingIntentStatus` should be `notOpenedViaSharing`).

- [x] **Switching Intents**:
    - Open the app with a sharing intent.
    - Put the app in the background.
    - Bring the app to the foreground without a sharing intent.
    - Verify that the app does not detect a sharing intent (`sharingIntentStatus` should be `notOpenedViaSharing`).
    - Put the app back in the background.
    - Share content to your app from another app.
    - Verify that the app detects the sharing intent (`sharingIntentStatus` should be `openedViaSharing`).

- [x] **App Restart**:
    - Open the app with a sharing intent.
    - Restart the app (either via a setting in the app or by force-closing and reopening).
    - Verify that the app does not detect a previous sharing intent (`sharingIntentStatus` should be `notOpenedViaSharing`).


src/app/state/usePendingAction.ts
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
```
acezard, line 1: file to delete

src/app/view/sharing/useSharing.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import {
+  getFilesToUpload,
+  initSharingMode
+} from '/app/domain/sharing/SharingService'
+
+export const useSharingMode = (): {
+  filesToUpload: Record<string, unknown>[]
+} => {
+  useEffect(() => {
```
acezard, line 11: this effect has to be removed, sharingMode is already init elsewhere (globalAppState)

src/constants/routes.ts
```diff
@@ -13,6 +13,7 @@ export const routes = {
   securize: 'securize',
   setPassword: 'setPassword',
   setPin: 'setPin',
+  sharing: 'sharing',
   stack: 'stack',
   welcome: 'welcome'
-}
+} as const
```
acezard, line 19: better typing inference

src/hooks/useGlobalAppState.ts
```diff
@@ -1,116 +1,51 @@
...
-
-import * as RootNavigation from '/libs/RootNavigation'
-import { getData, StorageKeys, storeData } from '/libs/localStore/storage'
-import { routes } from '/constants/routes'
-import {
-  showSplashScreen,
-  hideSplashScreen
-} from '/app/theme/SplashScreenService'
-import { determineSecurityFlow } from '/app/domain/authorization/services/SecurityService'
+import { useNavigationState } from '@react-navigation/native'
```
acezard, line 1: this file now "orchestrates" services according to app state. It should be renamed for sure

src/app/view/sharing/useSharing.ts
acezard: - [x] done

src/app/domain/sharing/SharingService.ts
```diff
@@ -0,0 +1,59 @@
...
+let isServiceInitialized = false
+
+const sharingLogger = Minilog('🗃️ SharingService')
+
+const initSharingMode = (): void => {
+  sharingLogger.info('Init sharing mode')
+
+  if (isServiceInitialized) {
+    sharingLogger.info('Service already initialized')
+    return // Exit early if the service is already initialized
```
Ldoppea, line 15: nit: this comment doesn't add knowledge as the `if(isServiceInitialized) return` is already self explanatory. Also the `.log()` add even more explanation. So in my opinion we can remove this comment

src/hooks/useGlobalAppState.ts
```diff
@@ -142,6 +77,29 @@ const onStateChange = (
...
   const client = useClient()
+  const initFlag = useRef(false)
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  const routeIndex = useNavigationState(state => state?.index)
+
+  // On app start
+  useEffect(() => {
+    log.info('useGlobalAppState: app start')
+
+    if (getIsSecurityFlowPassed() && !initFlag.current) {
```
acezard, line 88: securityFlow on app launch (not wakeup) is launched from homeView. Not great but looked a bit tricky to change now

src/app/view/sharing/useSharing.ts
Ldoppea: Maybe the entire hook can be removed? There seems to be no reactive code here nor any event to listen, so maybe `getFilesToUpload()` can be called directly where needed?

src/app/view/sharing/SharingScreen.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [numberOfFiles, setNumberOfFiles] = useState(0)
+  const { filesToUpload } = useSharingMode()
+
+  useEffect(() => {
+    setNumberOfFiles(filesToUpload.length)
+
+    return () => {
+      setNumberOfFiles(0)
+    }
+  }, [filesToUpload])
```
Ldoppea, line 16: Why do we need this effect? Refering `filesToUpload.length` directly in the jsx code seems enough?

src/hooks/useGlobalAppState.ts
```diff
@@ -1,116 +1,51 @@
...
-          navigation: RootNavigation.navigationRef as NavigationContainerRef<
-            Record<string, unknown>
-          >,
-          href: parsedRoute.params.href,
-          slug: parsedRoute.params.slug
-        }
-      : undefined
-  )
-}
+initSharingMode()
```
Ldoppea, line 30: If this is called in the script's root, this means we have no control of when this is called (i.e. this will be called as soon as we import `useGlobalAppState.ts`). So why putting this here but not directly in the `SharingService.ts`? (although it would be greater if done somewhere in the react lifecycle)

src/hooks/useGlobalAppState.ts
```diff
@@ -1,116 +1,51 @@
...
-        )
-        return hideSplashScreen()
-      }
-    })
-    .catch(reason => log.error('Failed when waking up', reason))
+  handleSecurityFlowWakeUp(client)
+
+  if (isSharingMode()) {
+    log.info('useGlobalAppState: handleWakeUp, sharing mode')
+    navigate(routes.sharing)
```
Ldoppea, line 47: As `handleSecurityFlowWakeUp` has async side effect that call `navigate`. Does this means that `navigate(routes.sharing)` will have race condition with `handleSecurityFlowWakeUp`?

src/hooks/useGlobalAppState.ts
```diff
@@ -142,6 +77,29 @@ const onStateChange = (
...
+        navigate(routes.sharing)
+      } else {
+        log.info('useGlobalAppState: app start, not sharing mode')
+      }
+
+      initFlag.current = true
+    } else {
+      log.info('useGlobalAppState: app start, security flow not passed')
+    }
+  }, [routeIndex])
```
Ldoppea, line 102: Does this mean that this effect will be called on every navigation (i.e. when home is opening a cozy-app)?

src/hooks/useGlobalAppState.ts
```diff
@@ -142,6 +77,29 @@ const onStateChange = (
...
+      log.info('useGlobalAppState: app start, security flow passed')
+
+      if (isSharingMode()) {
+        log.info('useGlobalAppState: app start, sharing mode')
+        navigate(routes.sharing)
+      } else {
+        log.info('useGlobalAppState: app start, not sharing mode')
+      }
+
+      initFlag.current = true
```
Ldoppea, line 98: Seems like you never revert `initFlag.current` to `false`. Does this mean that this effect can happens only one single time? What happens if the apps goes to background, locks, then receive a sharing, multiple times in the same app lifecycle?

src/app/view/sharing/SharingScreen.tsx
acezard: I guess yes, my only issue is calling the getter too early for instance.

src/hooks/useGlobalAppState.ts
acezard: yes initially I put it in App.js. I think I should use it once it the main useGlobalAppState() hook (which is rendered by App).

src/hooks/useGlobalAppState.ts
acezard: I didn't have issues because it redirects to home then sharing in sequence. I shall look into possible errors

src/hooks/useGlobalAppState.ts
acezard: yes, it's not great, but I need to listen to something. Maybe I could replace it with an event listener

src/hooks/useGlobalAppState.ts
acezard: the sharing flow is discarded as soon as the app goes into background/closes. At least I understood the spec like that

src/app/view/sharing/SharingProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const eventCallback = (files: ReceivedFile[]): void => {
+    sharingLogger.info(`SharingProvider Received ${files.length} new files`)
+    setFilesToUpload(files)
+  }
+
+  useEffect(() => {
+    // Start listening for incoming files
+    handleReceivedFiles((files: ReceivedFile[]) => {
+      sharingLogger.info('Emitting received files', files)
+      eventEmitter.emit(RECEIVED_NEW_FILES, files)
```
Ldoppea, line 40: Why not calling `eventCallback` here directly instead of using an `eventEmitter`?

src/app/domain/sharing/services/SharingService.ts
```diff
@@ -0,0 +1,75 @@
...
+import Minilog from 'cozy-minilog'
+
+import {
+  isAndroidFile,
+  isIOSFile,
+  ReceivedFile
+} from '/app/domain/sharing/models/ReceivedFile'
+
+export const sharingLogger = Minilog('🗃️ SharingService')
+const filesToUpload = new Map<string, ReceivedFile>()
```
Ldoppea, line 12: Do we really need this local state? Seems like `getFilesToUpload()` is never called outside of this file. And `SharingProvider` also stores this info in a reactive way.

The only methods that needs this is `hasFilesToUpload` but I'm wondering if we can move it in `SharingProvider`, can't we?

src/app/domain/sharing/services/SharingService.ts
```diff
@@ -0,0 +1,75 @@
...
+    filesToUpload.clear()
+  }
+
+  // Populate the `filesToUpload` Map with the files to upload
+  for (const file of files) {
+    let key: string | undefined | null
+
+    if (isAndroidFile(file)) {
+      key = file.filePath
+    } else if (isIOSFile(file)) {
```
Ldoppea, line 31: There seems to be no difference between iOS and Android? Bot do `key = file.filePath`

src/hooks/useGlobalAppState.ts
acezard: any race condition should now be handled since we await for securityFlow end on wakeup

src/app/domain/sharing/index.ts
```diff
@@ -0,0 +1,3 @@
+import Minilog from 'cozy-minilog'
+
+export const sharingLogger = Minilog('🗃️ SharingService')
```
Ldoppea, line 3: Do we really want to have a file just to make the economy of one variable declaration?

src/app/view/sharing/SharingProvider.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export const useSharingState = (): SharingState => {
+  const context = React.useContext(SharingStateContext)
+  if (context === undefined) {
+    throw new Error('useSharingState must be used within a SharingProvider')
+  }
+  return context
+}
+
+export const useSharingDispatch = (): Dispatch<SharingAction> => {
```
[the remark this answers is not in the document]
acezard: yes it will be centralizing post-me logic/state

src/app/domain/sharing/index.ts
acezard: multiple files use it and there's not really a single file more important than the others so in any case I would put it in a file apart

src/app/view/sharing/SharingProvider.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    return () => {
+      cleanupSharingIntent()
+    }
+  }, [dispatch])
+
+  return (
+    <SharingStateContext.Provider value={state}>
+      <SharingDispatchContext.Provider value={dispatch}>
+        {children}
```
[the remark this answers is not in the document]
acezard: I checked and didn't notice post-me issues but we will definitely see them arise in the upload state if they are possible

src/app/domain/sharing/services/SharingStatus.ts
```diff
@@ -0,0 +1,76 @@
...
+  Platform
+} from 'react-native'
+
+import { SharingIntentStatus } from '/app/domain/sharing/models/SharingState'
+import { sharingLogger } from '/app/domain/sharing'
+
+export const handleSharing = (
+  setStatus: (status: SharingIntentStatus) => void
+): (() => void) => {
+  const nativeModule = NativeModules as {
```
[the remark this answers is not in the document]
acezard: it will be null on iOS (sharingIntentModule, not NativeModules)

src/app/view/sharing/SharingProvider.tsx
```diff
@@ -0,0 +1,88 @@
...
+  })
+
+  useEffect(() => {
+    const cleanupSharingIntent = handleSharing(
+      (status: SharingIntentStatus) => {
+        dispatch({ type: SharingActionType.SetIntentStatus, payload: status })
+      }
+    )
+
+    handleReceivedFiles(files => {
```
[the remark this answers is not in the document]
acezard: updated, but at this time the library only provides `ReceiveSharingIntent.clearReceivedFiles()`, and I'm not sure it's good enough. It might be needed to handle this ourselves in native code

PR #898, opened 2023-08-18
Refactor `settings.gradle` to remove duplicate module references

acezard:
In this PR, we've made the following improvements to the `settings.gradle` file:

- Removed the redundant inclusion of the `@fengweichong/react-native-gzip` module. It was creating an error because of an import duplication: `Multiple projects in the build are located in the same directory: /home/anc/cozy/cozy-react-native/node_modules/@fengweichong/react-native-gzip/android`
- Formatted the file for better readability by adding spaces between module includes.

By making these changes, we've resolved the Java workspace error that was caused due to multiple projects being located in the same directory.

### DIFF
--- a/android/settings.gradle
+++ b/android/settings.gradle
@@ -1,13 +1,14 @@
 rootProject.name = 'CozyReactNative'
+
 include ':react-native-device-info'
 project(':react-native-device-info').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-device-info/android')
+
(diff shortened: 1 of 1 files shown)


Ldoppea:
Just to be sure, you removed `react-native-gzip` from settings.gadle because itwas an import duplication. But where is it imported at the first place?

Also you removed duplication for `react-native-fs` but did not mention it in the PR description.

PR #899, opened 2023-08-18
feat: implement webapp routing

acezard:
description to be completed


src/app/domain/sharing/models/SharingState.ts
```diff
@@ -8,14 +8,30 @@ export enum SharingIntentStatus {
...
   | { type: SharingActionType.SetIntentStatus; payload: SharingIntentStatus }
   | { type: SharingActionType.SetFilesToUpload; payload: ReceivedFile[] }
+  | {
+      type: SharingActionType.SetRouteToUpload
+      payload: { href: string; slug: string }
+    }
+  | { type: SharingActionType.SetFlowErrored; payload: boolean }
+  | { type: SharingActionType.SetRecoveryState }
+
+export interface ServiceResponse<T> {
```
Ldoppea, line 34: nit: Seems like you cannot have both result and error filled simultaneously, so I would reflect that in the type:

```typescript
interface ServiceResponseOk<T> {
  result: T
}

interface ServiceResponseError {
  error: string
}

export type ServiceResponse<T> = ServiceResponseOk<T> | ServiceResponseError
```

Hower this would complexify a bit type checks (exemple [here](https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/clientHelpers/types.ts#L70-L99) and [here](https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/clientHelpers/connectClient.ts#L47-L66)) but it ensures not type overlap is possible

src/hooks/useGlobalAppState.ts
```diff
@@ -75,51 +75,63 @@ const onStateChange = (
...
+  const client = useClient()
+  const { sharingIntentStatus, errored } = useSharingState()
 
   useEffect(() => {
     let subscription: NativeEventSubscription | undefined
 
     // If there's no client, we don't need to listen to app state changes
     // because we can't lock the app anyway
-    if (client && !subscription) {
+    if (!hasExecuted.current && client && !subscription) {
```
Ldoppea, line 96: Does this mean that if `client` changes (or any other array dependency) the subscription will be removed but will never register again (as `hasExecuted.current` is never set back to `false`)? Is that expected?

acezard: I will double-check the lifecycle of this event listener, I didn't notice a removal that broke my appStatus change flow but it's better to be safe

src/App.js
```diff
@@ -132,7 +132,9 @@ const WrappedApp = () => {
   if (client)
     return (
       <CozyProvider client={client}>
-        <Nav client={client} setClient={setClient} />
+        <SharingProvider>
```
Crash--: No way we call that SharingProvider, like I said you yesterday. We already have a `SharingProvider` https://github.com/cozy/cozy-libs/blob/master/packages/cozy-sharing/src/SharingProvider.jsx that doesn't do the same thing.

src/app/domain/sharing/models/SharingCozyApp.ts
```diff
@@ -0,0 +1,21 @@
...
+    accept_documents_from_flagship?: {
+      accepted_mime_types: string[]
+      max_number_of_files: number
+      max_size_per_file_in_MB: number
+      route_to_upload: string
+    }
+    accept_from_flagship?: boolean
+  }
+  slug: string
+}
```
Crash--, line 21: We should definitly stop adding types here but adding types to cozy-client. We have the type for the konnector https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/types/types.d.ts#L47 we should add the one for the app. 

This current type miss a lot of stuff. You can find the all stuff here https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.apps.md and there https://docs.cozy.io/en/cozy-apps-registry/#properties-meaning-reference

By the way, you should update the documentation  in order to add the new stuff in there.

src/app/domain/sharing/services/SharingNetwork.ts
```diff
@@ -0,0 +1,48 @@
+import CozyClient, { generateWebLink, Q } from 'cozy-client'
+
+import { sharingLogger } from '..'
+import { SharingCozyApp } from '../models/SharingCozyApp'
+
+export const fetchSharingCozyApps = {
+  definition: Q('io.cozy.apps').where({
+    'accept_documents_from_flagship.route_to_upload': { $exists: true },
```
Crash--, line 9: You should not add the `$exists:true` but instead you should index on this attribute. 

https://docs.cozy.io/en/tutorials/data/advanced/#indexes-why-is-my-document-not-being-retrieved

src/app/domain/sharing/services/SharingNetwork.ts
```diff
@@ -0,0 +1,48 @@
+import CozyClient, { generateWebLink, Q } from 'cozy-client'
+
+import { sharingLogger } from '/app/domain/sharing'
+import { SharingCozyApp } from '/app/domain/sharing/models/SharingCozyApp'
+import { ServiceResponse } from '/app/domain/sharing/models/SharingState'
+
+export const fetchSharingCozyApps = {
```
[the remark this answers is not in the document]
Crash--: Yep, by default, Q('io.cozy.apps') will only have the `source === stack` https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/AppCollection.js#L55 . So you'll only request the installed applications. 

Before you should have just add the `registry` to the source, but now, you have to make a different Query : Q('io.cozy.apps_registry') that will list the apps from the registry. https://github.com/cozy/cozy-client/pull/1376

src/app/domain/sharing/services/SharingNetwork.ts
```diff
@@ -0,0 +1,48 @@
...
+import { sharingLogger } from '..'
+import { SharingCozyApp } from '../models/SharingCozyApp'
+
+export const fetchSharingCozyApps = {
+  definition: Q('io.cozy.apps').where({
+    'accept_documents_from_flagship.route_to_upload': { $exists: true },
+    accept_from_flagship: true
+  }),
+  options: {
+    as: 'io.cozy.apps/fetchSharingCozyApps'
```
Crash--, line 13: We should not call the query by the functional meaning but by what it does: https://github.com/cozy/cozy-guidelines/#naming-of-queries

src/app/domain/sharing/services/SharingNetwork.ts
```diff
@@ -0,0 +1,48 @@
...
+      cozyApp.attributes.accept_documents_from_flagship?.route_to_upload
+    const slug = cozyApp.slug
+    if (!hash || !slug) return undefined
+
+    const href = generateWebLink({
+      cozyUrl: client.getStackClient().uri,
+      pathname: '',
+      slug,
+      subDomainType: session.subDomainType,
+      hash: hash.replace(/^\/?#?\//, ''),
```
Crash--, line 38: Why do we have to make this replace?

src/app/view/sharing/SharingProvider.tsx
```diff
@@ -41,10 +54,31 @@ interface SharingProviderProps {
...
+  const { data: fetchedApps } = useQuery(
+    fetchSharingCozyApps.definition,
+    fetchSharingCozyApps.options
+  )
+
+  useEffect(() => {
+    if (client === null) return
+
+    const route = getRouteToUpload(
+      fetchedApps as SharingCozyApp[],
```
Crash--: Are we ok to call this `getRouteToUpload` with a `fetchedApps` equals to null because fetchStatus===loading/pending/errored  ?

src/app/domain/sharing/services/SharingNetwork.ts
acezard: it's to handle route_to_upload looking like that: `"route_to_upload": "/#/createfromFlagshipUpload=true"`. I did not find a way to handle it directly with the generateWebLink(). I agree it doesn't look great overall. Probably best to remove the leading `/#` in the manifest. I don't think the manifest has to be aware we are using hash router anyway

src/app/view/sharing/SharingProvider.tsx
acezard: In my mind yes because it defers the null check to the low-level function (it will handle null or empty array data). That way the `SharingProvider` (to be renamed as we said) can focus on the user flow logic without too much coupling to data handling

src/app/domain/sharing/services/SharingNetwork.ts
Crash--: I don't know, actually in the manifest, we can define a few routes. Those routes are used by the server, so if you create a `/test` route, then the url will be : foo.mycozy.cloud/test. 

I don't know if it's a good idea to use `/createfromFlagship` directly in the manifest, since we can think this will be related to the server. 

The manifest is at the intersection between the server and the front end app. 

The manifest is linked to the app, so the manifest knows about the structure of the app and knows if the app use an hash router or not. 

@Ldoppea do you have a better idea?

src/app/view/sharing/SharingProvider.tsx
Crash--: if you say so, but you have to deal with the fact that the request can fail (fetchStatus===errored) somewhere.

acezard: > if you say so, but you have to deal with the fact that the request can fail (fetchStatus===errored) somewhere.

true, I didn't test for this scenario.

- [ ] To update

src/app/domain/sharing/services/SharingNetwork.ts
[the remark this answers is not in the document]
Ldoppea: Feedback from our latest meeting: We just changed the specs to optimize development delays
- We will start to list only installed app
- And then list "non installed" app **only after implementing mespapiers**.

So for this PR, listing `io.cozy.apps` is expected

cc @Crash--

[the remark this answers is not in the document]
acezard: (I will still update index/alias)

acezard:
closing for better readability, every commit can be seen on the last branch/pr: https://github.com/cozy/cozy-flagship-app/pull/906 (every comment remains meaningful and will be updated on the main branch)

PR #900, opened 2023-08-18
feat: Implement OAuthClientsLimitExceeded dialog

Ldoppea:
Cozy-stack will now limit the number of registered OAuth clients based on `cozy.oauthclients.max` flag

On web, when the limit is reached and the user tries to open a cozy-app then the browser is redirected to `/settings/clients/limit-exceeded`

On Flagship app, we want to intercept this mechanism in order to keep control on how application behaves

To handle this we manually do the limit check in `CozyProxyWebView` before initializing HTML content. This would allow us to handle both Stack and HttpServer based cozy-apps

Then we display a dedicated WebView that will display the `/settings/clients/limit-exceeded` route from cozy-stack

Also we chose to always display cozy-home even if the limit is reached. This is possible because cozy-home is always served from the local HttpServer and this will ensure we don't have side effects as the Flagship app architecture has been created around the fact that the HomeView is always displayed

Related PR: https://github.com/cozy/cozy-stack/pull/4080
Related PR: https://github.com/cozy/cozy-stack/pull/4086
Related PR: https://github.com/cozy/cozy-stack/pull/4087

___

### TODO:

- [x] ~~Wait for cozy-stack related PRs to be merged before merging this PR~~ The feature is now protected by a flag
- [x] ~~Check compatibility with the sharing screen that will come with #897~~ This PR is merged before the sharing screen PR so we will have to do the verification later

### DIFF
--- a/src/app/domain/limits/OauthClientsLimitService.ts
+++ b/src/app/domain/limits/OauthClientsLimitService.ts
@@ -20,9 +20,9 @@ export const OAUTH_CLIENTS_LIMIT_EXCEEDED = 'OAUTH_CLIENTS_LIMIT_EXCEEDED'
 
 export const oauthClientLimitEventHandler = new EventEmitter()
 
-export const showOauthClientsLimitExceeded = (): void => {
+export const showOauthClientsLimitExceeded = (href: string): void => {
   navigate('home')
-  oauthClientLimitEventHandler.emit(OAUTH_CLIENTS_LIMIT_EXCEEDED)
+  oauthClientLimitEventHandler.emit(OAUTH_CLIENTS_LIMIT_EXCEEDED, href)
 }
 
 export const interceptNavigation =
--- a/src/app/domain/limits/OauthClientsLimitService.ts
+++ b/src/app/domain/limits/OauthClientsLimitService.ts
@@ -12,6 +12,7 @@ import Minilog from 'cozy-minilog'
 
 import { routes } from '/constants/routes'
 import { navigateToApp } from '/libs/functions/openApp'
+import { getErrorMessage } from '/libs/functions/getErrorMessage'
 import { navigate, navigationRef } from '/libs/RootNavigation'
 
 const log = Minilog('⛔ OAuth Clients Limit Service')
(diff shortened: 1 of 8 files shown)


src/app/view/Limits/hooks/useOAuthClientsLimitExceeded.ts
```diff
@@ -0,0 +1,73 @@
...
+  const [popupUrl, setPopupUrl] = useState<string | null>(null)
+
+  const closePopup = (): void => {
+    setPopupUrl(null)
+  }
+
+  useEffect(() => {
+    const eventCallback = (href: string): void => {
+      if (client === null) {
+        log.error('Client is null, should not happen')
```
acezard, line 42: if it's really impossible to happen maybe we should reflect it in the typing. if it's possible but very unlikely, shouldn't we logout() the user?

src/app/domain/limits/checkOauthClientsLimit.ts
```diff
@@ -0,0 +1,30 @@
...
+export const checkOauthClientsLimitForApp = async (
+  slug: string,
+  client: CozyClient
+): Promise<boolean> => {
+  if (slug === 'home' || slug === 'settings') {
+    return false
+  }
+
+  const stackClient = client.getStackClient()
+  const result = await stackClient.fetchJSON<ClientsUsageResult>(
```
acezard, line 21: what should happen if the request fails here?

src/app/view/Limits/OauthClientsLimitExceeded.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <WebView
+        source={{ uri: popupUrl }}
+        onShouldStartLoadWithRequest={interceptNavigation}
+        onOpenWindow={interceptOpenWindow}
+      />
+    </View>
+  ) : null
+}
+
+const styles = StyleSheet.create({
```
acezard, line 40: shouldn't we try to fit this dialog into our local "cozy-ui"? it seems like a reusable thing

src/app/domain/limits/OauthClientsLimitService.ts
```diff
@@ -0,0 +1,113 @@
...
+  WebViewNavigation
+} from 'react-native-webview/lib/WebViewTypes'
+
+import type CozyClient from 'cozy-client'
+import { deconstructCozyWebLinkWithSlug } from 'cozy-client'
+import Minilog from 'cozy-minilog'
+
+import { routes } from '/constants/routes'
+import { navigateToApp } from '/libs/functions/openApp'
+import { navigate, navigationRef } from '/libs/RootNavigation'
```
acezard, line 15: I've thought about this recently and I've come to the conclusion a lot of issues we have with navigation come from react-decoupled calls, that don't fit well with the view lifecycles. I started to avoid as much as possible calling them from outside components in the sharing feature. What do you think?

acezard (commented):
- Errors: what happens in case of errors? I guess in the absence of client the navigation will reset to login, but without deleting local data?
- Testing: I think tests should be useful here as some functions present a decent amount of logic, the hook could be tested as well, maybe through the component

src/app/domain/limits/OauthClientsLimitService.ts
Ldoppea: I tried to use the `navigation` param only, but the problem is that `navigation.getCurrentRoute()` will always return the home's route here, even if cozy-settings is on focus. Seems like the `navigation` prop's context is bind to where the component is displayed and not where the app is.

When using `navigationRef.getCurrentRoute()`, then I correctly get the cozyapp route when cozy-settings is displayed.

Ldoppea:
> * Testing: I think tests should be useful here as some functions present a decent amount of logic, the hook could be tested as well, maybe through the component

Damned I forgot them, I focused too much on creating the demo video before leaving and then forgot to add tests. I'll create them.

acezard (approved):

PR #910, opened 2023-08-25
feat: Check log allowed level

doubleface:
See https://github.com/konnectors/libs/pull/964, this check was done at
contentscript level at first but we also need this check at the launcher
level.
If this check is not done, the flagship app will be stuck trying to send logs and get an error from the stack because the log level is unknown

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -97,10 +97,17 @@ class ReactNativeLauncher extends Launcher {
    *
    * @param  {ContentScriptLogMessage} message - log message
    */
-  log({ timestamp, ...logContent }) {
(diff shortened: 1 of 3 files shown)


LucsT (approved):

src/libs/ReactNativeLauncher.js
```diff
@@ -97,17 +97,25 @@ class ReactNativeLauncher extends Launcher {
    *
    * @param  {ContentScriptLogMessage} message - log message
    */
-  log({ timestamp, ...logContent }) {
+  log({ timestamp, level, ...logContent }) {
```
Crash--, line 100: please update the JSDoc 🙏

PR #911, opened 2023-08-25
Support notification permission for Android 13

zatteo:


src/hooks/useNotifications.ts
```diff
@@ -21,11 +20,11 @@ export const useNotifications = (): void => {
...
       if (!client) return
 
-      const permission = await requestAndGetNotificationPermission()
+      const permission = await requestNotifications()
 
-      if (permission === messaging.AuthorizationStatus.AUTHORIZED) {
+      if (permission.granted) {
         setAreNotificationsEnabled(true)
-      } else if (permission === messaging.AuthorizationStatus.DENIED) {
+      } else if (!permission.granted && !permission.canRequest) {
```
Ldoppea, line 27: nit `!permission.granted` is always true here as you are in the `else` part of `if(permission.granted)`

Ldoppea (approved):

src/hooks/useNotifications.ts
zatteo: Agree (and eslint told me this too 😄). But I find it more clear to have explicitly a case "granted" and a case "not granted and not requestable"

PR #913, opened 2023-08-28
Manage iCloud Shared Albums

zatteo:
**iOS only.** Previously, we created albums for each iCloud Shared Albums, but we were unable to upload medias from these albums.

Now, with a flag, we can chose what we want : 
- create albums from iCloud Shared Albums and upload correctly medias from these albums
- do nothing (so not creating album from iCloud Shared Albums and not uploading medias from these albums)

Related : 
https://github.com/react-native-cameraroll/react-native-cameraroll/pull/525
https://github.com/react-native-cameraroll/react-native-cameraroll/pull/526


Crash-- (approved):
Nice

PR #914, opened 2023-08-29
Fix/errors handling backup

zatteo:
refactor(backup): Better error handling

3 last backup possibilities :
- success : success notification + success message
- partial success : success notification + partial success message with reason
- error : no notification + error message with reason

Manage "file too big" error.

Clearer: Backup is now always set as 'done' when finished, even if there was an error or it if was manually stopped (no more sneaky set as 'ready').

Linked to https://github.com/cozy/cozy-drive/pull/2976

### DIFF
--- a/src/app/domain/backup/hooks/useInitBackup.tsx
+++ b/src/app/domain/backup/hooks/useInitBackup.tsx
@@ -4,8 +4,10 @@ import { useClient } from 'cozy-client'
 
 import {
   getLocalBackupConfig,
-  setBackupAsToDo
+  setBackupAsDone,
+  setLastBackup
 } from '/app/domain/backup/services/manageLocalBackupConfig'
+import { t } from '/locales/i18n'
 
 import Minilog from 'cozy-minilog'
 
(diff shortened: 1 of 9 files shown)


src/app/domain/backup/hooks/useInitBackup.tsx
```diff
@@ -25,7 +27,16 @@ export const useInitBackup = (): void => {
 
         if (backupConfig.currentBackup.status === 'running') {
           log.debug('A running backup has been set as to do.')
```
Crash--, line 29: does this log still accurate?

Crash-- (approved):
Should we add UT for the upload process? In order to test the error cases and how we handle them?

zatteo:
> Should we add UT for the upload process? In order to test the error cases and how we handle them?

It would be very nice. I will see if I have time to add some tests this week. But it is right that it will help to understand the multiple error cases.

src/app/domain/backup/hooks/useInitBackup.tsx
zatteo: It was not 😄

PR #915, opened 2023-08-29
Open flagship app and redirect to cozy app when opening local notification

zatteo:
Open flagship app and redirect to cozy app when opening local notification

### DIFF
--- a/src/libs/notifications/notifications.ts
+++ b/src/libs/notifications/notifications.ts
@@ -1,6 +1,4 @@
-import messaging, {
-  FirebaseMessagingTypes
-} from '@react-native-firebase/messaging'
+import messaging from '@react-native-firebase/messaging'
 
 import Minilog from 'cozy-minilog'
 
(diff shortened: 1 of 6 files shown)


src/libs/notifications/notifications.ts
```diff
@@ -62,16 +54,18 @@ export const navigateFromNotification = async (
...
+  if (notification?.data?.redirectLink) {
+    await navigateFromNotification(client, notification.data.redirectLink)
   }
 }
 
 export const handleNotificationOpening = (client: CozyClient): (() => void) => {
   return messaging().onNotificationOpenedApp(async notification => {
-    await navigateFromNotification(client, notification)
+    if (notification.data?.redirectLink) {
+      await navigateFromNotification(client, notification.data.redirectLink)
```
Crash--, line 76: we don't have to edit the jsdoc / type of this `navigateFromNotification` ?

Crash-- (approved):

src/libs/notifications/notifications.ts
zatteo: I updated the type in the first commit.

Crash--: Just a few lines above. Wondering why I didn't see it first..

PR #920, opened 2023-09-04
fix(backup): Stop background backup when not supported

zatteo:
On iOS, it is not supported.

On Android, it is supported until Android 11 because android-upload-service uses Service to start foreground task in background which is deprecated since Android 12.

### DIFF
--- a/src/app/domain/backup/services/manageBackup.ts
+++ b/src/app/domain/backup/services/manageBackup.ts
@@ -124,6 +124,20 @@ export const startBackup = async (
           postUploadLocalBackupConfig.currentBackup.totalMediasToBackupCount,
         message: partialSuccessMessage
       })
+
+      await showLocalNotification({
+        title: t('services.backup.notifications.backupPartialSuccessTitle'),
+        body: t('services.backup.notifications.backupPartialSuccessBody', {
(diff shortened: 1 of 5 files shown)


src/app/domain/backup/services/manageBackup.ts
```diff
@@ -124,6 +124,20 @@ export const startBackup = async (
           postUploadLocalBackupConfig.currentBackup.totalMediasToBackupCount,
         message: partialSuccessMessage
       })
+
+      await showLocalNotification({
```
Ldoppea, line 128: nit: The only difference between `if` and `else` blocks are the status, and the title/body translation keys. So you can ease identifying differences by:

```js
let status = 'success'
let titleKey = 'services.backup.notifications.backupSuccessTitle'
let bodyKey = 'services.backup.notifications.backupSuccessBody'

if (partialSuccessMessage) {
  status = 'partial_success'
  titleKey = 'services.backup.notifications.backupPartialSuccessTitle'
  bodyKey = 'services.backup.notifications.backupPartialSuccessBody'
}

await setLastBackup(...)

await showLocalNotification(...)
```

Ldoppea (approved):

PR #924, opened 2023-09-08
fix: Restore pod cache before installing node_modules on IOS CI

Ldoppea:
In 3b14bdb041d659addd79aa4883249f55cbf2acba we triggered the pods installation from yarn's `postinstall` event

The impact is that the pod cache should now be restored before calling `yarn install`

This would reduce the worlkflow duration by 5 to 10 min

This commit also move the Ruby installation before the `yarn install` so the CI steps are grouped in the following order:
- Environment installation
- Cache restoration
- Dependencies (node and pods) installation
- Build phase

Related commit: 3b14bdb041d659addd79aa4883249f55cbf2acba

### DIFF
--- a/.github/workflows/build-ios-dev.yml
+++ b/.github/workflows/build-ios-dev.yml
@@ -9,13 +9,19 @@ jobs:
     runs-on: macos-latest
 
     steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v3
+
       - name: Set NodeJS version
         uses: actions/setup-node@v3
         with:
           node-version: 16
 
-      - name: Check out Git repository
-        uses: actions/checkout@v3
+      - name: Setup Ruby (bundle)
+        uses: ruby/setup-ruby@v1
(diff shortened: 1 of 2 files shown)


Ldoppea:
Note that we may want to also remove the `Install Pods` step as it is redundant with the yarn's `postinstall`. However this would reduce the duration by only half a minute, so this is not critical.

@zatteo do you know if there are some differences between the node `pod-install` action and a raw `pod install`?

Ldoppea:
For now I didn't feel any regression on my workflow, but I don't yarn install everyday. Were you impacted?

zatteo:
> For now I didn't feel any regression on my workflow, but I don't yarn install everyday. Were you impacted?

Not sure. But now you know, if it slows us down too much we can remove it. I will pay attention.

zatteo (approved):

zatteo:
Well done for the optimization!

PR #925, opened 2023-09-08
Feat/retry upload

zatteo:
**feat: Add an upload with a retry strategy**

Allow to retry n times an upload.
Allow to specify errors triggering retry.
Default to no retry.

**feat(backup): Retry upload 1 time for some errors**

### DIFF
--- a/src/app/domain/upload/services/index.spec.ts
+++ b/src/app/domain/upload/services/index.spec.ts
@@ -1,0 +1,173 @@
+import { uploadFile } from '/app/domain/upload/services/upload'
+
+import * as UploadService from './index'
+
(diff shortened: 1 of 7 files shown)


src/app/domain/upload/services/index.spec.ts
```diff
@@ -0,0 +1,173 @@
+import { uploadFile } from '/app/domain/upload/services/upload'
+
+import * as UploadService from './index'
+
+import { IOCozyFile } from 'cozy-client'
```
Crash--, line 5: import type xxx, no?

Crash-- (approved):

PR #927, opened 2023-09-11
feat: (Clisk) Catch and log downloadFileInWorker errors

doubleface:
If there was an error in the downloadFileInWorker function, it would
cause the whole clisk konnector run to fail

Now, the error is caught and a warning error will be visible in the
konnector logs.

This is also coherent with the behavior of node konnectors.

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/libs/Launcher.js
+++ b/src/libs/Launcher.js
@@ -458,11 +458,25 @@ export default class Launcher {
       sourceAccount: job.message.account,
       sourceAccountIdentifier,
       // @ts-ignore
-      downloadAndFormatFile: async entry => ({
-        ...entry,
-        // @ts-ignore
-        dataUri: await this.worker.call('downloadFileInWorker', entry)
-      }),
+      downloadAndFormatFile: async entry => {
+        try {
+          // @ts-ignore
+          const dataUri = await this.worker.call('downloadFileInWorker', entry)
+          return {
+            ...entry,
+            dataUri
+          }
+        } catch (err) {
+          this.log({
+            level: 'warning',
+            // @ts-ignore
+            message: `Could not download entry: ${entry.fileurl}: ${err.message}`,
+            label: 'downloadAndFormatFile',
+            namespace: 'Launcher'
+          })
+          return entry
+        }
+      },
       existingFilesIndex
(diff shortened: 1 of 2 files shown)


Crash-- (commented):
So the execution is marked as successful, even though we failed to download the file? Right? So we'll have a lastSuccessDate to the date of execution and if we then want to optimize by recovering only what was done after the last success date, we'll be stuck. 


If the download has failed we should: 
- Retry once or twice 
- If after those retries it still fails, then we should throw an error. 

I really don't like the fact we don't tell the user that something was not working as expected.

doubleface:
@Crash-- But if we fail in the download of a file, we would not try to download the other files, which seems to me to be a poorer service. Is it really what we want ?

FYI we have a [retry](https://github.com/konnectors/libs/blob/8b2ecac5417090322025a92918b39577e96b3403/packages/cozy-clisk/src/launcher/saveFiles.js#L215) option in saveFiles but which is set to 0 by default (this comes from the nodejs implementation).

Crash--:
> @Crash-- But if we fail in the download of a file, we would not try to download the other files, which seems to me to be a poorer service. Is it really what we want ?

I don't have proposed anything like that ;). I just say that we should throw an error somewhere and don't tell the user that everything was good. There is a lot of possibility here. 

To move forward: 
- How can we have an error when downloading? What would be the causes? 
  - Internet connexion : After a retry, if there is still an internet connexion issue, then we should not try to download other files and we should stop 
  - Server down: Same
  - Other causes ?

doubleface:
> How can we have an error when downloading? What would be the causes?

>    - Internet connexion : After a retry, if there is still an internet connexion issue, then we should not try to download other > files and we should stop
>    - Server down: Same
>    - Other causes ?

- a type of file unexpected by the konnector for a type of account we don't have
- a temporary server side error, caused by too many file downloads in a row (retry will help here)

Crash--:
> a type of file unexpected by the konnector for a type of account we don't have

In that case, what should we do? Be silent? Throw an error? Let's see with the product team.

PR #928, opened 2023-09-13
Feat/update local data backup

zatteo:
This PR has 3 goals : 
- store more data in local storage for backed up media to be more future proof
- change the way we compare local media and backed up media
  - now we just compare local uri (which is saved when a file is uploaded or a restoration is done)
- change the way we consider an external media restored from Cozy Drive to be identical to a local media (instead of _name + path_ we use _name + creation date + modification date_)

### DIFF
--- a/src/@types/cozy-client.d.ts
+++ b/src/@types/cozy-client.d.ts
@@ -120,8 +120,12 @@ declare module 'cozy-client' {
     }[]
   }
 
+  interface MissingFileDocumentAttributes {
+    md5sum: string
+  }
+
   type IOCozyFile = {
-    attributes: FileDocument
+    attributes: MissingFileDocumentAttributes & FileDocument
   } & CozyClientDocument
 
   interface Collection {
--- a/src/app/domain/backup/models/Media.ts
+++ b/src/app/domain/backup/models/Media.ts
@@ -10,6 +10,7 @@ import { Album } from '/app/domain/backup/models'
  */
 export interface Media {
   name: string
+  uri: string
   path: string
   remotePath: string
   type: 'image' | 'video'
@@ -23,12 +24,18 @@ export interface Media {
 
 /**
  * A selection of media metadata stored locally to identify if a media has already been backuped
- * @member {string} name
- * @member {string} remotePath e.g. /Sauvegardé depuis mon mobile/My Android/Download
- * @member {string} remoteName e.g. IMG_001 (1).jpg (added only if different from name)
+ * @member {string} name e.g. "IMG_50.jpg"
+ * @member {string} uri e.g. "file:///storage/emulated/0/Download/IMG_50.jpg" on Android or "ph://0E75CF5E-4587-4BD7-AB07-E940BB627C4B/L0/001" on iOS
+ * @member {number} creationDate e.g. 1694523521000
+ * @member {number} modificationDate e.g. 1694172391000
+ * @member {string} remoteId e.g. "d78a3c84139d173dde3b87df0003b32e"
+ * @member {string} remotePath e.g. "/Sauvegardé depuis mon mobile/My Android/Download"
  */
 export interface BackupedMedia {
   name: string
+  uri: string
+  creationDate: number
+  modificationDate: number
+  remoteId: string
   remotePath: string
-  remoteName?: string
 }
--- a/src/app/domain/backup/models/Media.ts
+++ b/src/app/domain/backup/models/Media.ts
@@ -30,7 +30,6 @@ export interface Media {
  * @member {number} creationDate timestamp with ms set to 0, e.g. 1694523521000
  * @member {number} modificationDate timestamp with ms set to 0, e.g. 1694172391000
  * @member {string} remoteId e.g. "d78a3c84139d173dde3b87df0003b32e"
- * @member {string} remotePath e.g. "/Sauvegardé depuis mon mobile/My Android/Download"
  */
 export interface BackupedMedia {
   name: string
@@ -38,5 +37,4 @@ export interface BackupedMedia {
   creationDate: number
   modificationDate: number
   remoteId: string
-  remotePath: string
 }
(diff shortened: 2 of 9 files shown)


src/@types/cozy-client.d.ts
```diff
@@ -120,8 +120,12 @@ declare module 'cozy-client' {
...
   }
 
+  interface MissingFileDocumentAttributes {
+    md5sum: string
+  }
+
   type IOCozyFile = {
-    attributes: FileDocument
+    attributes: MissingFileDocumentAttributes & FileDocument
   } & CozyClientDocument
```
zatteo, line 129: I will soon create a PR on cozy-client with every types I added manually here.

src/app/domain/backup/models/Media.ts
```diff
@@ -23,12 +24,18 @@ export interface Media {
...
+ * @member {string} name e.g. "IMG_50.jpg"
+ * @member {string} uri e.g. "file:///storage/emulated/0/Download/IMG_50.jpg" on Android or "ph://0E75CF5E-4587-4BD7-AB07-E940BB627C4B/L0/001" on iOS
+ * @member {number} creationDate e.g. 1694523521000
+ * @member {number} modificationDate e.g. 1694172391000
+ * @member {string} remoteId e.g. "d78a3c84139d173dde3b87df0003b32e"
+ * @member {string} remotePath e.g. "/Sauvegardé depuis mon mobile/My Android/Download"
  */
 export interface BackupedMedia {
   name: string
+  uri: string
```
Crash--, line 37: nit: since we have remoteXXX, maybe we can have localXXX in contrast?

src/app/domain/backup/services/manageLocalBackupConfig.ts
```diff
@@ -116,11 +116,11 @@ export const setMediaAsBackuped = async (
...
       name: media.name,
+      uri: media.uri,
+      creationDate: media.creationDate,
+      modificationDate: media.modificationDate,
+      remoteId: documentCreated.id as string,
       remotePath: media.remotePath
-    } as BackupedMedia
-
-    if (documentCreated.attributes.name !== media.name) {
-      newBackupedMedia.remoteName = documentCreated.attributes.name
```
Crash--, line 123: normal d'avoir retiré le type et ce check dans ce commit?

src/app/domain/backup/services/manageRemoteBackupConfig.ts
```diff
@@ -233,10 +238,25 @@ export const createRemoteBackupFolder = async (
   return remoteBackupConfig
 }
 
+const isFileCorrespondingToMedia = (file: File, media: Media): boolean => {
+  return (
+    file.name === media.name &&
+    new Date(file.created_at).getTime() === media.creationDate &&
+    new Date(file.updated_at).getTime() === media.modificationDate
+  )
+}
```
Crash--, line 246: Is it resilient to time zone changes? aka: I took lots of photos while on vacation in Hawaii, and I don't do backups during that time. Or only a part of it. And then I go back to France. What happens then?

src/app/domain/backup/services/manageRemoteBackupConfig.ts
```diff
@@ -233,10 +238,25 @@ export const createRemoteBackupFolder = async (
...
+
 const formatBackupedMedia = (
+  allMedias: Media[],
   deviceRemoteBackupConfig: RemoteBackupConfig,
   file: File
-): BackupedMedia => {
+): BackupedMedia | undefined => {
+  const correspondingMedia = allMedias.find(media =>
+    isFileCorrespondingToMedia(file, media)
+  )
```
Crash--, line 255: What is the perf cost here? I remember that we already have a lot of loops, we're adding one?

zatteo: My point of view of this for the moment (we are indeed adding a loop) : 
- the loop cost is okay
- the line `const allMedias = await getAllMedias(client)` is less okay
- we should be able to add some cache to this big request to the native side
- it only happens when _restoring a backup_ so I propose to do nothing for the moment

But I am of course open to everything.

src/app/domain/backup/services/getMedias.ts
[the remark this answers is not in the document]
zatteo: I updated the commit description to 
```
We will need to use these functions without the callback in a following commit
to get the list of all available medias when restoring a backup. And we
do not want to display two times the loading of medias when restoring a
backup.
```

src/app/domain/backup/models/Media.ts
zatteo: I hesitated for this wording. But everything without `remote` prefix is local. I will at least update documentation to say it.

src/app/domain/backup/services/manageLocalBackupConfig.ts
zatteo: Hmm j'aurais pu faire un commit séparé.

- J'ai retiré le `as BackupedMedia` à ce moment parce qu'il induisait en erreur en faisant croire que `newBackupedMedia` était toujours correct même s'il manquait des attributs.

- Et j'ai retiré le remoteName de BackupedMedia dans ce même commit, donc plus besoin de le mettre.

src/app/domain/backup/services/manageRemoteBackupConfig.ts
zatteo: Original date comes from react-native-cameraroll which is an UTC timestamp. It is then sent to the stack with toISOString() which is also always on UTC time zone. And getTime return again the UTC timestamp from the stack date. So it should be okay!

Crash--: Great to hear!

src/app/domain/backup/services/manageRemoteBackupConfig.ts
Crash--: I don't know yet how to optimize that, but let's ship that way and once we know that it works, let's see how we can improve it if needed.

Crash-- (approved):

PR #929, opened 2023-09-14
feat: Handle App's icon in white labels

Ldoppea:
Now that we have a dedicated MaBulle app with its dedicated icon, we don't need to handle both Cozy and MaBulle icons versions in the same app

This commit removes the MaBulle alternative icon

Also we don't want the App's icon to change for MesPapiers in the Cozy version anymore. This should happen only on the MaBulle version


src/libs/icon/icon.ts
```diff
@@ -40,15 +38,7 @@ export const changeIcon = async (slug: string): Promise<void> => {
     return
   }
 
-  const changeAllowed = flag('flagship.icon.changeAllowed')
-  const defaultIcon =
-    (flag('flagship.icon.defaultIcon') as unknown as string) || DEFAULT_ICON
-
-  if (!changeAllowed) {
-    return
-  }
```
zatteo, line 43: You can remove these flag from flag list on the wiki.

zatteo (approved):

src/libs/icon/icon.ts
Ldoppea: Done ✅  (i did not removed them but I deprecated them with an explanation message)

PR #932, opened 2023-09-15
fix: Do not set the size of the worker to 0 when hidden

doubleface:
We had a problem with the MGEN clisk which, when we set with and height to 0, caused the website to loop infinitely. Because of this loop, the worker could be killed.

In order to stay as close as possible to the user experience when a clisk is launched, we decided to set the with and height to 100%.

#### Checklist

* [X] Tested on iOS
* [X] Tested on Android


Crash--:
> Because this seems to cause some infinite loops in webview and may kill
the worker in some cases.

Can you be more explicit: 

We had a problem with the MGEN clisk which, when we set with and height to 0, caused the website to loop infinitely. Because of this loop, the worker could be killed. 

In order to stay as close as possible to the user experience when a clisk is launched, we decided to set the with and height to 100%.

Crash-- (approved):

PR #943, opened 2023-09-21
 feat: Read user agent value from react-native-config

Ldoppea:
We want react-native-config to be able to read the user agent value so
we can configure it for HTTP queries in both JS, Android and iOS sides

The user agent value should now be set in `white_label/config.json`
file

Then the `yarn brand:configure:<brand_name>` command will add/replace
the user agent environment variable in `.env` file


Crash-- (approved):
Nice!

PR #950, opened 2023-09-22
Backup renaming

zatteo:
fix(backup): Change how we rename file when conflict
fix(backup): Use only name and creationDate to compare medias

### DIFF
--- a/src/app/domain/backup/services/manageRemoteBackupConfig.ts
+++ b/src/app/domain/backup/services/manageRemoteBackupConfig.ts
@@ -238,9 +238,11 @@ export const createRemoteBackupFolder = async (
 }
 
 const isFileCorrespondingToMedia = (file: File, media: Media): boolean => {
+  const creationDate = new Date(file.created_at)
+  creationDate.setMilliseconds(0)
+
   return (
-    file.name === media.name &&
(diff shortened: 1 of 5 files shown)


src/app/domain/backup/services/manageRemoteBackupConfig.ts
```diff
@@ -238,9 +238,11 @@ export const createRemoteBackupFolder = async (
...
 const isFileCorrespondingToMedia = (file: File, media: Media): boolean => {
+  const creationDate = new Date(file.created_at)
+  creationDate.setMilliseconds(0)
+
   return (
-    file.name === media.name &&
-    new Date(file.created_at).getTime() === media.creationDate
+    file.name === media.name && creationDate.getTime() === media.creationDate
   )
 }
```
Crash--, line 248: @nono are you aware about the need to do that?

Crash-- (approved):

src/app/domain/backup/services/manageRemoteBackupConfig.ts
nono: I think I was a message about a precision issue (with milliseconds in one case, and without in the other case)

PR #951, opened 2023-09-22
Dedup first version

zatteo:
Goal : add a backup which upload only what is NOT already in the whole cozy. This is a first version we want to try on real data/user.

Technically, before all backup :
- fetch all media from Cozy
- compare them with local media with name and creation date
- if media exists on phone AND is not already considered as backup, we considered it as backup

=> avoid to upload media already present it cozy

Behind a flag('flagship.backup.dedup')

### DIFF
--- a/src/app/domain/backup/services/manageRemoteBackupConfig.ts
+++ b/src/app/domain/backup/services/manageRemoteBackupConfig.ts
@@ -247,7 +247,6 @@ const isFileCorrespondingToMedia = (file: File, media: Media): boolean => {
 
 const formatBackupedMedia = (
   allMedias: Media[],
-  deviceRemoteBackupConfig: RemoteBackupConfig,
   file: File
 ): BackupedMedia | undefined => {
   const correspondingMedia = allMedias.find(media =>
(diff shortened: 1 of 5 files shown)


src/app/domain/backup/services/manageRemoteBackupConfig.ts
```diff
@@ -271,20 +271,27 @@ const removeUndefined = (
...
   const allMedias = await getAllMedias(client)
 
-  const filesQuery = buildFilesQuery(deviceId)
+  let data
+
+  if (flag('flagship.backup.dedup')) {
+    const filesQuery = buildAllFilesQuery()
 
-  const data = (await client.queryAll(filesQuery)) as FilesQueryAllResult
+    data = (await client.queryAll(filesQuery)) as FilesQueryAllResult
```
Crash--, line 285: dans l'idée d'éviter trop d'AR réseaux, tu pourrais utiliser une limiteBy(1000). La limite par défaut étant 100. https://github.com/cozy/cozy-stack/blob/master/web/files/files.go#L1498-L1500

Crash-- (approved):

PR #956, opened 2023-09-28
fix: Call showOauthClientsLimitExceeded only when Cookie exists

Ldoppea:
When showing OauthClientsLimitExceeded we should ensure that a Session Cookie is set on the app

If not, the popup will display a cozy-stack error

After the login scenario, when cozy-home does its first render, then no Session Cookie exist yet. In that case we should call `getIndexHtmlForSlug` at least once to generate the Session Cookie

So in that case we want to call `showOauthClientsLimitExceeded` only after `getIndexHtmlForSlug`

But if a Session Cookie exists, then we want to continue doing the check before calling `getIndexHtmlForSlug` so we prevent unnecessary `getIndexHtmlForSlug` calls (this queries cozy-stack) when OauthClientsLimit is exceeded and when the Session Cookie exists

### DIFF
--- a/src/components/webviews/CozyProxyWebView.js
+++ b/src/components/webviews/CozyProxyWebView.js
@@ -1,149 +1,20 @@
-import Minilog from 'cozy-minilog'
 import { useFocusEffect } from '@react-navigation/native'
 import React, { useCallback, useState, useEffect } from 'react'
-import { AppState, Platform, View } from 'react-native'
+import { AppState, View } from 'react-native'
 
 import { useClient } from 'cozy-client'
+import Minilog from 'cozy-minilog'
 
-import { styles } from './CozyProxyWebView.styles'
-import { CozyWebView } from './CozyWebView'
-
-import { checkOauthClientsLimit } from '/app/domain/limits/checkOauthClientsLimit'
-import { showOauthClientsLimitExceeded } from '/app/domain/limits/OauthClientsLimitService'
 import { RemountProgress } from '/app/view/Loading/RemountProgress'
-import { updateCozyAppBundleInBackground } from '/libs/cozyAppBundle/cozyAppBundle'
-import { getCookie } from '/libs/httpserver/httpCookieManager'
+import { initHtmlContent } from '/components/webviews/CozyProxyWebView.functions'
+import { CozyWebView } from '/components/webviews/CozyWebView'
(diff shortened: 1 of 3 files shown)


src/components/webviews/CozyProxyWebView.js
```diff
@@ -81,19 +82,27 @@ const initHtmlContent = async ({
...
-      return
-    }
+  const cookieAlreadyExists = (await getCookie(client)) !== undefined
+  log.debug(`Check cookie already exists: ${cookieAlreadyExists}`)
+
+  if (
+    cookieAlreadyExists &&
+    (await doesOauthClientsLimitPreventsLoading(client, slug, href))
+  ) {
+    log.debug('Stop loading HTML because OAuth client limit is reached (pre)')
```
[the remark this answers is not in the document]
zatteo: Okayyy understood with the next log

src/components/webviews/CozyProxyWebView.js
```diff
@@ -111,6 +120,31 @@ const initHtmlContent = async ({
...
+ * @param {string} href - The WebView requested href
+ * @returns true if the WebView rendering should be prevented, false otherwise
+ */
+const doesOauthClientsLimitPreventsLoading = async (client, slug, href) => {
+  const isOauthClientsLimitExeeded = await checkOauthClientsLimit(client)
+
+  if (isOauthClientsLimitExeeded) {
+    if (slug === 'home') {
+      showOauthClientsLimitExceeded(href)
+      return false
```
zatteo: Why do we want to render the home even if oauth clients limit is exceeded ?

zatteo (approved):

src/components/webviews/CozyProxyWebView.js
Ldoppea: Here the home will be rendered behind the pop so it won't be visible by the user

We do this to improve user experience so when they fix the number of OAuth client, they will instantaneously go to the home (as it is already loaded)

More generally, we try to always have the homeView loaded for this reason (except when no user is logged)

Note that we do this for the HomeView because it has its own container. We cannot do this for cozy-app because one container is used for all cozy-app, so for example we cannot keep drive and mespapiers loaded simultaneously. But if we could we would have done it.

zatteo: Thank you for the explanation!

PR #958, opened 2023-09-28
fix: Log `checkInitialParam` alert using `debug` level instead of `warn`

Ldoppea:
The `warn` level generates too much noise on the debuging console as it outputs the entire stack trace

The 🚨 icon used in the log is enough to catch the attention so we now want to use `debug` log


zatteo:
❤️❤️❤️

Should have done it weeks ago...

PR #959, opened 2023-09-29
feat: Add OCR support for cozy-apps

Ldoppea:
We want cozy-apps to be able to call OCR on a base64 image

The flagship now exposes two new methods through cozy-intent:
- `isOcrAvailable`: Check if the OCR feature is supported by the current app
- `ocr`: Process OCR on the given base64 image and return the OCR result + the image size

Related PR: cozy/cozy-libs#2297

___

TODO:

- [x] Upgrade cozy-intent when cozy/cozy-libs#2297 is merged
- [x] Check that no request is made by the ocr library to the Google servers


zatteo:
Looks good

zatteo (approved):

PR #936, opened 2023-09-18
feat: Merge JSON files instead of replacing them in white labels and add MaBulle versions for Welcome screen and app name

Ldoppea:
Some JSON files that are impacted by white labels are too big so it is difficult to visualize the difference between them when comparing two brands

So instead of having full JSON files that will override targets, we want to declare subset of those files with just the keys/values that should override the target values

This PR also implement changes to Welcome screen and app name for MaBulle version

PR #962, opened 2023-10-03
feat: Set missing translations for Ma Bulle

Ldoppea:
In #936 we added some custom texts for french version of Ma Bulle that were not declared on english and spanish versions

When this is the case, then the Cozy version of those texts is used

Here we want to have Ma Bulle specific texts for those languages

> **Warning**
> Those are auto-translated versions. I checked the English result, but I can't check the spanish version as I don't speak spanish

### DIFF
--- a/white_label/brands/mabulle/js/locales/extra-es.json
+++ b/white_label/brands/mabulle/js/locales/extra-es.json
@@ -1,5 +1,23 @@
 {
+  "logout_dialog": {
+    "content": "¿Estás seguro de que quieres desconectar esta aplicación de tu cuenta?"
+  },
   "screens": {
+    "cozyBlocked": {
+      "title": "Su cuenta ha sido suspendido"
+    },
+    "cozyNotFound": {
+      "body": "Esta dirección no parece existir. Puede tratarse de un error tipográfico. Para averiguarlo, compruebe la dirección en el correo electrónico que recibió cuando se creó.",
+      "title": "Esta dirección no existe."
+    },
+    "SecureScreen": {
+      "passwordprompt_body": "Como el acceso a tu dispositivo no es seguro, elige una contraseña para asegurar el acceso a tu cuenta.",
+      "passwordprompt_title": "Asegurar este acceso",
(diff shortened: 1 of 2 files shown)


white_label/brands/mabulle/js/locales/extra-es.json
```diff
@@ -1,5 +1,23 @@
 {
+  "logout_dialog": {
+    "content": "¿Estás seguro de que quieres desconectar esta aplicación de tu espacio personal?"
```
KillianCourvoisier, line 3: Regarding the wording of cozy, the use of "espacio personal" make sens, just a little nit :  usually we would use "su/tu cuenta", so it could be discuss, as you wish 👍

white_label/brands/mabulle/js/locales/extra-es.json
```diff
@@ -1,5 +1,23 @@
...
+      "title": "Su espacio personal ha sido suspendido"
+    },
+    "cozyNotFound": {
+      "body": "Esta dirección no parece existir. Puede tratarse de un error tipográfico. Para averiguarlo, compruebe la dirección en el correo electrónico que recibió cuando se creó.",
+      "title": "Esta dirección no existe."
+    },
+    "SecureScreen": {
+      "passwordprompt_body": "Como el acceso a tu dispositivo no es seguro, elige una contraseña para asegurar el acceso a tu espacio personal.",
+      "passwordprompt_title": "Asegurar este acceso",
+      "pinprompt_body": "Acceda más rápidamente a su aplicación activando el desbloqueo mediante código PIN"
```
KillianCourvoisier, line 16: Here instead of "mediante", we can use "con el", it may be more commonly used

white_label/brands/mabulle/js/locales/extra-es.json
```diff
@@ -1,5 +1,23 @@
...
+    },
+    "login": {
+      "invalidMagicCode": "El enlace a su espacio personal está truncado o ha caducado"
```
KillianCourvoisier, line 19: Replace "está truncado" with  "está cortado", it seems more accurate

PR #964, opened 2023-10-05
feat: Use ensureKonnectorFolder from cozy-client

doubleface:
And also update the destination folder id saved in the trigger when
needed.

Upgrade cozy-client to 41.4.1 to be able to update the trigger and to
have ensureKonnectorFolder function.

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android
* [X] Localized in English and French


src/libs/Launcher.js
```diff
@@ -271,9 +270,10 @@ export default class Launcher {
     launcherClient.setAppMetadata({
       sourceAccount: account._id
     })
-    const folder = await ensureKonnectorFolder(client, {
-      konnector,
-      account
+    const folder = await models.konnectorFolder.ensureKonnectorFolder(client, {
+      konnector: { ...konnector, _id: konnector._id },
```
Crash--, line 274: Why do you spread?

src/libs/Launcher.js
```diff
@@ -289,6 +289,27 @@ export default class Launcher {
       trigger = triggerResponse.data
       result.createdTrigger = trigger
       log.debug(`ensureAccountAndTriggerAndJob: created trigger`, trigger)
+      // @ts-ignore
```
Crash--, line 292: can you add a comment to explain why you need to ignore 🙏

src/libs/Launcher.js
doubleface: _id attribute is missing in konnector object, which causes the reference to the konnector in the destination folder to be null

I think there is something to fix in harvest when the launcher context is sent

src/libs/Launcher.js
doubleface: Done, the message attribute of IOCozyTrigger is kind of a free object. Its content depends on the kind of trigger. But all the konnector trigger (client or node) which have a destination folder will have the `.message.folder_to_save` attribute.

src/libs/Launcher.js
Crash--: I don't understand: 

You do: 
```
      konnector: { ...konnector, _id: konnector._id }
```

and you say : "_id attribute is missing in konnector object,". But you do: `_id: konnector._id` so _id is present in the konnector object. If this is not the case, you'll have `_id: undefined` is it really wanted?

src/libs/Launcher.js
Crash--: So let's update the IOCozyTrigger type on Cozy-client 🙏

src/libs/Launcher.js
doubleface: Sorry, I miss typed. The right code is `konnector: { ...konnector, _id: konnector.id }` and I fixed it now.

doubleface:
@Crash-- Should be updated now

Crash-- (approved):

doubleface:
Upgraded cozy-client with the updated triggers types and removed some @ts-ignore

PR #971, opened 2023-10-10
feat: implement app selection in file upload

acezard:
image

image


src/app/domain/osReceive/index.ts
```diff
@@ -1,3 +1,15 @@
 import Minilog from 'cozy-minilog'
 
+import { OsReceiveState } from './models/OsReceiveState'
+
 export const OsReceiveLogger = Minilog('🗃️ OsReceiveService')
+
+export const removeSourceProperty = (
```
Ldoppea, line 7: nit: Maybe we should move this into a logHelper file or similar? (just to give an it that this is not used for business logic)

src/locales/es.json
```diff
@@ -98,6 +98,15 @@
...
       }
+    },
+    "osReceive": {
+      "documentType": "Tipo de documento :",
+      "submit": "Siguiente",
+      "abort": "Regresar",
+      "nElementsTitle": "{{n}} elementos",
+      "errors": {
+        "uploadFailed": "Error al subir {{filename}}"
+      }
```
Ldoppea, line 109: Note to myself, I forgot to add candidate app translations here

src/App.js
```diff
@@ -89,13 +91,15 @@ const InnerNav = ({ client, setClient }) => {
...
   return (
     <NativeIntentProvider
-      localMethods={localMethods(
-        client,
-        OsReceiveApi(client, osReceiveState, osReceiveDispatch)
-      )}
+      localMethods={localMethods(client, {
+        ...OsReceiveApi(client, osReceiveState, osReceiveDispatch),
+        hideSplashScreen: () =>
+          hideSplashScreen(undefined, filesToUpload.length > 0)
```
Ldoppea, line 101: Since this override the one form `/libs/intents/localMethods`, should we remove the `hideSplashScreen` it?

src/app/domain/osReceive/index.ts
acezard: I renamed it to be more clear about its intent. For the file I'm not sure, I think it's a good idea to at least keep it with "osreceivelogger"

src/App.js
acezard: I'm not sure, I think we should keep both to avoid potential race conditions. The first one would be a fallback if this one didn't override it yet.
I would need to test a bit more to be sure

Ldoppea: I'm not sure this scenario is possible as we inject this directly in the root component (app.js). But i'm ok with that, this should not block the approval

PR #972, opened 2023-10-10
feat: Separate the worker webview from the navigation

doubleface:
When the worker webview was displayed to fill the login form, the
navigation buttons were displayed above the webview and the some
elements of the login form were not accessible in some cases.

Now the separation is done and the bottom is displayed in white with
dark navigation buttons.

image

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/screens/konnectors/LauncherView.js
+++ b/src/screens/konnectors/LauncherView.js
@@ -32,7 +32,7 @@ import { deactivateKeepAwake } from '/app/domain/sleep/services/sleep'
 const log = Minilog('LauncherView')
 
 const colors = getColors()
-const { statusBarHeight } = getDimensions()
+const { statusBarHeight, navbarHeight } = getDimensions()
 const HEADER_PADDING_TOP = statusBarHeight + 8
 const HEADER_PADDING_BOTTOM = 8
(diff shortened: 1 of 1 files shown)


src/screens/konnectors/LauncherView.js
```diff
@@ -242,6 +242,16 @@ export class LauncherView extends Component {
...
 
+    if (workerVisible) {
+      setFlagshipUI({
+        bottomTheme: 'dark'
+      })
+    } else {
+      setFlagshipUI({
+        bottomTheme: 'light'
+      })
+    }
```
Ldoppea, line 253: Not sure if this is enough, the HomeView can have a white background and so in that cas we should reapply a `dark` bottom theme

@acezard should we use a callerName or call resetUiState here?

Crash--: @acezard ?

acezard: @Crash-- @Ldoppea in case of doubt, yes `resetUIState` is the best way to handle that since it's linked to the home theme (normal/inverted)

doubleface:
Another PR will be created which handle bottom bar style better

PR #973, opened 2023-10-10
feat: Fetch and save debug data on konnector errors

doubleface:
This is a draft. Probably the way to package the trace is not the best, but the data is very useful. Please comment this.

This will save some useful debug data :

- current url on worker
- current stringified dom on worker
- all the console.log in pilot and worker

In a trace file in the form : `trace_<date>_<slug>_<accountId>.html`

Under the following conditions :

- if the debug.clisk.html-on-error flag is set to true
- if the konnector throws an error

This way, the user has the possibility to inspect the data sent to the
cozy team before sharing it with a member of the cozy team.

Here is the resulting file format :

```html
<!-- https://mysite.com/currenturl/ -->
<html>
...current worker page html
</html>
<!--
console.logs : one line per log
-->


#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -16,6 +16,7 @@ import { updateCozyAppBundle } from '/libs/cozyAppBundle/cozyAppBundle'
 import { sendKonnectorsLogs } from '/libs/konnectors/sendKonnectorsLogs'
 
 import { wrapTimerFactory } from 'cozy-clisk'
+import flag from 'cozy-flags'
 
 import {
   activateKeepAwake,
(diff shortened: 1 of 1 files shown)


Crash--:
I think it's a smart way to do that. Maybe we can just create a `txt` file insteaf of an html. Like that, the user will be able to display the content within Drive for instance without having to download the file (and I think the html extension is not mandatory, right?)

doubleface:
Right, html is not mandatory. It's just that it's actually a html file and this way, when I fetch the file, I can directly display it in firefox. 

But in my android phone, the file is not downloaded, it is directly displayed like a text file : 
image

Crash--:
Great!

PR #988, opened 2023-10-20
Add 'ocr', 'scanner' and 'backup' features to isAvailable

zatteo:
**refactor: Support 'ocr' feature in isAvailable**

We decided to move from isFeatureAvailable() to isAvailable(featureName: string) to check if a feature is available.

isOcrAvailable has not been released yet. So let's move it now before having to handle a breaking change.

 **refactor: Support 'scanner' and 'backup' feature in isAvailable**

We decided to move from isFeatureAvailable() to
isAvailable(featureName: string) to check if a feature is available.

These feature have already been released in production. So I add the
possibility here to check if they are present by calling
isAvailable('scanner') or isAvailable('backup') but I do not remove
isScannerAvailable method and backup_available injected variable.

We will need to wait that the new isAvailable is released and
distributed to all of our users before removing the deprecated
methods.

### DIFF
--- a/src/libs/intents/localMethods.ts
+++ b/src/libs/intents/localMethods.ts
@@ -83,6 +83,10 @@ const isAvailable = (featureName: string): Promise<boolean> => {
     return Promise.resolve(true)
   } else if (featureName === 'ocr') {
     return Promise.resolve(isOcrAvailable())
+  } else if (featureName === 'backup') {
+    return Promise.resolve(true)
+  } else if (featureName === 'scanner') {
+    return Promise.resolve(isScannerAvailable())
   }
 
   return Promise.resolve(false)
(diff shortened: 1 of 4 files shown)


src/libs/intents/localMethods.ts
```diff
@@ -251,9 +257,8 @@ export const localMethods = (
     openAppOSSettings,
     isNativePassInstalledOnDevice,
     scanDocument,
-    isScannerAvailable: () => Promise.resolve(isScannerAvailable()),
+    isScannerAvailable: () => Promise.resolve(isScannerAvailable()), // deprecated
```
Ldoppea, line 260: maybe we can console log in this method to alert about deprectated method?

Ldoppea (approved):

PR #989, opened 2023-10-23
fix: Make cozy-apps responsive to keyboard on Android

Ldoppea:
On Android and on small screen, some cozy-app fields may be hidden behind the keyboard when it is open

This can happen on Harvest unlock form, in MesPapiers when adding some documents, or in cozy-notes when editing a text on the bottom of a note

To prevent this we want to add a `KeyboardAvoidingView` around cozy-app WebViews like it is already done for login, onboarding and lock screens

Also, in the addition of `KeyboardAvoidingView`, we need to remove the `justifyContent: 'center'` on the HomeScreen. With this value the HomeView would stay center in the screen referential when the keyboard is open, so the HomeView would become partially hidden by the keyboard and would have a large padding on its top

Removing `justifyContent: 'center'` on the HomeScreen seems to have no side effect. This value is here since the beginning of the project and it seems not to be useful anymore

# Before/after on Harvest unlock form
Screenshot_1698065209b Screenshot_1698065164b

# Before/after on cozy-notes
Screenshot_1698065450b Screenshot_1698065431b

> **Warning**
> on cozy-notes, the bottom tool panel is still hidden. This seems to be an issue on the cozy-notes side. So this should be fixed in another PR

# Before/after on MesPapiers
Screenshot_1698065325b Screenshot_1698065364b

### DIFF
--- a/src/components/webviews/CozyProxyWebView.js
+++ b/src/components/webviews/CozyProxyWebView.js
@@ -1,6 +1,6 @@
 import { useFocusEffect } from '@react-navigation/native'
 import React, { useCallback, useState, useEffect } from 'react'
-import { AppState, View } from 'react-native'
+import { AppState, KeyboardAvoidingView, Platform, View } from 'react-native'
 
 import { useClient } from 'cozy-client'
 import Minilog from 'cozy-minilog'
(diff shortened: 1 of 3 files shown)


Ldoppea:
TODO

- [ ] Check with CLISK team that the removal of `justifyContent: 'center'` has no side effect on CLISK interface

zatteo (approved):

zatteo:
Not important: Do you know why it is an Android only issue?

Ldoppea:
> Not important, just asking: Do you know why it is an Android only issue?

I have no precise explanation, this is due on how the OS browser handles the keyboard place in the UI view

PR #991, opened 2023-10-25
[VIO-104] feat: Implement IAP

zatteo:
This PR adds support for IAP by:
- Implementing the ClouderyOffer view that is responsible for displaying the Cloudery webpage with premium plans description
- Opening this ClouderyOffer view when the OAuthClientLimitExceeded popup's corresponding button is clicked
- Opening this ClouderyOffer view when any cozy-app redirects to its url (not tested yet, this will be done soon)
- Intercepting premium plans redirection and trigger IAP process
  - Providing productID and cozy's UUID in the IAP process so the Cloudery can link the IAP Receipt to the corresponding Cozy
- Adding rudimentary overlay over the ClouderyOffer view during the IAP process (this will be replaced by a designed loading spinner)

> **Warning**
> This PR contains a temporary commit to make the IAP button visible in the OAuthClientLimitExceeded popup for testing purpose. The final behavior will be implemented in the following days


Ldoppea:
@zatteo @Crash-- I cleaned all current-progress commits and I added a description to the PR.

Current progress can be reviewed and I think it can be merged (i still have to fix the CI error).

The following steps will consist to:
- Implement the "loading" UI that is displayed in the app during the IAP process (iOS sometimes shows the app between IAP steps)
- Add cozy's domain next to the UUID on the IAP params (this will be needed by @taratatach for the Android's part)
- Handle the refresh notification
- Some UI refinements (i.e. status bar and navigation bar colors)

They may be done in another PR

src/app/view/IAP/ClouderyOffer.tsx
```diff
@@ -35,18 +52,33 @@ const ClouderyOfferWithIAPContext = (): JSX.Element | null => {
...
+          activeOpacity={0.5}
+          onPress={hidePopup}
+          style={styles.headerTouchableStyle}
+        >
+          <BackTo color={colors.primaryColor} width={16} height={16} />
+          <Text style={styles.headerTextStyle}>
+            {t('screens.clouderyOffer.backButton')}
+          </Text>
+        </TouchableOpacity>
+      </View>
```
zatteo, line 82: nit: Could be taken out as a separate component to avoid having lot of style stuff in ClouderyOffer.tsx

src/app/domain/iap/services/clouderyOffer.ts
```diff
@@ -36,18 +43,34 @@ export const isClouderyOfferUrl = (url: string): boolean => {
...
+    subscribed: () => void
+  ) =>
   (request: WebViewNavigation): boolean => {
+    log.debug('Navigating to', request.url)
     try {
       if (isStartIapUrl(request.url)) {
         const url = new URL(request.url)
-        const productId = url.searchParams.get('productId')
-        log.debug('Should start IAP with productId', productId)
+        const productId = url.searchParams.get('productId') ?? 'test_comfort'
```
zatteo, line 69: 'test_comfort' to remove?

src/app/domain/iap/services/clouderyOffer.ts
```diff
@@ -60,3 +83,104 @@ export const interceptNavigation =
...
+    }
+
+    subscribed()
+  } catch (error: unknown) {
+    if (
+      typeof error === 'object' &&
+      error !== null &&
+      'code' in error &&
+      error.code === 'E_USER_CANCELLED'
+    ) {
```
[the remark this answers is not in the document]
Ldoppea: Fixed in https://github.com/cozy/cozy-flagship-app/compare/f34e698f2c17e07fc5a685a6a30e94e33fe13056..a98b0b1ac9326b3f2adbe22b18aed721e83c3456

zatteo:
J'approuve

src/app/domain/iap/services/clouderyOffer.spec.ts
Crash--, line 1: Can you explain the reasoning behind this event emiter implementation?

src/app/view/IAP/hooks/useClouderyOffer.ts
```diff
@@ -12,15 +14,28 @@ import {
 
 const log = Minilog('💳 Cloudery Offer')
 
+const IOS_OFFERS = ['test_comfort_monthly', 'test_super_comfort_monthly']
+const ANDROID_OFFERS = ['test_comfort', 'test_super_comfort']
```
Crash--, line 18: Is it mandatory to store the offers within the flagship app? (also `test_` seems weird)

src/locales/fr.json
```diff
@@ -25,6 +25,9 @@
     "title": "Déconnexion"
   },
   "screens": {
+    "clouderyOffer": {
+      "backButton": "RETOUR À MON COZY"
```
Crash--, line 29: OUAIS VIVE LES MAJ!

src/app/domain/iap/services/clouderyOffer.spec.ts
Ldoppea: This is the same one we implemented for some other dialogs (ex: the one for [OauthClientsLimitService](https://github.com/cozy/cozy-flagship-app/blob/790e90c1fd04150ea168978b9b42472a59d12d38/src/app/domain/limits/OauthClientsLimitService.ts#L25))

The goal is that from anywhere in the code we can trigger this event to display the corresponding dialog.

Here we have two places that can trigger the ClouderyOffer dialog: the OauthClientsLimitService's navigation interception (in `onShouldStartLoadWithRequest`) and the ReloadInterceptorWebView's one

The `ClouderyOffer` dialog is declared somewhere in the app, and listen this event in order to chose when it should be displayed.

Another approach would be to create a root Context, but there are already a lot of root provider and I find this method easier to handle (one time event, no global rerender, no risk to forget declaring the context or to call the openDialog method outside of the context)

src/app/view/IAP/hooks/useClouderyOffer.ts
Ldoppea: Yes this is mandatory. The app choose which offer to handle or not. I think this has been chosen by Android and iOS stores in order to allow creating new offers that targets only most recent app versions. The app won't display a new offer until we explicitly declare it.

However we can put it in the stack, but this would re-introduce the problem I thing the stores tried to fix.

Ldoppea: I forgot to answer the `test_` part. This is just the offers we use for tests. For now the feature is behind a flag so this shouldn't be problematic.

But that being said, maybe we should avoid to publish a version with those values so we won't have to check the version when adding the flag to everyone.

I'll check with @anthonycozy if we can create final values here

src/app/domain/iap/services/clouderyOffer.ts
```diff
@@ -0,0 +1,31 @@
+import { EventEmitter } from 'events'
+
+import { navigate } from '/libs/RootNavigation'
+
+import { Platform } from 'react-native'
+
+export const CLOUDERY_OFFER = 'CLOUDERY_OFFER'
+
+const START_IAP_REDIRECT_URL = 'https://iapflagship'
```
acezard: Those strings look more like config than code, maybe we should put them somewhere else

src/app/view/IAP/hooks/useClouderyOffer.ts
```diff
@@ -29,7 +44,13 @@ export const useClouderyOffer = (): ClouderyOfferState => {
...
       }
 
-      setPopupUrl(current => current ?? href)
+      const doAsync = async (): Promise<void> => {
+        await initConnection()
+        await getSubscriptions({ skus: SKUS })
+        setPopupUrl(current => current ?? href)
+      }
+
+      void doAsync()
```
acezard, line 56: seems a bit risky without try catch here, it isn't clear if the imported functions will handle errors gracefully

src/app/view/IAP/ClouderyOffer.tsx
```diff
@@ -68,6 +88,29 @@ const WebViewWithLoadingOverlay = ({
...
+        <BackTo color={colors.primaryColor} width={16} height={16} />
+        <Text style={styles.headerTextStyle}>
+          {t('screens.clouderyOffer.backButton')}
+        </Text>
+      </TouchableOpacity>
+    </View>
+  )
+}
+
 const styles = StyleSheet.create({
```
acezard, line 115: starting to have a lot of styles LoC in that file maybe we can extract it. Also it seems some of the value are magic numbers at first glance

acezard (approved):
Left some comments but they are non blocking 👍

src/app/view/IAP/hooks/useClouderyOffer.ts
Ldoppea: I put final offers here: https://github.com/cozy/cozy-flagship-app/pull/991/commits/b4b31984bdfaa5a0fa14b8720afc2b5d72a9b756#diff-de03f5378e032b3c91fdfaf88f0ad11e3569fe199c7d3db80d5f235c0fe1b904R17-R21

src/app/view/IAP/hooks/useClouderyOffer.ts
Ldoppea: This has been fixed in this other PR: https://github.com/cozy/cozy-flagship-app/pull/1131

PR #993, opened 2023-10-26
feat: Use activity detection handler for motion changes

paultranvan:
We used to rely on the activity stored in locations points to send
motion_activity to the server. However, we noticed the activity is not
always specified in the points, which was causing bad section
segmentation.

Now, we rely on the activity change API: each time an activity is
detected, we store it locally. Then, at upload time, we extract the
relevant activities (i.e. timestamp lower than the last location point,
and avoid activity duplicates), and remove it afterwards.


src/app/domain/geolocation/tracking/storage.js
```diff
@@ -0,0 +1,143 @@
...
+    const activitiesBeforeTs = activities.filter(activity => {
+      return activity?.data?.ts <= beforeTs
+    })
+    return activitiesBeforeTs
+  }
+  return activities
+}
+
+export const removeActivities = async ({ beforeTs } = {}) => {
+  console.log('remove activities')
```
zatteo: If you want to keep it, use the `Log` which uses cozy-minilog logger

src/app/domain/geolocation/tracking/index.js
```diff
@@ -120,6 +121,11 @@ export const setTrackingConfig = async newTrackingConfig => {
   await storeData(StorageKeys.GeolocationTrackingConfig, newTrackingConfig)
 }
 
+export const handleActivityChange = async event => {
+  Log('[ACTIVITY CHANGE] - ' + JSON.stringify(event))
```
zatteo, line 125: These log is duplicated with the one from the headless task no ?

`Log('[ACTIVITYCHANGE] -' + JSON.stringify(params))`

src/app/domain/geolocation/tracking/storage.js
paultranvan: Yep sorry, the rebase was not perfect

src/app/domain/geolocation/tracking/index.js
paultranvan: Good catch :+1:

src/app/domain/geolocation/tracking/tracking.js
```diff
@@ -247,14 +271,21 @@ const addPoint = (content, point, filtered) => {
...
-    content.push(motionActivity)
+const getActivityMode = activity => {
+  const modeKeys = [
+    'cycling',
+    'running',
+    'walking',
+    'automotive',
+    'stationary',
+    'unknown'
+  ]
```
zatteo, line 282: You can move this constant outside the function.

zatteo:
Nice refacto with the storage 👍

zatteo (approved):

src/app/domain/geolocation/tracking/tracking.js
paultranvan: Fair enough, I move it

PR #997, opened 2023-10-31
feat: Enable CSS vars in disableGetIndex mode

acezard:
When disabling the HTTP server in local development
 environment (see `dev-config.ts`),
The CSS variables `--flagship-top-height`
and `--flagship-bottom-height` are not set
in the WebView. To counter this, we inject them manually at runtime.
This code block is never intended to be executed in production.

Still, we wrap it in a try/catch bloc to avoid breaking the WebView
in case of error.


src/libs/intents/localMethods.ts
```diff
@@ -193,6 +193,7 @@ interface CustomMethods {
   setGeolocationTrackingId: typeof setGeolocationTrackingId
   forceUploadGeolocationTrackingData: typeof forceUploadGeolocationTrackingData
   getDeviceInfo: typeof getDeviceInfo
+  isAvailable: typeof isAvailable
```
zatteo, line 196: It should not be needed to add this because isAvailable is typed in cozy-intent. Are you up to date with master in this branch?

acezard: > It should not be needed to add this because isAvailable is typed in cozy-intent. Are you up to date with master in this branch?

that was probably the issue, I'm going to verify

src/components/webviews/jsInteractions/jsCozyInjection.js
```diff
@@ -1,39 +0,0 @@
...
-  window.cozy.isFlagshipApp = true
-  window.cozy.ClientKonnectorLauncher = 'react-native'
-  window.cozy.ClientConnectorLauncher = 'react-native' // deprecated
-  window.cozy.flagship = ${makeMetadata(routeName)}
-  window.cozy.isSecureProtocol = ${isSecureProtocol || 'false'}
-  // We have random issue on iOS when the app's script is executed
-  // before the view port. To fix that, we wait the load Event
-  // and then we dispatch the resize even to wake up
-  // cozy-ui's breakpoint.
-  // for instance https://stackoverflow.com/questions/5508455/mobile-safari-window-reports-980px/35987682#35987682
```
Ldoppea, line 35: This link disapeared in the new file, is that wanted?

Ldoppea (approved):

PR #999, opened 2023-11-06
fix: Add check on language files and add missing translations for en and es files

Ldoppea:
Some languages files were missing translations and also extra languages were not in sync with each other in white labels

Those new checks should ensure that all langages files contain the same keys and that all extra language override the same keys

### DIFF
--- a/white_label/configure-brand.ts
+++ b/white_label/configure-brand.ts
@@ -7,6 +7,7 @@ import fs from 'fs-extra'
 import Minilog from 'cozy-minilog'
 
 import configs from './config.json'
+import { diffJsonStructure } from '../src/utils/jsonUtils'
 
 export const logger = Minilog('Configure Brand')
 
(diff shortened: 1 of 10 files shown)


Ldoppea:
@KillianCourvoisier can you review the new Spanish translations?

white_label/configure-brand.ts
```diff
@@ -25,6 +26,20 @@ export const checkCozyBrandIso = async (): Promise<boolean> => {
...
 }
 
+export const checkLanguages = (): boolean => {
+  const areEnAndFrEquivalent = areLanguagesEquivalent('en', 'es')
+  const areEnAndEsEquivalent = areLanguagesEquivalent('en', 'fr')
+  const areLangagesOk = areEnAndFrEquivalent && areEnAndEsEquivalent
+
+  const areExtraFrEquivalent = areExtraLanguagesEquivalent('en')
+  const areExtraEnEquivalent = areExtraLanguagesEquivalent('en')
+  const areExtraEsEquivalent = areExtraLanguagesEquivalent('fr')
```
zatteo, line 36: `en en fr` where as it should be `fr en es`? Or I did not understand?

zatteo:
```
const extraLanguages = ['fr', 'en', 'es']
const areExtraLangagesOk = extraLanguages.every(extraLanguage => areExtraLanguagesEquivalent(extraLanguage))
```

seems clearer ?

Ldoppea: Nice catch, i fixed it [here](https://github.com/cozy/cozy-flagship-app/compare/1bd78f72075b6bad4eb44961893d75f51de6a096..2bd5d7d0f4158d8ca4c3dc10ef1ca8a2e62e5b38)

Some considerations:
- I applied your suggestion also for `areLanguagesEquivalent` calls
- This implies that there will be a useless call to `areLanguagesEquivalent('en', 'en')`, comparing english to itself
  - I consider this not a problem as the result will be always be `true` and this is fast enough to not be perceived. Is is acceptable considering the improvement in code readibility
- I could have put `areLanguagesEquivalent` and `areExtraLanguagesEquivalent` in the same loop, but I prefer it this way so the console prints will be grouped by file type (languages, then extra-languages)

zatteo (approved):

KillianCourvoisier (approved):
LGTM 👍

PR #1004, opened 2023-11-08
feat: Remove not useful anymore rn-async-storage-flipper

zatteo:
2 plugins were installed to debug AsyncStorage with flipper : rn-async-storage-flipper to view and
rn-flipper-async-storage-advanced to edit.

When updating to rn-flipper-async-storage-advanced to 1.0.8 (see https://github.com/lbaldy/flipper-plugin-async-storage-advanced/pull/9), we do not need anymore both plugins.

You need to update rn-flipper-async-storage-advanced on Flipper side.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,6 @@
     "react-scripts": "4.0.3",
     "redux-logger": "3.0.6",
     "redux-persist": "^6.0.0",
-    "rn-async-storage-flipper": "^0.0.10",
     "rn-flipper-async-storage-advanced": "^1.0.4",
     "semver": "^7.3.2"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -103,7 +103,6 @@
     "react-scripts": "4.0.3",
     "redux-logger": "3.0.6",
     "redux-persist": "^6.0.0",
-    "rn-async-storage-flipper": "^0.0.10",
     "rn-flipper-async-storage-advanced": "^1.0.4",
```
Crash--, line 106: No need to upgrade this plugin?!

zatteo: No because just the front end part changed!

Crash--: So how do we fetch the front-end part? Just by upgrading the module in flipper?

Crash-- (approved):

package.json
zatteo: Yes. I do not know where to put it. I think I will just send a message to every flagship-app developers to update their front end in flipper, and new developers will automatically get the last one.

PR #1005, opened 2023-11-08
Improve translations for geolocation

zatteo:


src/locales/en.json
```diff
@@ -112,7 +112,9 @@
       }
     },
     "geolocationTracking": {
-      "androidBackgroundPermissionMessage": "Recording your movements requires to ≪ {backgroundPermissionOptionLabel} ≫"
+      "androidBackgroundPermissionMessage": "Recording your movements requires to ≪ {backgroundPermissionOptionLabel} ≫",
+      "androidNotificationTitle": "Location service on",
+      "androidNotificationDescription": "Your journey will be recorded into your Cozy"
```
Ldoppea, line 117: No ES version?

Also this mention "Cozy", so I suppose we should get an alternative text in extra-language files

Ldoppea (approved):

PR #1006, opened 2023-11-08
Improve trip segmentation

paultranvan:
- Change activities filtering strategy: we now exclude consecutive stationary activities, and when their confidence is low.
- Code refactoring to simplify the processing before upload and avoid edge cases
- Exclude network locations occuring after last stationary event, that was extending trip duration


zatteo:
Can't wait to try the algorithm updates ! 🎉 :tada:

zatteo (approved):

PR #1008, opened 2023-11-09
feat: Support xcode 15

zatteo:
I encountered two problems when updating to macos 14 with xcode 15.

- No template named ‘unary_function’ in namespace ‘std’; did you mean ‘__unary_function’?
- Cycle inside App; building could produce unreliable results.

I followed steps described here :
https://toyboy2.medium.com/xcode-15-react-native-troubleshooting-guides-e559ac246bae


ios/CozyReactNative.xcodeproj/project.pbxproj
```diff
@@ -1010,7 +1010,7 @@
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 i386";
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = i386;
```
Crash--, line 1013: We have issue with arm compilation. We excluded specifically https://github.com/cozy/cozy-flagship-app/pull/1008/files#diff-281ded35b124f5160d0a57e47e521ee6f49f233933e091a0d33f3fd42a74abc8L61 

Weird, no?

zatteo: I do not understand why pod install always remove this exclusion, even without this commit. Still trying to find why for the moment.

zatteo: After one hour of reseach, I still have no answer why my `pod install` remove this exclusion. So I removed the issue manually from the .pbxproj.

For information, I also saw that it may not be the best solution
https://stackoverflow.com/questions/63607158/xcode-building-for-ios-simulator-but-linking-in-an-object-file-built-for-ios-f/66233165#66233165

Ldoppea (approved):
It still works on XCode 14 👍

PR #1010, opened 2023-11-10
Refactor multiple old files to TypeScript

acezard:
Key Changes:
1. **TypeScript Integration**:
   - Implemented typings across numerous previously untyped files (`8b112609`), ensuring a more robust and error-resistant codebase.
   - Added typings to most of the functions (`50b5ed73`), providing clearer interfaces and improving developer experience.

2. **File Renaming and Reorganization**:
   - Renamed several files to `.ts` (`8ef5488d`, `a2ff6330`), aligning with TypeScript conventions and improving code consistency.
   - Converted relevant `.ts` files to `.tsx` where JSX is present (`fbe4db2a`), ensuring proper file type representation and IDE support.

3. **Cleanup and Refinement**:
   - Removed redundant '.js' imports (`63a75e73`), streamlining the code and reducing confusion around module systems.


src/app/domain/geolocation/helpers/index.ts
```diff
@@ -4,26 +4,26 @@ import Minilog from 'cozy-minilog'
...
+export const getTs = (location: { timestamp: string }): number => {
   return parseISOString(location.timestamp).getTime() / 1000
 }
 
-export const parseISOString = ISOString => {
-  let b = ISOString.split(/\D+/)
+export const parseISOString = (ISOString: string): Date => {
+  const b = ISOString.split(/\D+/).map(Number)
   return new Date(Date.UTC(b[0], --b[1], b[2], b[3], b[4], b[5], b[6]))
 }
```
zatteo, line 14: @paultranvan is it okay for you?

paultranvan: I'm not sure what it means for [CozyGPSMemory](https://github.com/cozy/CozyGPSMemory): will it require some extra ts config? 
I'm a bit worried that it may introduce more harm than good

acezard: > I'm not sure what it means for [CozyGPSMemory](https://github.com/cozy/CozyGPSMemory): will it require some extra ts config? I'm a bit worried that it may introduce more harm than good

I don't think it should impact CozyGPSMemory, I think the issue here could be an unexpected behaviour at runtime with the casting to Number explicitly (this was required for stronger type inference).

paultranvan: > I think the issue here could be an unexpected behaviour at runtime with the casting to Number explicitly

I'm not sure what you imply here: do you mean an issue could arise if the variable is actually not a number (which is normal), or that this issue could arise anyway?

src/libs/functions/getHostname.ts
```diff
@@ -1,7 +1,22 @@
...
+export const getHostname = (
+  nativeEvent?: { url?: string | unknown } | unknown
+): string | undefined => {
+  if (
+    !nativeEvent ||
+    typeof nativeEvent !== 'object' ||
+    !('url' in nativeEvent) ||
+    typeof nativeEvent.url !== 'string'
+  )
+    return
```
Ldoppea, line 14: Shouldn't we throw instead? Can this happen?

src/components/webviews/jsInteractions/jsCozyInjection.ts
```diff
@@ -21,13 +21,16 @@ const makeMetadata = routeName => {
...
+  routeName?: string,
+  isSecureProtocol?: boolean
+): string => `
   if (!window.cozy) window.cozy = {}
   window.cozy.isFlagshipApp = true
   window.cozy.ClientKonnectorLauncher = 'react-native'
   window.cozy.ClientConnectorLauncher = 'react-native' // deprecated
-  window.cozy.flagship = ${makeMetadata(routeName)}
-  window.cozy.isSecureProtocol = ${isSecureProtocol || 'false'}
+  window.cozy.flagship = ${routeName ? makeMetadata(routeName) : ''}
```
Ldoppea, line 32: Maybe whe should instead
```
  window.cozy.flagship = ${makeMetadata(routeName ?? '')}
```

Seems like having an empty `flagship` field would produce more side effect than having a wrong `immersive` data

src/components/webviews/jsInteractions/jsCozyInjection.ts
```diff
@@ -21,13 +21,16 @@ const makeMetadata = routeName => {
...
+  window.cozy.isSecureProtocol = ${isSecureProtocol ? 'true' : 'false'}
```
Ldoppea, line 33: This `isSecureProtocol ? 'true' : 'false'` seems weird

src/components/webviews/jsInteractions/jsLogInterception.spec.tsx
```diff
@@ -1,16 +1,26 @@
...
+  type,
+  args
+}: {
+  type: string
+  args?: string[]
+}): {
+  nativeEvent: {
+    data: string
+  }
+} => ({
```
Ldoppea, line 19: Those types should definitively be extracted somewhere else. Inlining them make the code very difficult to read.

src/components/webviews/jsInteractions/jsLogInterception.ts
```diff
@@ -23,19 +23,30 @@ export const jsLogInterception = `
...
   try {
+    if (!payload || !logger || !logId) return console.error('no payload')
+
     const { data: rawData } = payload.nativeEvent
 
-    const dataPayload = JSON.parse(rawData)
+    const dataPayload = JSON.parse(rawData) as {
+      type: string
+      data?: { type: string; log: string[] }
+    }
```
Ldoppea, line 39: nit: would be more correct to remove this `as` and to convert the next `if` to a typeguard

Ldoppea (approved):
Nice one.

Only the comment about `makeMetadata` is important, other ones are just nit and double-checks

src/libs/functions/getHostname.ts
acezard: I don't feel safe in throwing an error here as I don't know if the caller will expect it (fearing regressions). Logging a message instead then

src/components/webviews/jsInteractions/jsCozyInjection.ts
acezard: I agree but TS doesn't like outputting non-string values in template literals in this case

src/components/webviews/jsInteractions/jsCozyInjection.ts
acezard: Updated

src/components/webviews/jsInteractions/jsLogInterception.spec.tsx
acezard: updated

src/components/webviews/jsInteractions/jsLogInterception.ts
acezard: the issue is JSON methods all return any, have to cast them

PR #1011, opened 2023-11-10
docs: Update backup schema file format to simplify documentation

zatteo:
Previously, I added in docs folder two schemas in two formats about backup :
- the .drawio source to modify it on draw.io
- the .drawio.png result to view it easily on GitHub

Now I added a SVG which is the source and can be easily viewed on GitHub.


Ldoppea (approved):
It would be too complicated to check the potential diffs between png and SVGs, I trust you on this one 👍

PR #1023, opened 2023-11-21
fix: apply correct mimeType in iOS

acezard:
This PR will fix:
- Incorrect video/picture names on iOS
- Crash on mespapiers on iOS
- Incorrect encoded names (%20 instead of spaces) on iOS
- Empty iOS uploads
- Wrong file type on Drive upload (iOS)


src/app/domain/osReceive/services/OsReceiveData.ts
```diff
@@ -42,13 +45,19 @@ const mapFilesToUploadToArray = (
...
+      filePath: decodeURI(file.filePath),
+      fileName: decodeURI(file.fileName),
+      mimeType: getMimeType({
+        name: decodeURI(file.fileName),
+        mimeType: Platform.OS === 'android' ? file.mimeType : undefined // mimeType is not reliable on iOS
+      } as Media)
     },
     status: OsReceiveFileStatus.toUpload,
-    type: file.mimeType
+    type: 'application/pdf'
```
Crash--, line 60: what about jpg/mov or else? 🤔

src/app/domain/osReceive/services/OsReceiveData.ts
```diff
@@ -60,6 +69,8 @@ const onReceiveFiles = (files: ReceivedFile[]): OsReceiveFile[] => {
   return filesArray
 }
 
+export const _onReceiveFiles = onReceiveFiles // for testing
+
```
zatteo, line 73: Still needed ?

src/app/domain/osReceive/services/OsReceiveData.ts
acezard: good catch wasn't supposed to be there at all. I guess the unit test failed at detecting it, I'm going to update it too

src/app/domain/osReceive/services/OsReceiveData.ts
```diff
@@ -12,6 +13,8 @@ import {
   ReceivedFile,
   OS_RECEIVE_PROTOCOL_NAME
 } from '/app/domain/osReceive/models/ReceivedFile'
+import { Media } from '/app/domain/backup/models'
+import { getMimeType } from '/app/domain/backup/services/getMedias'
```
zatteo, line 17: We should move getMimeType from backup service to upload service one day.

src/app/domain/osReceive/services/OsReceiveData.ts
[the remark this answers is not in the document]
acezard: forgot to remove it when I switched from the mac to the PC 🤦

Crash--: Did we check the iOS code to see why we can't rely on the mimeType? It sounds weird 🤔

acezard: well in iOS the mimetype is just ".pdf" for pdf for instance. The stack does not recognize it and the file has no icon. (not really breaking in itself). On Android the mimeType has the expected web format

src/app/domain/osReceive/services/OsReceiveData.ts
```diff
@@ -38,18 +41,36 @@ const processReceivedFiles = (
...
+
+const determineMimeType = (fileName: string, mimeType: string): string =>
+  getMimeType({
+    name: decodeFileName(fileName),
+    mimeType: Platform.OS === 'android' ? mimeType : undefined
+  } as Media)
+
+const createOsReceiveFile = (file: ReceivedFile): OsReceiveFile => {
+  const decodedFileName = decodeFileName(file.fileName)
+  const mimeType = determineMimeType(file.fileName, file.mimeType)
```
Ldoppea, line 54: I remember there was a diff on checking mimeType for Android received files and for iOS received files, did you check that on both platform?

acezard: > I remember there was a diff on checking mimeType for Android received files and for iOS received files, did you check that on both platform?

yes, on Android I just use `mimeType` directly since it's correct. Only on iOS I try to guess it

Ldoppea: Ok and so there is no diff on the name part?

Crash--: > Only on iOS I try to guess it

Don't you fixed that here? https://github.com/cozy/cozy-flagship-app/pull/1023/files#diff-e5fbd9f4b85d62986896fb1d8408ad2d8d5a0f2267b5e0b22a9caef7c97255bb

acezard: > > Only on iOS I try to guess it
> 
> Don't you fixed that here? https://github.com/cozy/cozy-flagship-app/pull/1023/files#diff-e5fbd9f4b85d62986896fb1d8408ad2d8d5a0f2267b5e0b22a9caef7c97255bb

that's true, the Platform checking is then redundant

Ldoppea (approved):

PR #1029, opened 2023-11-22
feat: add file thumbnail/icon on sharing

acezard:
Displays a thumbnail for a file.
Will display an image thumbnail if the file is an image,
otherwise will display an icon.
 Will handle bad/empty file paths and mime types gracefully,
but if it gets to that point, there are bigger problems.
Since this is just a UI component,
it does not handle or correct errors of that nature.

image
image

Image fallback if loading fail
image


src/ui/ImageThumbnail/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  style
+}: ImageThumbnailProps): JSX.Element => (
+  <View style={[createViewStyles(size).view, style]}>
+    {isImageType(mimeType) ? (
+      <Image
+        source={{ uri: getImageUri(filePath) }}
+        // Casting to ImageStyle is necessary here to resolve a TypeScript type mismatch.
+        // The style object returned by StyleSheet.create includes a union of ViewStyle, TextStyle, and ImageStyle.
+        // However, the Image component's style prop specifically expects an ImageStyle.
+        // This cast ensures that TypeScript recognizes the style as ImageStyle, aligning with the expected type for the Image component.
```
Ldoppea: nit: this cast could have been done in the `createImageStyles` itself

Ldoppea (approved):

src/ui/ImageThumbnail/index.tsx
acezard: I was unable to cast it at the style time, only at "run time"

src/ui/ImageThumbnail/helpers.ts
```diff
@@ -0,0 +1,73 @@
...
+    mimeType && typeof mimeType === 'string' && mimeType.includes('image')
+  )
+
+// Param filePath is optional only to handle scenarios where it is not provided.
+// However, in practice, filePath should always be provided.
+export const getImageUri = (filePath?: string): string | undefined =>
+  (filePath &&
+    typeof filePath === 'string' &&
+    (filePath.startsWith('file://') ? filePath : 'file://' + filePath)) ||
+  undefined
```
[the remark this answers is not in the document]
acezard: yeah it's a bit overkill like that, but I assumed this was a low-level check with not much importance so preferred brevity

PR #1032, opened 2023-11-24
fix: Do not intercept manager and scheme links in cozyAppFallbackURL

Ldoppea:
With current implementation, when clicking on a login email, if the app is already connected to a Cozy, then the app tries to open a cozy-app with that link

We don't want that to happen as it would produce an error and also this is not an expected scenario

To prevent that we want to ignore links that start with a scheme or manager links

Note that this implementation is not bullet proof. Instead of preventing forbidden URLs we should instead hande only authorised ones. This should be done in another commit

Also this bug highlighted another bug: when the login by email scenario ends and when the flagship certification is automatic, then the Linking `url` event is triggered another time (in useAppBootstrap). So this would trigger the navigation to a cozy-app

By ignoring manager links we removed that bug's side effect but the incorrect behavior (double event) is still present. We must fix that bug in the future

### DIFF
--- a/src/app/domain/deeplinks/services/DeeplinksParserService.ts
+++ b/src/app/domain/deeplinks/services/DeeplinksParserService.ts
@@ -95,7 +95,17 @@ export const parseFallbackURL = (url: string | null): FallbackUrl => {
 
   try {
     const makeURL = new URL(url)
-    const fallback = makeURL.searchParams.get(FALLBACK_PARAM) ?? undefined
+    let fallback = makeURL.searchParams.get(FALLBACK_PARAM) ?? undefined
+
+    if (
+      fallback?.startsWith('cozy://') ||
+      fallback?.startsWith('https://manager.cozycloud.cc') ||
+      fallback?.startsWith('https://staging-manager.cozycloud.cc') ||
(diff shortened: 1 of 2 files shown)


acezard (approved):

src/app/domain/deeplinks/services/DeeplinksParserService.ts
```diff
@@ -95,7 +95,17 @@ export const parseFallbackURL = (url: string | null): FallbackUrl => {
...
   try {
     const makeURL = new URL(url)
-    const fallback = makeURL.searchParams.get(FALLBACK_PARAM) ?? undefined
+    let fallback = makeURL.searchParams.get(FALLBACK_PARAM) ?? undefined
+
+    if (
+      fallback?.startsWith('cozy://') ||
+      fallback?.startsWith('https://manager.cozycloud.cc') ||
+      fallback?.startsWith('https://staging-manager.cozycloud.cc') ||
+      fallback?.startsWith('https://manager-dev.cozycloud.cc')
```
Crash--, line 104: you have `strings.cloudery.{prod|int|dev}Uri` 

Also, you should test the `mabulle://` schema, no ?

PR #1033, opened 2023-11-24
feat: add cliskTraceFile metadata to clisk trace files

doubleface:
This allows to find these trace files without weak assumptions on the
file name.

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -355,6 +355,9 @@ class ReactNativeLauncher extends Launcher {
       await client.save({
         _type: 'io.cozy.files',
         type: 'file',
+        metadata: {
+          cliskTraceFile: true
+        },
         data:
           `<!-- ${url} -->
         ` +


src/libs/ReactNativeLauncher.js
```diff
@@ -355,6 +355,9 @@ class ReactNativeLauncher extends Launcher {
       await client.save({
         _type: 'io.cozy.files',
         type: 'file',
+        metadata: {
+          cliskTraceFile: true
+        },
```
Crash--, line 360: I don't know what to think about it. 

First: It should be documented if used 
Second: I don't think that we should add whatever we want to the io.cozy.files. The doctype is already pretty big and complex. Adding one more metadata make it mare complex.
Third: Since this is only for debug purpose, I don't think it should be added to the doctype 
4th: I understand the idea, but once you have the regex done based on the name, it's enough now? Why do you consider the filename as a `weak assumptions`?

@paultranvan your thoughts on that?

doubleface: Nothing prevents users to create or import files with the same name in their drive.
If the criteria is a metadata, it is impossible to make a mistake.

paultranvan: From my external perspective, that sounds kind of hacky. And this feeling is worsen by the lack of doc.
I think we should prevent ourselves to add undocumented custom attributes in doctypes, particularly for io.cozy.files that are already very big and shared by many clients :pray: 
At least, if you really need those kind of attributes, I think it would be much more nicer to group them into a `metadata.konnector` object. 
The `metadata` object is already quite a mess, and I am pushing for grouping all scope-related attributes in a dedicated object. See the [qualifications](https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.files_metadata/#files-metadata-examples) for instance.

doubleface:
Since this trace file is not the final target to debug clisk konnectors, the file name is enough at the moment

PR #1035, opened 2023-11-25
refactor: make candidateApps action less noisy

acezard:
It was spamming too much info in logs.
Rather than disabling the whole logging for the feature,
trim the spammy action


zatteo:
Nice !!!!

zatteo (approved):

PR #1036, opened 2023-11-27
fix: don't call getCurrentRoute if nav isn't ready

acezard:
This will removes error logs at app start when jsCozyInjection
was trying to get routeName too fast in the app cycle.
Now it will go through the getCurrentRouteName facade
which handles non-ready nav objects.
Also remove unused injections linked to navigation in useGlobalAppState

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -69,11 +69,7 @@ const App = ({ setClient }) => {
 
   const { initialRoute, isLoading } = useAppBootstrap(client)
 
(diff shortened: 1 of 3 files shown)


zatteo (approved):

zatteo:
**Nothing to do**, but _Also remove unused injections linked to navigation in useGlobalAppState_ could have been another commit.

PR #1037, opened 2023-11-27
fix: don't fetch for candidateapps at start

acezard:
This is triggering unecessary rerenders elsewhere in the app
when the user doesn't has files to upload.
Instead we fetch registry if we detect presence of files.
It might be slower in file upload scenario, but it cleans up a lot
app start rerenders (file upload state is linked to localMethods)

### DIFF
--- a/src/app/view/OsReceive/OsReceiveProvider.tsx
+++ b/src/app/view/OsReceive/OsReceiveProvider.tsx
@@ -64,7 +64,7 @@ export const OsReceiveProvider = ({
   useEffect(() => {
     if (!client || didCall.current || state.filesToUpload.length === 0) return
 
-    const fetchRegistry = async (): Promise<void> => {
+    const fetchAppsAndSetCandidates = async (): Promise<void> => {
       const res = (await client.fetchQueryAndGetFromState({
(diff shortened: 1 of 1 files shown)


src/app/view/OsReceive/OsReceiveProvider.tsx
```diff
@@ -62,7 +62,7 @@ export const OsReceiveProvider = ({
   }, [client?.isLogged])
 
   useEffect(() => {
-    if (!client || didCall.current) return
+    if (!client || didCall.current || state.filesToUpload.length === 0) return
 
     const fetchRegistry = async (): Promise<void> => {
```
Crash--, line 67: nit: We do not call the registry. We only call the cozy to fetch the apps. The method name should reflect that and should not be misleading https://github.com/cozy/cozy-flagship-app/blob/a9ff7f47579260290505bd62e617424c4f550d9e/src/app/domain/osReceive/models/OsReceiveCozyApp.ts#L58

acezard: That's true, updating and then merging

Crash-- (approved):

PR #1038, opened 2023-11-27
feat: Implement print intent

acezard:
- [x] AA - Interface permettant au cozy-app de demander une impression (cozy-intent.call(print) - la cozy-app fourni la base 64 du fichier à imprimer )
- [x] AA - Intégrer la librairie  https://github.com/christopherdro/react-native-print
- [x] AA - Sauvegarde du fichier dans un dossier temporaire (voir OCR)
- [x] AA - Afficher un toast si erreur
- [x] AA - Avoir une méthode de type (isprintavailable)
- [x] Add tests

### DIFF
--- a/src/libs/intents/printBase64Doc.ts
+++ b/src/libs/intents/printBase64Doc.ts
@@ -2,10 +2,12 @@ import { format } from 'date-fns'
 import { Platform } from 'react-native'
 import RNFS from 'react-native-fs'
 import RNPrint from 'react-native-print'
+import Toast from 'react-native-toast-message'
 
 import MiniLog from 'cozy-minilog'
 
 import { getErrorMessage } from '/libs/functions/getErrorMessage'
+import { t } from '/locales/i18n'
 
 const log = MiniLog('intents:print')
 
@@ -58,7 +60,11 @@ export const printBase64Doc = async (base64?: string): Promise<void> => {
         err
       )}`
     )
-    return undefined
+
+    Toast.show({
+      type: 'error',
+      text1: t('error.unknown_error')
+    })
   } finally {
     await safeDeleteFile(path)
   }
--- a/src/libs/intents/printBase64Doc.ts
+++ b/src/libs/intents/printBase64Doc.ts
@@ -1,0 +1,7 @@
+import MiniLog from 'cozy-minilog'
+
+const log = MiniLog('intents:print')
+
+export const printBase64Doc = (base64?: string): void => {
+  log.info('printBase64Doc', base64)
+}
--- a/src/libs/intents/printBase64Doc.ts
+++ b/src/libs/intents/printBase64Doc.ts
@@ -1,7 +1,9 @@
+import RNPrint from 'react-native-print'
+
 import MiniLog from 'cozy-minilog'
 
 const log = MiniLog('intents:print')
 
 export const printBase64Doc = (base64?: string): void => {
-  log.info('printBase64Doc', base64)
+  log.info('printBase64Doc', base64, RNPrint)
 }
--- a/src/libs/intents/printBase64Doc.ts
+++ b/src/libs/intents/printBase64Doc.ts
@@ -1,9 +1,65 @@
+import { format } from 'date-fns'
+import { Platform } from 'react-native'
+import RNFS from 'react-native-fs'
 import RNPrint from 'react-native-print'
 
 import MiniLog from 'cozy-minilog'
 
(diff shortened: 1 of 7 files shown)


src/libs/intents/printBase64Doc.ts
```diff
@@ -11,56 +11,66 @@ import { t } from '/locales/i18n'
...
     if (!base64) throw new Error('No base64 provided')
+    path = getTemporaryPrintFilePath(getFileTypeFromDataURI(base64))
+    const fileContent = getContentFromBase64(base64)
 
     await ensurePrintFolder()
-    await RNFS.writeFile(path, base64, 'base64')
-
-    await RNPrint.print({ filePath: path })
+    await RNFS.writeFile(path, fileContent, 'base64')
+    await RNPrint.print({ filePath: path.replace('file://', '') })
```
zatteo, line 71: Why do you need to do a `path.replace` again ? It is already in `getTemporaryPrintFilePath` no ?

acezard: Yes, need to remove it. With the prefix the print lib can't find the file. Maybe writeFile could not use it so it's clearer

acezard: @zatteo to sum up: from what I understand I need the file prefix when writing the file, but when reading it with the print library, having that prefix will make the print software unable to read the file. I made it clearer by abstracting this file prefix in a "writeFile" local function

zatteo: For backup feature, it seems that I remove 'file://' both for iOS and Android.

I do not know what it is kept on o[cr feature / if it's needed here](https://github.com/cozy/cozy-flagship-app/blob/master/src/app/domain/ocr/services/ocr.ts#L61).

src/libs/intents/printBase64Doc.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (await RNFS.exists(path)) {
+      await RNFS.unlink(path)
+    }
+  } catch (err) {
+    log.error(`Error deleting document: ${getErrorMessage(err)}`)
+  }
+}
+
+const writeFile = async (filePath: string, content: string): Promise<void> => {
+  return await RNFS.writeFile(`${filePath}`, content, 'base64')
```
Ldoppea, line 40: nit: we don't need this templated string anymore here.
```suggestion
  return await RNFS.writeFile(filePath, content, 'base64')
```

src/libs/intents/printBase64Doc.ts
```diff
@@ -0,0 +1,82 @@
...
+    filePath = getTemporaryPrintFilePath(getFileTypeFromDataURI(base64))
+    const fileContent = getContentFromBase64(base64)
+    await ensurePrintFolder()
+    await writeFile(filePath, fileContent)
+    await RNPrint.print({ filePath })
+  } catch (err) {
+    log.error(`Error while printing document: ${getErrorMessage(err)}`)
+    Toast.show({
+      type: 'error',
+      text1: t('error.unknown_error')
```
Ldoppea, line 77: Should we be more specific in the toast message to state that this is about the print feature? Something like `Failed to print, an error occured`

Ldoppea (approved):
Nice PR, commits are well split so it was easy to read 👍

src/libs/intents/printBase64Doc.ts
```diff
@@ -11,56 +11,66 @@ import { t } from '/locales/i18n'
...
 const getPrintFolderPath = (): string => {
   return `${RNFS.TemporaryDirectoryPath}/print/`
 }
 
-const getTemporaryPrintFilePath = (): string => {
+const getTemporaryPrintFilePath = (fileExtension: string): string => {
   const date = format(new Date(), 'yyyyMMddHHmmssSSS')
-  const tempFileName = `temp_print_${date}.printfile`
-
+  const tempFileName = `temp_print_${date}.${fileExtension}`
```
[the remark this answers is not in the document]
acezard: I didn't see it, and when saving as PDF Android discards the temp file name and creates its own filename from scratch. Though in any case, I make the name more end-user presentable if it's ever shown

src/libs/intents/printBase64Doc.ts
acezard: ✅

src/libs/intents/printBase64Doc.ts
acezard: agree ✅

PR #1042, opened 2023-11-29
fix: Convert JSInterception console.warn to log

Crash--:
We still want to send the console.warn() from the flagship app to sentry, but we don't want to send the console.warn coming from the WebApp to sentry. It makes too much noise and since we can't fix them from the flagship app, we don't need to send them. They should be already sent by the wepabb to its own sentry.

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android


acezard (approved):
Maybe adding a comment in the code why we do this (the commit describes it but just to be even more explicit)

PR #1047, opened 2023-12-04
feat: Move LauncherView to App.js

doubleface:
This is needed to allow mespapiers app to control clisk konnectors.

Also needed to convert launcher context to a global context to allow it
to be used in multiple places (CozyWebView, App)

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/__tests__/jest.config.js
+++ b/__tests__/jest.config.js
@@ -13,7 +13,7 @@ const config = {
       '<rootDir>/__tests__/transformer/imageTransformer.js'
   },
   transformIgnorePatterns: [
-    'node_modules/(?!((jest-)?react-native(-.*)?|@react-native(-community)?|p-timeout?|p-wait-for?|@notifee?)/)'
+    'node_modules/(?!((jest-)?react-native(-.*)?|@react-native(-community)?|p-timeout?|p-wait-for?|@notifee?)|@fengweichong/react-native-gzip?/)'
   ],
   rootDir: '../'
 }
--- a/src/App.js
+++ b/src/App.js
@@ -48,6 +48,13 @@ import { useOsReceiveApi } from '/app/view/OsReceive/hooks/useOsReceiveApi'
 import { useSecureBackgroundSplashScreen } from '/hooks/useSplashScreen'
 import { hideSplashScreen } from '/app/theme/SplashScreenService'
 
+import {
+  useLauncherContext,
+  LauncherContextProvider
+} from './screens/home/hooks/useLauncherContext'
+
+import LauncherView from '/screens/konnectors/LauncherView'
+
 // Polyfill needed for cozy-client connection
 if (!global.btoa) {
   global.btoa = encode
@@ -76,11 +83,37 @@ const App = ({ setClient }) => {
   useGeolocationTracking()
   useCozyEnvironmentOverride()
 
+  const {
+    LauncherDialog,
+    canDisplayLauncher,
+    launcherClient,
+    launcherContext,
+    onKonnectorLog,
+    onKonnectorJobUpdate,
+    resetLauncherContext,
+    setLauncherContext
+  } = useLauncherContext()
+
   if (isLoading) {
     return null
   }
 
-  return <RootNavigator initialRoute={initialRoute} setClient={setClient} />
+  return (
+    <>
+      <RootNavigator initialRoute={initialRoute} setClient={setClient} />
+      {canDisplayLauncher() && (
+        <LauncherView
+          launcherClient={launcherClient}
+          launcherContext={launcherContext.value}
+          retry={resetLauncherContext}
+          setLauncherContext={setLauncherContext}
+          onKonnectorLog={onKonnectorLog}
+          onKonnectorJobUpdate={onKonnectorJobUpdate}
+        />
+      )}
+      {LauncherDialog}
+    </>
+  )
 }
 
 const InnerNav = ({ client, setClient }) => {
@@ -111,7 +144,9 @@ const InnerNav = ({ client, setClient }) => {
           translucent
         />
         <IconChangedModal />
-        <App setClient={setClient} />
+        <LauncherContextProvider>
+          <App setClient={setClient} />
+        </LauncherContextProvider>
       </View>
     </NativeIntentProvider>
   )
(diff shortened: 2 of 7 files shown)


src/App.js
```diff
@@ -48,6 +48,12 @@ import { useOsReceiveApi } from '/app/view/OsReceive/hooks/useOsReceiveApi'
 import { useSecureBackgroundSplashScreen } from '/hooks/useSplashScreen'
 import { hideSplashScreen } from '/app/theme/SplashScreenService'
 
+import {
+  useLauncherContext,
+  LauncherContextProvider
+} from './screens/home/hooks/useLauncherContext'
+import LauncherView from './screens/konnectors/LauncherView'
```
Crash--, line 55: alias 🙏

doubleface: Fixed

__tests__/jest.config.js
```diff
@@ -13,7 +13,7 @@ const config = {
       '<rootDir>/__tests__/transformer/imageTransformer.js'
   },
   transformIgnorePatterns: [
-    'node_modules/(?!((jest-)?react-native(-.*)?|@react-native(-community)?|p-timeout?|p-wait-for?|@notifee?)/)'
+    'node_modules/(?!((jest-)?react-native(-.*)?|@react-native(-community)?|p-timeout?|p-wait-for?|@notifee?)|@fengweichong/react-native-gzip?/)'
```
Ldoppea, line 16: Why is this edit needed? Seems like this is not related with the PR, is it?

src/screens/home/hooks/useLauncherContext.tsx
```diff
@@ -95,7 +96,11 @@ export const LauncherContextProvider: React.FC<Props> = ({
             <ErrorParallelKonnectors
               currentRunningKonnector={launcherContext.value.konnector.slug}
               concurrentKonnector={concurrentKonnector}
-              onClose={(): void => setConcurrentKonnector(undefined)}
+              onClose={(): void => {
+                setConcurrentKonnector(undefined)
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+                launcherEvent.emit('launchResult', { cancel: true })
```
Ldoppea, line 102: I'm not sure to understand this one. How do we distinguish this `launchResult` from the one that would be emited by the "already running" connector?

src/components/webviews/CozyWebView.js
```diff
@@ -136,6 +172,10 @@ export const CozyWebView = ({
     return () => BiometryEmitter.off('change', injectSettings)
   }, [injectSettings, webviewRef])
 
+  if (shouldShowCliskDevMode() && launcherContext.state === 'default') {
+    return <CliskDevView setLauncherContext={trySetLauncherContext} />
+  }
+
```
Ldoppea, line 178: By doing this is the CozyWebView, this means that both homescreen and cozyappscreen will display the CliskDevView when `shouldShowCliskDevMode`. So it will be displayed twice. Is that expected/problematic?

src/components/webviews/CozyWebView.js
```diff
@@ -182,6 +222,14 @@ export const CozyWebView = ({
         tryConsole(m, log, logId)
         nativeIntent?.tryEmit(m)
 
+        const data = m?.nativeEvent?.data
+        if (data) {
+          const { message, value } = JSON.parse(data)
+          if (message === 'startLauncher') {
+            trySetLauncherContext({ state: 'launch', value })
+          }
+        }
```
Ldoppea, line 231: nit: It would be perfect if this is moved in an external method like it is done by `tryConsole(m,...), `tryNavigatorShare(m,...)` and `tryCrypto(m,...)`. But only if this is easy. Current implementation is good.

src/App.js
```diff
@@ -76,11 +83,37 @@ const App = ({ setClient }) => {
...
+    <>
+      <RootNavigator initialRoute={initialRoute} setClient={setClient} />
+      {canDisplayLauncher() && (
+        <LauncherView
+          launcherClient={launcherClient}
+          launcherContext={launcherContext.value}
+          retry={resetLauncherContext}
+          setLauncherContext={setLauncherContext}
+          onKonnectorLog={onKonnectorLog}
+          onKonnectorJobUpdate={onKonnectorJobUpdate}
```
Ldoppea, line 111: Now that LauncherView is under LauncherContext, we can retrieve all those props directly in the component using `useLauncherContext`, so we don't need to inject them here.

src/components/webviews/CozyWebView.js
Ldoppea: PS: after some reflexion I indeed think that this may be problematic as we expect all our `CozyProxyWebView`, `CozyWebView`, `ReloadInterceptorWebView` (etc) to always return a WebView and to be interchangeable. So maybe we can keep this code in the HomeScreen where it was?

__tests__/jest.config.js
doubleface: This is needed to run unit tests. Now that CozyWebView imports useLauncherContext, which ends up importing `src/libs/cozyAppBundle/cozyAppBundle.ts` which imports react-native-gzip.

src/screens/home/hooks/useLauncherContext.tsx
doubleface: Actually, there is no distinction. The current account/konnector opened in harvest will receive this message. And since the current opened connector is the one which caused to open the "already running" connector, this is the one we want to cancel.

src/components/webviews/CozyWebView.js
doubleface: Done

src/App.js
doubleface: Yes but the LauncherView is a class component and I can't use hooks in it.

src/components/webviews/CozyWebView.js
doubleface: Right, this is done now

src/App.js
doubleface: And changing the LaucherView to a function would be more difficult at the moment.

doubleface:
@Crash-- @Ldoppea I think the PR is now ready

Ldoppea (approved):

PR #1048, opened 2023-12-05
feat: Remove all react-native-camera-roll patches

zatteo:
Every features in these patches have been merged upstream. Some small modifications of code to fit with what has been merged upstream.


Crash-- (approved):
<3

PR #1049, opened 2023-12-05
feat: download app to share if not installed

acezard:


src/app/domain/osReceive/models/OsReceiveCozyApp.ts
```diff
@@ -55,9 +56,16 @@ export interface AppForUpload {
   slug: string
 }
 
-export const fetchOsReceiveCozyApps = {
-  definition: Q('io.cozy.apps'),
-  options: {
-    as: 'io.cozy.apps/accept_from_flagship'
+export const fetchStackApp = (
```
Crash--, line 59: our convention is to name that : 

```
buildXXXQuery()
```

And then use it like that: 
```
const appQuery = buildAppQuery(slug)
const {data: apps, fetchStatus } = useQuery(appQuery.definition, appQuery.options)
```

`fetchXXX` should only be used when we make the call. Not the step before.

src/app/view/OsReceive/hooks/useOsReceiveScreen.ts
Crash--, line 1: I don't know much about this feature, but did we alreay make a call to `/apps` somewhere before displaying this screen? Can we navigation between several selected app and then make several call to `/apps/{slug}` ?

I'm a bit worried about this request at this point. If we are sure to already have make a call to `/apps` then we should read the cozy-client redux store instead of making an http request. 

(ATM Cozy-Client is not smart enough to know that if you made a `/apps` request, you have already what you need to make the `/apps/{id}` without making the request)

src/app/domain/osReceive/models/OsReceiveCozyApp.ts
```diff
@@ -55,9 +56,16 @@ export interface AppForUpload {
...
+  slug: string
+): {
+  definition: QueryDefinition
+  options: QueryOptions
+} => {
+  return {
+    definition: Q('io.cozy.apps').getById(`io.cozy.apps/${slug}`),
```
Crash--, line 66: You should add a `select` to only get the `_id` for instance since you don't care about the object, you just care about the throw or not. 

(in that case, you should rename the `as` to reflect that you only get the `_id` field)

src/app/view/OsReceive/hooks/useOsReceiveScreen.ts
acezard: @Crash-- No, we only requested the registry afaik (have to double-check on other features). Here I want to check at the last possible moment if the user has application XYZ installed on his stack when trying to share a file. From what I saw it's pretty fast. Requesting the registry for accept_from_flagship app though is incredibly slow (sometimes as long as 1200ms). I asked if something could be done on the backend but it seems it wasn't possible

Crash--: Yeah, `/registry` can be pretty slow ATM. `/apps/:id` is pretty fast but if we can avoid a request, we should. Specially if the `/apps` request is already done.

acezard:
Closing the pull request for now as it is stale and most probably will get ditched as a feature for now (not deleting it though)

PR #1061, opened 2023-12-11
fix: handle disabled button styles

acezard:
They weren't well adjusted. We have to rationalize quickly
these themes issues.
This commit is more a quick fix.
Some buttons in the app don't use the disabled style some do.
Also we need 1 disabled style per variant per theme (that we use in the
app though).

image

### DIFF
--- a/src/app/view/Secure/SetPinView.tsx
+++ b/src/app/view/Secure/SetPinView.tsx
@@ -47,6 +47,9 @@ const SetPinViewSimple = ({
     }
   }
 
+  const isPasswordComplete = (password: string): boolean =>
+    password.length === 4
+
   return (
     <Container>
       <Grid container direction="column" justifyContent="space-between">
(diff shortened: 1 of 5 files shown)


src/app/view/Secure/SetPinView.tsx
```diff
@@ -144,8 +147,14 @@ const SetPinViewSimple = ({
...
             </Grid>
 
-            <Button onPress={handleSecondInputSubmit}>
-              <Typography color="primary" variant="button">
+            <Button
+              onPress={handleSecondInputSubmit}
+              disabled={secondInput.length !== 4}
+            >
+              <Typography
+                color={secondInput.length !== 4 ? 'disabled' : 'primary'}
```
zatteo, line 155: nit: `const isPasswordComplete = (password) => password.length === 4` to factorize and make it very clear

zatteo (approved):

src/ui/Button/index.tsx
```diff
@@ -18,7 +18,7 @@ export const Button = ({
     style={[
       styles.button,
       styles[variant],
-      disabled ? styles.disabled : {},
+      disabled && styles[`${variant}Disabled`],
```
zatteo, line 21: Maybe to check with @Ldoppea who also worked on this here https://github.com/cozy/cozy-flagship-app/pull/1062/commits/272093d81998c6fafd3c622a73862fb2cfa0b087

acezard: bit hard to update, @Ldoppea I think the best way to handle that is you just overwrite my styles during conflict. I update to follow same naming convention in styles though

PR #1062, opened 2023-12-11
feat: Add CozyTheme provider to handle normal and inverted themes and adapt existing UI

Ldoppea:
> [!TIP]
> As usual this PR is meant to be read commit by commit
> Also, to ease reading `styles.ts` files diffs you can hide whitespace diffs

We want the UI to adapt to `normal` and `inverted` themes

Until now we did not support Theming in the UI and we manually applied colors to component based on the expected result (`normal` or `inverted`)

In the IAP feature we will need to reuse the `PromptingPage` component using `normal` theme, but this component is implemented for the `inverted` theme only

So it is time to implement some Theming mechanism

The `CozyTheme` provider will be responsible to enforce a theme (`normal` or `inverted`) to all its children

This provider is inspired by the one on cozy-ui

Then `useCozyTheme` hook can be used on every component and should return the current theme name and its corresponding colors palette

This hook takes a `forcedVariant` parameter that can be used by top level components that are not embedded in the provider

Related files:
https://github.com/cozy/cozy-ui/blob/master/react/providers/CozyTheme/index.jsx

___

This PR also ports some of the following components:
- Button
- Typography
- TextField

And the following views:
- LockScreen
- PinPrompt
- SetPinView
- OsReceiveScreen

Some other components still have to be adapted but this can be done in another PR. The requirement was to adapt the PromptingPage. All edits on previous components were side effects as they would have been broken when porting PromptingPage

Next to this PR, the goal would be to adapt remaining components AND to get rid of [this code](https://github.com/cozy/cozy-flagship-app/blob/master/src/ui/palette.ts#L107-L121) that should not be in the palette (its place is in colours.ts as semantic variables)

___

### TODO:

- [x] Adapt Colors for MaBulle

### DIFF
--- a/src/ui/Button/index.tsx
+++ b/src/ui/Button/index.tsx
@@ -1,7 +1,8 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import { TouchableOpacity, TouchableOpacityProps } from 'react-native'
 
-import { styles } from '/ui/Button/styles'
+import { ButtonStyles, styles as computeStyles } from '/ui/Button/styles'
+import { useCozyTheme } from '/ui/CozyTheme/CozyTheme'
 
 type ButtonProps = TouchableOpacityProps & { variant?: 'primary' | 'secondary' }
 
@@ -12,13 +13,17 @@ export const Button = ({
   style,
   variant = 'primary',
   ...props
-}: ButtonProps): JSX.Element => (
+}: ButtonProps): JSX.Element | null => {
+  const { colors } = useCozyTheme()
+  const styles = useMemo<ButtonStyles>(() => computeStyles(colors), [colors])
+
+  return (
   <TouchableOpacity
     onPress={onPress}
     style={[
       styles.button,
       styles[variant],
-      disabled ? styles.disabled : {},
+        disabled ? styles[`disabled_${variant}`] : {},
       style
     ]}
     disabled={disabled}
@@ -27,3 +32,4 @@ export const Button = ({
     {children}
   </TouchableOpacity>
 )
+}
--- a/src/ui/Button/index.tsx
+++ b/src/ui/Button/index.tsx
@@ -3,11 +3,27 @@ import { TouchableOpacity, TouchableOpacityProps } from 'react-native'
 
 import { ButtonStyles, styles as computeStyles } from '/ui/Button/styles'
 import { useCozyTheme } from '/ui/CozyTheme/CozyTheme'
+import { Typography } from '/ui/Typography'
 
-type ButtonProps = TouchableOpacityProps & { variant?: 'primary' | 'secondary' }
+type ButtonPropsbase = TouchableOpacityProps & {
+  variant?: 'primary' | 'secondary'
+}
+
+type ButtonPropsWithChildren = ButtonPropsbase & {
+  label?: never
+  children: React.ReactNode
+}
+
+type ButtonPropsWithLabel = ButtonPropsbase & {
+  label: string
+  children?: never
+}
+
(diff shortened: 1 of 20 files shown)


Ldoppea:
Note that I'm not happy with the way styles has to be initialized. This code should be done in every stylized component:
```ts
  const { colors } = useCozyTheme()
  const [styles, setStyles] = useState<ButtonStyles | undefined>(undefined)

  useEffect(() => {
    setStyles(computeStyles(colors))
  }, [colors])

  if (styles === undefined) {
    return null
  }
```

If you know a better way to handle this?

Ldoppea:
Just made [this edit](https://github.com/cozy/cozy-flagship-app/compare/9172bb70f9b517ea86e2682e78b6746dee7fef4b..d858cf7ac119ef26510d6c6dd3fc0b9657d43527) and [this one](https://github.com/cozy/cozy-flagship-app/compare/d858cf7ac119ef26510d6c6dd3fc0b9657d43527..f0671d178db22c2ae4eb5c57d1d1bcccaad99c8c) to clean Typography usage

src/ui/Button/index.tsx
```diff
@@ -12,13 +13,25 @@ export const Button = ({
...
+  const { colors } = useCozyTheme()
+  const [styles, setStyles] = useState<ButtonStyles | undefined>(undefined)
+
+  useEffect(() => {
+    setStyles(computeStyles(colors))
+  }, [colors])
+
+  if (styles === undefined) {
+    return null
+  }
```
Crash--, line 42: It feels weird to have to handle that at the component level. We should think about an abstraction to handle that automatically (or use an existing design system)

src/ui/Button/index.tsx
```diff
@@ -26,18 +42,32 @@ export const Button = ({
...
+              color: colors.primaryContrastTextColor
+            },
+            disabled && { color: colors.actionColorDisabled }
+          ]}
+        >
+          {label}
+        </Typography>
+      )}
+    </TouchableOpacity>
+  )
```
Crash--, line 72: With this `label` props it tends to follow cozy-ui, great move 👍

Crash-- (approved):
Nice !

zatteo:
> Note that I'm not happy with the way styles has to be initialized. This code should be done in every stylized component:
> 
> ```ts
>   const { colors } = useCozyTheme()
>   const [styles, setStyles] = useState<ButtonStyles | undefined>(undefined)
> 
>   useEffect(() => {
>     setStyles(computeStyles(colors))
>   }, [colors])
>   
>   // styles is never undefined if I check use cozyTheme
>   if (styles === undefined) {
>     return null
>   }
> ```
> 
> If you know a better way to handle this?

About the `if (styles === undefined)`, `styles` is never undefined ? So we can remove it ?

```ts
  const { colors } = useCozyTheme()
  const styles = computeStyles(colors)
  
  // Or even if we are afraid about performance
  const styles = useMemo(() => computeStyles(colors), [colors])
```

Ldoppea:
I reverted @acezard PR in https://github.com/cozy/cozy-flagship-app/pull/1065

And I reapplied the disable logic for SetPinView's buttons here: https://github.com/cozy/cozy-flagship-app/pull/1062/commits/1046a7ef68173071bdc4aa9cc2d0ff1872ff827c

src/ui/Button/index.tsx
Ldoppea: I changed this a bit based on @zatteo suggestion: https://github.com/cozy/cozy-flagship-app/pull/1062#issuecomment-1852154054

zatteo (approved):

PR #1063, opened 2023-12-12
feat(android): replace legacy icon with adaptive icon

acezard:
- This will fix trimmed icon in share menu
- This will allow future management of app icon dark theme
- This updates our application to latest standards

image asset creator reference
Untitled


white_label/brands/mabulle/android/app/src/main/res/values/base_background.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="base_background">#0065FD</color>
```
Ldoppea, line 3: Where does this color comes from? Shouldn't this be the Primary color? (`#0066FF` for MaBulle)

acezard: That was the color I got on Figma, maybe got an issue when translating from HSL to hex, using your value and removing this file, using colors.xml instead

PR #1066, opened 2023-12-13
feat: add new non-native Modal component

acezard:
This is needed on screens where we do not have agency in what is displayed on top of the screen. While it may be tempting to use it everywhere for consistency, I believe it's better to rely on the react-native Modal component when possible. This should be used as a last resort where FullWindowOverlay for instance disallow us to display anything above it (on the native UI side).

### DIFF
--- a/src/ui/CozyDialogs/ConfirmDialog.tsx
+++ b/src/ui/CozyDialogs/ConfirmDialog.tsx
@@ -8,15 +8,44 @@ import { Typography } from '/ui/Typography'
 import { styles } from '/ui/CozyDialogs/styles'
 
 interface ConfirmDialogProps {
+  /**
+   * Actions to be displayed in the footer of the dialog. This can be a single React element
+   * or an array of elements, typically buttons that the user can interact with.
+   */
   actions:
     | ReactElement<{
         children: ReactElement<{ style: StyleProp<ViewStyle> }>[]
       }>
     | ReactElement<{ style: StyleProp<ViewStyle> }>
+
+  /**
+   * Content of the dialog. This can be a simple string or a React element for more complex content.
+   */
   content: string | ReactElement
+
+  /**
+   * A function that will be called when the dialog is requested to be closed. This does not
+   * handle the closing itself, which should be controlled by the parent component's state.
+   */
   onClose: () => void
+
+  /**
+   * Optional title text to be displayed at the top of the dialog.
+   */
   title?: string
+
+  /**
+   * Optional style or styles to apply to the header component of the dialog. Can be used to
+   * customize the appearance of the dialog's header.
+   */
   headerStyle?: StyleProp<ViewStyle>
+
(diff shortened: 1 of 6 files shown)


src/ui/CozyDialogs/ConfirmDialog.tsx
```diff
@@ -61,14 +98,56 @@ export const ConfirmDialog = ({
 
         <View style={styles.content}>
           {typeof content === 'string' ? (
-            <Typography>{content}</Typography>
```
acezard, line 64: the text is lighter than before here, is it intended? 
image

src/ui/CozyTheme/CozyTheme.tsx
```diff
@@ -10,17 +10,22 @@ interface CozyThemeState {
...
 interface CozyThemeProps {
-  variant: CozyThemeVariant
+  variant?: CozyThemeVariant
   children: ReactNode
 }
 
+// Will default to `normal` variant if none is provided
 export const CozyTheme = ({
-  variant,
+  variant = 'normal',
```
Ldoppea, line 28: I would prefer to have this not optional, as I find `<CozyTheme variant="normal">` easier to understand when reading the code compared to `<CozyTheme>` and the gain is significant enough to justify it.

src/ui/CozyDialogs/ConfirmDialog.tsx
```diff
@@ -41,34 +71,72 @@ const renderActions = (
...
 }
 
+const ModalContent = ({
+  onClose,
+  title,
+  headerStyle,
+  content,
+  actions
+}: Partial<ConfirmDialogProps>): JSX.Element => (
+  <CozyTheme>
```
Ldoppea, line 81: If we put `<CozyTheme>` here this means that we don't expect to have `inverted` modals in the future. Maybe we should put it in the caller side? (LockScreen)

src/ui/Typography/index.tsx
```diff
@@ -22,7 +22,7 @@ type TypographyVariant =
   | 'underline'
   | 'link'
 
-type TypographyColor =
+export type TypographyColor =
```
Ldoppea, line 25: You export this but it seems like you don't use it. Is that on purprose?

src/ui/CozyTheme/CozyTheme.tsx
acezard: ✅

src/ui/CozyDialogs/ConfirmDialog.tsx
acezard: Agreed ✅

src/ui/Typography/index.tsx
acezard: ✅

Ldoppea (approved):

PR #1069, opened 2023-12-13
refactor: Migrate HttpServer code to Typescript

Ldoppea:
We will need to consume HttpServer code from Typescript code, so it is time to convert it to Typescript

This refactor has some parts that may changes the code behaviour, but I consider this should not happens:
- [fix: Rethrow error on fetchAppDataForSlug](https://github.com/cozy/cozy-flagship-app/commit/6017c51505eda6c9118043d0094ba0bfba47043f)
- [fix: Correctly handle isRunning() as async method](https://github.com/cozy/cozy-flagship-app/commit/af367aabd8b38e192f0bff0ffde5cf6bc0fc6f09)
- [fix: Do not await getBaseFolderForFqdnAndSlug in indexGenerator](https://github.com/cozy/cozy-flagship-app/commit/154c1d636f85f12407d69fb47d5dade31ef35690)
- [This throw](https://github.com/cozy/cozy-flagship-app/commit/89bd3f2e4d364e154535cc4767bba550e59a02f6#diff-bd3c9eb780ada9e05e0187f2710977ade83417667aca109513de06d1d0991217R55-R59) that is here to ensure correct typing but this scenario should never happen as a running server always has an URL

### DIFF
--- a/src/libs/httpserver/httpServerProvider.js
+++ b/src/libs/httpserver/httpServerProvider.js
@@ -79,7 +79,8 @@ export const HttpServerProvider = props => {
     }
   }, [path, port])
 
-  const isRunning = () => serverInstance?.isRunning()
+  const isRunning = async () =>
+    (await serverInstance?.isRunning()) ?? false
 
   const getIndexHtmlForSlug = async (slug, client) => {
     try {
--- a/src/components/webviews/CozyProxyWebView.js
+++ b/src/components/webviews/CozyProxyWebView.js
@@ -64,24 +64,28 @@ export const CozyProxyWebView = ({
   }, [htmlContentCreationDate, reload])
 
   useEffect(() => {
-    if (httpServerContext.isRunning()) {
-      if (slug) {
-        initHtmlContent({
-          httpServerContext,
-          slug,
-          href,
-          client,
-          dispatch,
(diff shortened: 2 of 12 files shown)


src/libs/httpserver/httpServerProvider.js
```diff
@@ -79,7 +79,8 @@ export const HttpServerProvider = props => {
     }
   }, [path, port])
 
-  const isRunning = () => serverInstance?.isRunning()
+  const isRunning = async () =>
+    (await serverInstance?.isRunning()) ?? false
```
zatteo, line 110: What is the purpose of `?? false` here ?

isRunning can return something else than a bool ?

zatteo:
Crazy what you found just by migrating 1 feature to ts!

src/libs/httpserver/httpServerProvider.js
Ldoppea: isRunning is not called if `serverInstance` in `undefined` so this would return `bool | undefined` without the `?? false`

zatteo: Oh I did not see the `?.` ..

zatteo (approved):

PR #1070, opened 2023-12-13
[VIO-104] feat: Implement IAP (part 2)

Ldoppea:
This PR:
- Adds cozy's base domain to Android IAP requests si the Cloudery can differentiate between dev and prod environments
- Implements loading and error pages that are displayed during IAP process so the user knows that something is happening
- Implements `restart` push notifications


docs/how-to-debug-notifications.md
```diff
@@ -118,3 +118,52 @@ Be careful to **not** have any overrides concerning cozy-banks in CouchDB `secre
...
+    "notification": {
+      "title": "Titre",
+      "body": "Description"
+    },
+    "data": {
+      "refresh": "true"
+    }
+  }
+}
+```
```
Crash--, line 169: Why don't you use the cozy-stack API to send the notification? I don't know if it's a good thing to directly call fcm.

src/libs/notifications/notifications.ts
```diff
@@ -110,6 +110,59 @@ export const handleServerNotificationOpening = (
...
+ * We want this message to be redirected as a displayed toast
+ *
+ * @returns Method to unregister the handler
+ */
+export const handleServerNotificationOnForeground = (): (() => void) => {
+  return messaging().onMessage(async notification => {
+    log.debug('Received notification from onMessage event')
+
+    if (!notification.notification?.title || !notification.notification.body) {
+      log.error('Received a notification without title or body. Skip.')
```
zatteo, line 174: log.error will go to sentry no ? Is it really needed ?

zatteo (approved):

src/libs/notifications/notifications.ts
Ldoppea: Yes. This case should never happen, and if this happens I suppose we should warn the dev team using Sentry.

zatteo: Don't you think that one day it will ? Like tomorrow product team may decide that we want a notification with just a title ? And then every user who do not update will trigger the log.

src/app/view/IAP/IapProgress.tsx
```diff
@@ -0,0 +1,72 @@
...
+    top: 0,
+    left: 0,
+    right: 0,
+    bottom: 0,
+    justifyContent: 'center',
+    alignItems: 'center',
+    flex: 1,
+    backgroundColor: colors.primaryColor
+  },
+  headerTextStyle: {
```
acezard, line 74: Wondering why we don't use our Typography component instead

acezard (approved):

PR #1072, opened 2023-12-14
docs: Add tips for a better experience

zatteo:


docs/tips-for-a-better-developer-experience.md
```diff
@@ -0,0 +1,19 @@
+# Tips for a better developer experience
+
+## Disable lock screen in Cozy Settings
```
Ldoppea, line 3: This works only if the user's phone is protected behind a OS Lock screen. Otherwise the app will ask the user to add a PIN code at every start. So this may not work on Android emulators by default and this would need to add a lock screen in the emulator's settings as an extra step

So maybe we should mention that?

Ldoppea (approved):

docs/tips-for-a-better-developer-experience.md
zatteo: Right I misunderstood how it works. So best is to : 
- add a PIN to android
- that's all ?

Ldoppea: If enabled in Cozy Settings you'll always get a lock screen in the app

If disabled in Cozy Settings and your phone is not protected, you'll get a CTA screen that ask you to set a pin

If disabled in Cozy Settings and your phone is protected, you won't be bothered anymore

zatteo: Awesome thanks

paultranvan:
This is great, thanks @zatteo !

PR #1075, opened 2023-12-14
fix: start listening to backpress in lock view

acezard:
It seems nothing was listening to the backpress anymore.
Fixed that by starting listening to them directly in the view
as it feels safer than listening from a distant service layer

### DIFF
--- a/src/app/view/Lock/useLockScreen.ts
+++ b/src/app/view/Lock/useLockScreen.ts
@@ -21,6 +21,7 @@ import {
   validatePin
 } from '/app/domain/authorization/services/LockScreenService'
 import { useI18n } from '/locales/i18n'
+import { SecurityNavigationService } from '/app/domain/authorization/services/SecurityNavigationService'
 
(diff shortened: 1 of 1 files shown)


src/app/view/Lock/useLockScreen.ts
```diff
@@ -36,6 +37,17 @@ export const useLockScreenProps = (props: LockScreenProps): LockViewProps => {
   const { fqdn } = getInstanceAndFqdnFromClient(client)
   const { t } = useI18n()
 
+  // It is safer to start listening to backpress events in the UI layer.
```
Ldoppea, line 40: It is "safer" compared to what? Maybe this phrase is missing some context

Ldoppea (approved):

src/app/view/Lock/useLockScreen.ts
acezard: after thinking about it I think this comment doesn't add anything, removing it. the following sentence is useful though

PR #1082, opened 2023-12-18
[VIO-103] feat: Implement new FlagshipUI API

Ldoppea:
The FlagshipUI is responsible to apply the correct theme to status and navigation bars base on the displayed screen or cozy-app

Previous `setFlagshipUI` API had numerous caveats:
- It was difficult to keep track of which component was rendered on top
- It was difficult to handle components unmounting as we had to know which theme to apply by guessing which view would be shown after unmounting the components
- Previous point implied the creation of many hacks in order to have enough context to make a choice (i.e. hideSplashscreen, leaving LockScreen, handle cozy-home theme, keep track of home's konnector state, leaving OsReceiveScreen etc)

This new implementation is an attempt to simplify the API and to reduce the need of hacks and side effects

It is based on new out-of-react service (`FlagshipUIService`) that keeps track of all components that use the `useFlagshipUI()` hook

When a component registers, unregister or change its color, then the `FlagshipUIService` updates its reference array and recompute the new FlagshipUI to be applied

To ease the rendering order calculation, we hardcoded array of ZIndex that every component should use when calling the `useFlagshipUI()` hook

This can be done because the application UI is simple enough so we know which screen should be on top of others

This seems to be the only reliable way to predict screens order. We tried to guess components positions based on React tree using different
hooks (i.e. useMemo, useEffect, useLayoutEffect, useFocus), but none of them would allows us to keep track of screens order

We chose to put this mechanism out of React so changing the status and navigation bars would not produce any re-rendering

Related PR: https://github.com/cozy/cozy-libs/pull/2360

### What can be done next

Some improvements can be done for this API

First one is that this API wouldn't be usable on generic components (i.e. modals). To allow this, we can create a React Provider that would allow any component to retrieve its ZIndex from the corresponding Context. This provider would have to be put in a parent component like we do for the `useFlagshipUI()` hook

Second one it that current implementation only handle react-native side, but we keep the old architecture on cozy-apps. If this new API is proven to be stable and useful, then we can port it on cozy-intent so cozy-apps would be able to use it

Some parts of the code can still be cleaned. This is the case of `setFlagshipUI.ts` and `themeManager.ts` that still have some complex code. It is a bit early to clean them as it is unclear how much we need them with the new API. But we should keep in mind to recheck this part in a few weeks

___

TODO:

- [x] Update cozy-intents after https://github.com/cozy/cozy-libs/pull/2360 being merged
- [x] Migrate LauncherView to the new API
- [ ] Create some documentation
- [x] Create unit tests and fix existing ones
- [ ] Heavily test the new API on multiple devices and scenarios
- [x] Test the removal of LockScreenBar in [feat: Implement useFlagshipUI() in LockScreen view](https://github.com/cozy/cozy-flagship-app/pull/1082/commits/609f120bf3390cc8c351a8d115f42ec243eef528) don't break Biometry more recent Android devices (tested only on Android 8)


src/app/view/FlagshipUI/FlagshipUIService.ts
```diff
@@ -0,0 +1,174 @@
...
+
+const log = Minilog('🖌️ FlagshipUIService')
+
+/**
+ * Declare all screens ZIndex here
+ *
+ * This array should be used everytime we call useFlagshipUI to apply a ZIndex to the component
+ * All ZIndexes are grouped here to ensure no screen overlap
+ */
+export const ScreenIndexes = {
```
acezard, line 16: For me all this is declarative config and shouldn't really be in the same place

acezard: yes a new file as I don't see an existing one filling that role. but maybe it can be handled later as it's non critical

Ldoppea: Done here : https://github.com/cozy/cozy-flagship-app/pull/1082/commits/7b2a6c9394ef67df5b5b4b74b361742eb5bd9872

PR #1088, opened 2023-12-20
feat: implement file sharing from cozy-stack to external apps

acezard:
- new cozy intent
- download file
- share file to external app
- display loading screen


src/app/view/OsReceive/OsReceiveProvider.tsx
```diff
@@ -120,10 +124,53 @@ export const OsReceiveProvider = ({
...
+
+            dispatch({ type: OsReceiveActionType.SetFilesToShare, payload: [] })
+          }
+        )
+      }
+    }
+
+    safePromise(fetchFilesByIdsAndShare)()
+  }),
+    [client, state.filesToShare.length]
```
Ldoppea, line 166: I'm not sure that we should put the `share` feature that is about sending files "out" of the app, inside `OSReceive` feature that is about sending files "in" the app.
The features names are confusing enough, I think we should avoid mixing those concepts

src/locales/en.json
```diff
@@ -18,7 +18,9 @@
     "platformNotSupported": "Platform not supported for this request",
     "serverError": "An error occurred while processing your request",
     "cacheCheckError": "An error occurred while processing your request",
-    "print": "Failed to print, please try again."
+    "print": "Failed to print, please try again.",
+    "shareFiles_one": "We couldn't share your file, please try again.",
+    "shareFiles_other": "We couldn't share your files, please try again."
```
Ldoppea, line 23: In other cozy-apps we have a "plural" form that is declared in the same translation key. Like in this example: 
- locale: https://github.com/cozy/cozy-ui/blob/5900e4fd30ab044fd7ea717f672095337abebbd1/react/UploadQueue/locales/en.json#L4
- code usage: https://github.com/cozy/cozy-ui/blob/5900e4fd30ab044fd7ea717f672095337abebbd1/react/UploadQueue/index.jsx#L287-L290

Do you think we can use the same form here?

src/ui/LoadingOverlay/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  // Always want inverted here because the overlay is on top of a dark background
+  return (
+    <CozyTheme variant="inverted">
+      <View style={[styles.container, style]}>
+        <BlurView style={styles.absolute} blurType="dark" blurAmount={10} />
+
+        <View style={styles.containerContent}>
+          <Typography variant="body1" style={{ marginBottom: hPadding * 0.5 }}>
+            {t('services.osReceive.shareFiles.downloadingFiles')}
```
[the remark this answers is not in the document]
acezard: Definitely, mistake here

- [ ] todo

src/app/view/OsReceive/OsReceiveProvider.tsx
acezard: Well I think it's linked by the fact it's about interacting with files to the OS. I felt creating a whole new domain was overkill. But you're right that the naming is bad. What do you think about keeping all that in a single folder but adapting naming?
I think an important thing for me here was to be able to re-use the structure of the file receiving feature (provider, state reducer, models...). It definitely allowed me to implement this feature much faster than from scratch.

src/locales/en.json
acezard: I can look into it, would be better to have same syntax indeed

- [ ] todo

src/app/domain/osReceive/services/shareFilesService.ts
```diff
@@ -0,0 +1,89 @@
...
+  client: CozyClient,
+  fileIds: string[],
+  callback?: () => void
+): Promise<void> => {
+  try {
+    const fileInfos = await Promise.all(
+      fileIds.map(fileId => fetchFileMetadata(client, fileId))
+    )
+
+    const authToken = client.getStackClient().token.accessToken
```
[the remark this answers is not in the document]
acezard: agree for first point

- [ ] todo

For the second, you mean that's a typing issue right? Should look into it at some point then, for now doing the first point should remove the need to do it for that feature though

src/locales/en.json
acezard: using https://www.i18next.com/translation-function/plurals#interval-plurals

src/app/view/OsReceive/OsReceiveProvider.tsx
Ldoppea: I think I need a bit more explanation to understand why you implemented this faster with this.

There seems to be no overlap between the 2 features except that for the `dispatch({ type: OsReceiveActionType` part, so you would have to implement the same things if in a different feature?

Maybe it is too late to question this part, so I guess **you can merge it as is**.

But let me write a proposal so you can take it or leave it (I leave you the choice):

My understanding of the feature is that it has only 3 states: `IDLE`, `Error` and `Sending`:
- `IDLE` is not a state that needs to be described as this is juste the application running normally
  - For now it is here because OsReceive is a stateful component
- `Error` state is not handled in this feature as you previously extracted it in the ErrorProvider
  - I find this part great as every feature may need to display an error
- `Sending` state is here just to display the new `LoadingOverlay` component
  - Maybe we can make a LoadingProvider like we did for the ErrorProvider?
  - Every feature may need to display a LoadingOverlay. This is already the case of Konnectors that sometime display an equivalent screen.

If we have an ErrorProvider and a LoadingProvider, then we don't need to store any state for the `share` feature as it would only need to call something like `showLoadingOverlay()` then download/share the files in a single method.

The only cons is that the `share` method is outside of react, so we would need to create an event handler the same way we did for the OAuthLimit dialog or the FlagshipUI and this is not really in the React philosophy.
But with this I'm convinced the entire feature would fit in one single file equivalent to `shareFileService.ts` with no side effect on OsReceive part (but you would have to create the LoadingOverlay component anyway)

src/app/domain/osReceive/services/shareFilesService.ts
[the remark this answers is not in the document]
Ldoppea: > you mean that's a typing issue right?

Maybe yes, I'm not sure as I don't know if it is possible to have a token without any accessToken in it.

But yes this is non blocking and should be handled in cozy-client's code now (only if there is a bug)

src/locales/en.json
Ldoppea: Seems weird that we still need the `_other` key but ok.

acezard: @Ldoppea I think it's to handle wrong count like non-numbers, or zero. Not sure how it's handled on cozy-ui I can take a look

Ldoppea (approved):

PR #1089, opened 2023-12-20
feat: Send existing files index to pilot fetch

doubleface:
This index is sent only if the konnector has a permission on
io.cozy.files, to avoid unneccessary requests.

This existing files index is needed to have `shouldReplaceFile`
functionnal in konnectors. But shouldReplaceFile won't throw if it is
not present.

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -22,6 +22,7 @@ import {
   activateKeepAwake,
   deactivateKeepAwake
 } from '/app/domain/sleep/services/sleep'
+import { hasPermission } from '/app/domain/manifest/permissions'
 
 const log = Minilog('ReactNativeLauncher')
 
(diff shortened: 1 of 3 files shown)


src/libs/ReactNativeLauncher.js
```diff
@@ -346,6 +346,16 @@ class ReactNativeLauncher extends Launcher {
...
+
+      const hasFilesPermissions = Object.values(
+        manifest.permissions || []
+      ).some(perm => perm.type === 'io.cozy.files')
+      if (hasFilesPermissions) {
+        const existingFilesIndex = await this.getExistingFilesIndex()
+        const serializableExistingFilesIndex =
+          Object.fromEntries(existingFilesIndex)
+        pilotContext.existingFilesIndex = serializableExistingFilesIndex
+      }
```
Crash--, line 358: Since the flagship app will have more and more to check the permissions. We should start creating a new domain (aka folder) to deal with that instead of adding the complexity here

doubleface: @Crash-- @acezard I updated the code to have a manifest domain. I think this is ready.

Crash--: Can you add a wrapTime to that method?

Crash-- (approved):

src/libs/ReactNativeLauncher.js
doubleface: A wrapTime to `_start` or `hasPermission` ? I'm not sure

Crash--: getExistingFilesIndex ?

doubleface: Already done : https://github.com/cozy/cozy-flagship-app/blob/1a5b00e92f97dc057d65842074a3fbb6c3ff1193/src/libs/ReactNativeLauncher.js#L73

PR #1090, opened 2023-12-20
Manage geolocation tracking quota

zatteo:
We added a geolocation tracking quota flag that says "you can save for free up to X days of geolocation tracking data". This is the implementation of the flag on flagship app side. 

=> Disable geolocation tracking at startup if quota exceeded and show a notification
=> Return quota status when front end ask for geolocation tracking status

### DIFF
--- a/src/app/domain/geolocation/helpers/quota.ts
+++ b/src/app/domain/geolocation/helpers/quota.ts
@@ -45,7 +45,7 @@ const getFirstTimeserie = async (
     } else {
       const { data } = (await client.fetchQueryAndGetFromState({
         definition: Q('io.cozy.timeseries.geojson')
-          .where({ _id: { $gt: null } })
+          .where({ startDate: { $gt: null } })
           .select(['_id', 'startDate'])
           .indexFields(['startDate'])
           .sortBy([{ startDate: 'asc' }])
--- a/src/app/domain/geolocation/helpers/quota.ts
+++ b/src/app/domain/geolocation/helpers/quota.ts
@@ -3,6 +3,7 @@ import { differenceInDays } from 'date-fns'
 import { Q } from 'cozy-client'
 import CozyClient from 'cozy-client'
 import flag from 'cozy-flags'
+import Minilog from 'cozy-minilog'
 
 import { showLocalNotification } from '/libs/notifications/notifications'
 import { t } from '/locales/i18n'
@@ -10,6 +11,8 @@ import { getData, storeData, StorageKeys } from '/libs/localStore/storage'
 
 const MAX_DAYS_TO_CAPTURE_UNLIMITED = -1
 const ONE_DAY = 24 * 60 * 60 * 1000
+
+const log = Minilog('📍 Geolocation')
 
 export interface FirstTimeserie {
   id: string
@@ -60,6 +63,7 @@ export const isGeolocationQuotaExceeded = async (
       await storeData(StorageKeys.FirstTimeserie, firstTimeserie)
     }
   } catch (error) {
+    log.warn('Failed to get or fetch first timeserie', error)
     return false
   }
 
--- a/src/app/domain/geolocation/helpers/quota.ts
+++ b/src/app/domain/geolocation/helpers/quota.ts
@@ -4,6 +4,9 @@ import { Q } from 'cozy-client'
 import CozyClient from 'cozy-client'
 import flag from 'cozy-flags'
 
+import { showLocalNotification } from '/libs/notifications/notifications'
+import { t } from '/locales/i18n'
+
 const MAX_DAYS_TO_CAPTURE_UNLIMITED = -1
 const ONE_DAY = 24 * 60 * 60 * 1000
 interface FirstTimeserie {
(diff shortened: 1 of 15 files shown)


src/app/domain/geolocation/helpers/quota.ts
```diff
@@ -0,0 +1,97 @@
...
+    new Date(firstTimeserie.startDate)
+  )
+
+  return daysSinceFirstCapture > maxDaysToCapture
+}
+
+export const showQuotaExceededNotification = async (): Promise<void> => {
+  const maxDaysToCapture = flag('coachco2.max-days-to-capture') as number | null
+
+  if (typeof maxDaysToCapture !== 'number') {
```
Ldoppea, line 81: I understand this is an optional security check as this method should be called only after `isGeolocationQuotaExceeded()` that already has this check, but is there a reason why you don't re-do the check on the `UNLIMITED` value here like you did at line 31?

Ldoppea (approved):

src/app/domain/geolocation/helpers/quota.ts
```diff
@@ -0,0 +1,118 @@
...
+): Promise<FirstTimeserie | undefined> => {
+  try {
+    const firstTimeserieCachedLocally = await getData<FirstTimeserie>(
+      StorageKeys.FirstTimeserie
+    )
+
+    if (firstTimeserieCachedLocally) {
+      return firstTimeserieCachedLocally
+    } else {
+      const { data } = (await client.fetchQueryAndGetFromState({
```
zatteo, line 45: @paultranvan What do you think about this query ? 

Goal : first geojson timeserie ever

src/app/domain/geolocation/services/tracking.ts
```diff
@@ -43,13 +46,24 @@ export const forceUploadGeolocationTrackingData = async (): Promise<void> => {
...
 }
 
-export const getGeolocationTrackingStatus =
-  async (): Promise<GeolocationTrackingStatus> => {
-    const status = await BackgroundGeolocation.getState()
+export const getGeolocationTrackingStatus = async (
+  client: CozyClient | undefined
+): Promise<GeolocationTrackingStatus> => {
+  if (!client) {
+    throw new Error('You must be logged in to use geolocation tracking feature')
```
acezard, line 62: those kind of errors are not shown to the end user I think we shouldn't use a syntax that looks like end user style, it can give the impression we're supposed to show an error toaster

src/app/domain/geolocation/hooks/tracking.ts
```diff
@@ -14,10 +22,25 @@ export const useGeolocationTracking = (): void => {
     const initializeTracking = async (): Promise<void> => {
       if (!client) return
 
-      const shouldStartTracking = (await getShouldStartTracking()) as boolean
+      const trackingEnabled = await isGeolocationTrackingEnabled()
+
+      if (trackingEnabled) {
```
acezard, line 27: this part might be too nested, it's a bit hard to grok at first sight what's going on, maybe it can be refactor with named functions etc.

Also is there error handling above this hook? We can see a good number of async operations which could be points of failure. But maybe it is handled inside the async functions themselves

acezard (approved):
non blocking comments

src/app/domain/geolocation/hooks/tracking.ts
zatteo: I push two new commits with named function it is very clean now !

src/app/domain/geolocation/helpers/quota.ts
```diff
@@ -0,0 +1,118 @@
+import { differenceInDays } from 'date-fns'
```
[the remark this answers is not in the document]
zatteo: Not really tested for the moment

src/app/domain/geolocation/services/tracking.ts
zatteo: Interesting, never thought about this. Actually it goes back to the intent call so it could be displayed. But here it is not. For photo backup, it is. I will do nothing for the moment but I will keep that in mind.

src/app/domain/geolocation/helpers/quota.ts
```diff
@@ -0,0 +1,118 @@
...
+      StorageKeys.FirstTimeserie
+    )
+
+    if (firstTimeserieCachedLocally) {
+      return firstTimeserieCachedLocally
+    } else {
+      const { data } = (await client.fetchQueryAndGetFromState({
+        definition: Q('io.cozy.timeseries.geojson')
+          .where({ _id: { $gt: null } })
+          .select(['_id', 'startDate'])
```
paultranvan, line 49: You don't need the `_id`, it will be returned anyway. Then, you can remove it from the `.select` and adapt the selector: ` .where({ startDate: { $gt: null } })`

Crash--: Can't we keep the `_id` in the `select`? 

1/ It's more explicit 
2/ Will be easier to have type checking later

paultranvan: Actually I might confuse the files behaviour, where the stack force some fields: https://github.com/cozy/cozy-stack/blob/master/web/files/files.go#L1488
I see nothing like this for other doctypes: https://github.com/cozy/cozy-stack/blob/master/web/data/data.go#L347
So if you need the `_id`, you actually need to add it in the select

zatteo: I just updated `.where({ startDate: { $gt: null } })` . Thanks!

PR #1092, opened 2023-12-22
[VIO-129] feat: Extract Lock related screens from router

Ldoppea:
The Lock related screens are responsible to lock the app when it goes to OS background and to display a CTA to the user when their phone is not protected

Previous implementation declared those screens inside of the App's router

This had numerous caveats:
- It was difficult to keep track of when the app would need to be locked
- It was difficult to keep track of where the user was before the app being locked
  - For example we needed to wait for the HomeView to be initialized and check if a cozy-app or the OSReceive screen should be displayed to give it to the Lock screen as a context (cf [here](https://github.com/cozy/cozy-flagship-app/blob/master/src/screens/home/components/HomeView.js#L209-L230))
- It was difficult to keep track of when hiding the splashscreen
- If a component was calling `navigate()` in reaction of an event while behind the lock screen then the app would leave the lock screen
  - For example this occurred in [OauthClientsLimitService](https://github.com/cozy/cozy-flagship-app/blob/master/src/app/domain/limits/OauthClientsLimitService.ts#L28) when the following conditions were met:
    - the OAuth client limit was exceeded
    - the default app was a cozy-app
    - the app was configured to display the lock screen
- We had to rely on some tricks to ensure the navigation to be done on HomeScreen first and then on the Lock screen so the HomeScreen's webview can start loading while the user enters their credentials 
- In some situation, the Lock screen was impacted by re-renders of the app
  - This was visible when the Biometry was triggered on iOS. Sometime it would appear 2 times, sometimes it would directly dismiss with the following error in the console: `Canceled by another authentication`
 
This PR removes those screen outside of the router

With the new implementation, those screen are now displayed on top of the UI without interaction with the router

This allow the application to live behind the Lockscreen, so:
- The HomeScreen is, by default, pre-loading behind the lock screen
- If the app should start on the HomeScreen, on an AppScreen, or on the OSReceiveScreen has no impact anymore on the LockScreen flow
- There is no need to track where the user was, when hiding the lock screen, the app is already where it was before displaying the lock screen
- There is no risk to have `navigate()` collision anymore, the app can do navigation in background based on events and this won't affect the fact that the Lock screen is always displayed until the user enters their credentials
- The Lock screens are rendrered in parallel of the `<App />` component, so there should be no risk of rerender side effects between each-other

___

TODO:

- [x] clean remaining trace of security flow in HomeView (`hasRenderedOnce` ref, `useEffect` name and logs)
- [x] fix unit tests

### DIFF
--- a/src/app/view/Lock/useLockScreenWrapper.ts
+++ b/src/app/view/Lock/useLockScreenWrapper.ts
@@ -1,0 +1,67 @@
+import { EventEmitter } from 'events'
+
+import { useEffect, useState } from 'react'
+
+const lockScreenEventHandler = new EventEmitter()
+
+const lockScreenEvents = {
+  SHOW: 'SHOW',
+  HIDE: 'HIDE'
+} as const
+
+export const lockScreens = {
+  LOCK_SCREEN: 'LOCK_SCREEN',
+  PASSWORD_PROMPT: 'PASSWORD_PROMPT',
+  PIN_PROMPT: 'PIN_PROMPT',
+  SET_PASSWORD: 'SET_PASSWORD',
+  SET_PIN: 'SET_PIN'
+} as const
+
+export type LockScreenEnum = keyof typeof lockScreens
+
+export const showSecurityScreen = (screen: LockScreenEnum): void => {
+  lockScreenEventHandler.emit(lockScreenEvents.SHOW, screen)
+}
+export const hideSecurityScreen = (screen: LockScreenEnum): void => {
+  lockScreenEventHandler.emit(lockScreenEvents.HIDE, screen)
+}
+
+interface LockScreenState {
+  currentSecurityScreen: LockScreenEnum | undefined
+}
+
+export const useLockScreenWrapper = (): LockScreenState => {
+  const [currentScreen, setCurrentScreen] = useState<
+    LockScreenEnum | undefined
+  >(undefined)
+
+  useEffect(() => {
+    const showHandler = (screen: LockScreenEnum): void => {
+      setCurrentScreen(screen)
+    }
+
+    const hideHandler = (): void => {
+      // for now we don't display multiple secure screens simultaneously
+      // so let's hide them all
+      setCurrentScreen(undefined)
+    }
+
+    const subscriptionShow = lockScreenEventHandler.addListener(
+      lockScreenEvents.SHOW,
+      showHandler
+    )
+    const subscriptionHide = lockScreenEventHandler.addListener(
+      lockScreenEvents.HIDE,
+      hideHandler
+    )
+
+    return (): void => {
+      subscriptionShow.removeListener(lockScreenEvents.SHOW, showHandler)
+      subscriptionHide.removeListener(lockScreenEvents.HIDE, hideHandler)
+    }
+  }, [])
+
+  return {
+    currentSecurityScreen: currentScreen
+  }
+}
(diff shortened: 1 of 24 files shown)


Ldoppea:
I fixed a bug I introduced in the CozyApp fallback detection here: https://github.com/cozy/cozy-flagship-app/compare/c69d4ec154983da33c475119762b5dab1ae09da0..bf946dd7533e9403ed88f4697525eeb00f192b80

Also I fixed unit tests here: https://github.com/cozy/cozy-flagship-app/pull/1092/commits/d89c4911ed7dbf38867b4e03c9d2fe41dfa4124e

src/app/view/Lock/useLockScreenWrapper.ts
```diff
@@ -0,0 +1,67 @@
...
+export const hideSecurityScreen = (screen: LockScreenEnum): void => {
+  lockScreenEventHandler.emit(lockScreenEvents.HIDE, screen)
+}
+
+interface LockScreenState {
+  currentSecurityScreen: LockScreenEnum | undefined
+}
+
+export const useLockScreenWrapper = (): LockScreenState => {
+  const [currentScreen, setCurrentScreen] = useState<
```
acezard, line 34: why not directly use a reducer hook here (instead of useState)? seeing the logic in the component is strongly reminiscent of a reducer structure

src/hooks/useGlobalAppState.ts
```diff
@@ -57,8 +57,6 @@ const onStateChange = (
  */
 
 export const useGlobalAppState = (): void => {
-  // Ref to track if the logic has already been executed
-  const hasExecuted = useRef(false)
```
acezard, line 66: I agree in principle, we must make sure that if for whatever reason this hook is launched twice, it doesn't break the app at other places though

src/hooks/useGlobalAppState.ts
```diff
@@ -57,12 +64,14 @@ const onStateChange = (
  */
 
 export const useGlobalAppState = (): void => {
+  const isFirstStart = useRef(true)
```
acezard, line 67: this looks like the removed `  const hasExecuted = useRef(false)	` no?

src/hooks/useGlobalAppState.ts
Ldoppea: Yes, but it was not used anymore. If you check the current code, there is no place where we call `hasExecuted.current = true` anymore.

This has been remove in https://github.com/cozy/cozy-flagship-app/commit/dbfcf54caff9511d3f3cefa1ed015f60f8c6b023#diff-cee1a6f779963233af0e8891de5e8aec2ca24cdb7e9ba8cae9b8ec4f908dc51d

Also the entire block that used it has been removed. It was just about checking if `getIsSecurityFlowPassed()` was `true` or `false` to display a dedicated log in the console

src/hooks/useGlobalAppState.ts
Ldoppea: Kind of, yes.

The logic is reversed, but it has the same goal of checking if this is the first run or not.

The only difference is that in previous code it was used to display a log (and before that it was used to prevent double navigation), in proposed code it is used to clear the AsyncStorage from `LastActivity` info.

src/app/view/Lock/useLockScreenWrapper.ts
Ldoppea: What are the advantages of using a reducer here? My thoughts was that reducers were only useful when having a complex object in the state. Here we have only a string in it.

acezard (approved):
Approved but in my opinion it is difficult to detect regressions from the code, will have to check carefully at runtime

src/app/view/Lock/useLockScreenWrapper.ts
acezard: yes it might be overkill if it's not expected to scale, let's keep it that way then

PR #1100, opened 2024-01-04
Backup dedup performance

zatteo:
- improve backup dedup performance (10x)
- improve backup dedup readability

### DIFF
--- a/src/app/domain/backup/queries/index.ts
+++ b/src/app/domain/backup/queries/index.ts
@@ -55,8 +55,8 @@ export interface File {
   id: string
   name: string
   path: string
-  created_at: number
-  updated_at: number
+  created_at: string
+  updated_at: string
   md5sum: string
   metadata?: {
     idFromLibrary?: string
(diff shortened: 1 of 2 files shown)


Ldoppea (approved):
You mention the fact that you want to avoid unnecessary `new Date()` calls

Does this mean that date comparizon is heavy enough to be a thing we want to optimize?

If yes, is there any chance that using `date-fns` would be more effective at comparing dates?

zatteo:
> You mention the fact that you want to avoid unnecessary `new Date()` calls
> 
> Does this mean that date comparizon is heavy enough to be a thing we want to optimize?
> 
> If yes, is there any chance that using `date-fns` would be more effective at comparing dates?

Interesting I will keep that in mind especially if it can fasten the comparison in itself. 

For the record, I tried to instanciate every date object before the loop to avoid `new Date` calls, but after the "name comparison first" fix, it gave unsignificant results but complicated the code.

PR #1105, opened 2024-01-05
feat: add LoadingProvider to the app

acezard:
### Overview
The `useLoadingOverlay` hook provides a simple and flexible way to manage a loading overlay in the application. It allows you to show or hide an overlay with a customizable message, ideal for indicating loading states during asynchronous operations like API calls or data processing.

### Using the hook

   Import and use the `useLoadingOverlay` hook in your components to control the overlay.

   ```typescript
   import { useLoadingOverlay } from 'app/view/Loading/LoadingOverlay';

   const MyComponent = () => {
       const { showOverlay, hideOverlay } = useLoadingOverlay();

       // Use showOverlay and hideOverlay as needed
   };
   ```

#### `showOverlay(message?: string): void`
- **Description**: Shows the loading overlay with an optional message.
- **Parameters**:
  - `message`: *(Optional)* The message to display on the overlay.
- **Example**:
  ```typescript
  showOverlay('Loading data...');
  ```

#### `hideOverlay(): void`
- **Description**: Hides the loading overlay.
- **Example**:
  ```typescript
  hideOverlay();
  ```

### Best Practices

- **Explicit Control**: Always ensure to hide the overlay explicitly once the loading process is complete.
- **Error Handling**: In case of asynchronous operations, use `hideOverlay` in both `.then()` and `.catch()` to handle successful and error scenarios.
- **Cleanup**: For operations tied to the component lifecycle, use the `useEffect` cleanup function to hide the overlay.

### Example Usage

```typescript
const FileDownloader = () => {
    const { showOverlay, hideOverlay } = useLoadingOverlay();

    const downloadFiles = async () => {
        showOverlay('Downloading files...');
        try {
            await downloadFileLogic();
        } catch (error) {
            console.error(error);
            // Handle error
        } finally {
            hideOverlay();
        }
    };

    return (
        <button onClick={downloadFiles}>Download</button>
    );
};
```

### Screenshot

Screenshot_1704634993


zatteo (approved):

zatteo:
Like the documentation and PR description a lot.

It seems like we have now two approach to display something "above everything" : 
- overlay with a context
- OAuthClientLimitxxx with events

My opinion is that for the moment it is good to try both to see advantages/drawbacks.

Ldoppea:
> It seems like we have now two approach to display something "above everything" :
> 
> * overlay like here with a context
> * OAuthClientLimitxxx with events (made by @Ldoppea)

Both approaches have their pros and their cons.

@acezard 's implementation seems great since the Overlay is displayed directly from the Provider component. This was not possible for `OAuthClientLimitxxx` because it had to be displayed in the HomeScreen component to allow CozyAppScreen to be displayed on top of him.

Events based way has the advantage of easily preventing react's rerenders and can be called from everywhere (even from out-of-react methods), but the main drawback is that we don't rely on React's mechanisms anymore and we are drifting towards an object/event based application instead of a functional one.

My questions about the Overlay implement are the following:
- Where will the Provider be placed? Around `<App />`? Or somewhere else? Do we expect to declare it in multiple places?
  - I suppose the only component that can be on top of him is the LockScreen so we should check that
- Will we need to display the Overlay from a post-me localMethod?
- Will calling `showOverlay()` trigger a rerender on the `children` component or on the caller component?

acezard:
@Ldoppea 

> - Where will the Provider be placed? Around <App />? Or somewhere else? Do we expect to declare it in multiple places?

I decided to place it directly under the `<ErrorProvider />` inside the `<Nav />` wrapper. It is quite low in the "provider pyramid" but still higher than everything that is an actual view, except the `LockScreen` as you mentioned it. I'm not sure how to handle that we should discuss about it. But you're right that it displays above it if put above in the React tree.

> - Will we need to display the Overlay from a post-me localMethod?

Yes it is needed. That's why the provider should be above the `<NativeIntentProvider>`, so some component can use both providers without issues, or even the `NativeIntentProvider` itself.

> - Will calling showOverlay() trigger a rerender on the children component or on the caller component?

It calls a render cycle only on component consuming the context values returned by the hook. Other children components are not expected to get a render cycle, in theory or in some tests I did (if HomeView use the hook, HomeScreen isn't triggering a render when we hide the overlay)

Ldoppea (approved):
We still have to move the LockScreenWrapper (and [its corresponding SafeAreaProvider](https://github.com/cozy/cozy-flagship-app/pull/1092/commits/0661fd68812e13d43f9daef322839f5264f60441#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67R149-R152))on top of the LoadingOverlay but everything else is good 👍

PR #1097, opened 2023-12-29
[Backup] Experimentation of react-native-mmkv to store local backup config

zatteo:
This branch is **not meant to be merged.** It is just an **experimentation to understand the advantages, drawbacks and implications of react-native-mmkv.**

Backup current implementation uses react-native-async-storage to store local backup config. I decided to store the FULL config (backup status, uploaded media list, media to upload list, ...) in ONE key as a stringified JSON object to avoid any state issue.

But this lead to multiples issues because the local backup config can take tens of MB when there are a lot of medias, so we have : 
- very slow read and write on iOS and Android
- cursor issues on Android ([solved here](https://github.com/cozy/cozy-flagship-app/pull/1003))
- cursor issues on Samsung devices (not solved)

There are multiple solutions to these issues. Here I try react-native-mmkv instead of react-native-async-storage which is known to be very fast. It may be a quick win without changing the algorithm too much and benefiting the entire app.

In this branch, I only try react-native-mmkv with backup feature.

**Performance result**

| | react-native-async-storage  | react-native-mmkv |
| ------------- | ------------- | ------------- |
| getData read 1 | 87.39 ms. |  21.07 ms. 😍 |
| getData parse 1 | 77.93 ms. |  113.39 ms. ⁉️  |
| getLocalBackupConfig 1 | 166.04 ms. |  136.83 ms. |
| getData read 2 | 86.70 ms. |  9.45 ms. 😍 |
| getData parse 2 |  91.78 ms. |  152.54 ms. ⁉️ |
| getLocalBackupConfig 2 | 179.14 ms. |  163.91 ms. |
| storeData parse | 66.36 ms. |  83.10 ms. |
| storeData write | 1609.73 ms. |  205.22 ms. 😍 |

We can see a 5x to 10x performance boost. UX is better. I do not understand the parsing differences and I do not know if they are relevant.

**Other results**

- With RN < 0.71, we need to install react-native-mmkv 2.5.1.
- To migrate completely, we should before use /lib/localStore/storage helpers everywhere before.
- Migration code supplied by react-native-mmkv seems to work correctly.

**Tested**
- [x] very slow read and write on iOS and Android : SOLVED
- [x] cursor issues on Android : SOLVED
- [x] cursor issues on Samsung devices : SOLVED

PR #1114, opened 2024-01-08
Dedup before upload

zatteo:
backup dedup = do not upload medias already present in the cozy

To do so, we need to fetch all remote medias and then compare them with all local medias. It can take some times, especially if we have thousands of medias remotely and locally (up to tens of minutes). That's why we now want to make the dedup media per media before every upload and not for all media before every backup.


src/app/domain/backup/services/manageLocalBackupConfig.ts
```diff
@@ -158,6 +159,43 @@ export const setMediaAsBackuped = async (
...
+  localBackupConfig.lastBackupDate = Date.now()
+
+  // remove media from current backup
+  localBackupConfig.currentBackup.mediasToBackup =
+    localBackupConfig.currentBackup.mediasToBackup.filter(
+      mediaToBackup => !isSameMedia(mediaToBackup, media)
+    )
+
+  localBackupConfig.currentBackup.deduplicatedMediaCount =
+    localBackupConfig.currentBackup.deduplicatedMediaCount + 1
```
Ldoppea, line 194: nit: `+=1`? :p

src/app/domain/backup/services/uploadMedias.ts
```diff
@@ -77,6 +84,22 @@ export const uploadMedias = async (
...
+      const correspondingRemoteFile = getCorrespondingRemoteFile(mediaToUpload)
+
+      if (correspondingRemoteFile) {
+        await setMediaAsBackupedBecauseDeduplicated(
+          client,
+          mediaToUpload,
+          correspondingRemoteFile
+        )
+
+        void onProgress(await getBackupInfo(client))
```
Ldoppea, line 97: Not to be handled in this PR but maybe one possible perf improvement would be to keep an in-memory cache of `getBackupInfo()`?

The idea would be:
- `setBackupInfo` always saves backupInfo in the asyncStorage AND in an in-memory variable
- `getBackupInfo` returns the in-memory variable if defined, else in reads the asyncStorage

I'm not sure how much perf improvement we would get here as this would only save time when reading the state but would change nothing when saving it. Do you have some stats on read vs write perfs?

Ldoppea (approved):

src/app/domain/backup/services/uploadMedias.ts
zatteo: There are some stats here with 20K pictures on Android https://github.com/cozy/cozy-flagship-app/pull/1097

So write are a lot slower but read are stil significant. Actually I like your idea a lot. Keep it in mind for next optimization.

PR #1116, opened 2024-01-08
fix: startTracking should return only when geolocation tracking has been started or has failed

zatteo:
Fixes a bug on coachco2 where we enable geolocation, and toggle does not turn on.

### DIFF
--- a/src/app/domain/geolocation/tracking/index.js
+++ b/src/app/domain/geolocation/tracking/index.js
@@ -70,10 +70,9 @@ export const startTracking = async () => {
         smallIcon: 'mipmap/ic_stat_ic_notification'
       }
     })
-    BackgroundGeolocation.start(() => {
-      Log('Tracking started')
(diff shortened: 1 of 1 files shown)


src/app/domain/geolocation/tracking/index.js
```diff
@@ -70,10 +70,8 @@ export const startTracking = async () => {
         smallIcon: 'mipmap/ic_stat_ic_notification'
       }
     })
-    BackgroundGeolocation.start(() => {
-      Log('Tracking started')
-      storeData(StorageKeys.ShouldBeTrackingFlagStorageAdress, true)
-    })
+    await BackgroundGeolocation.start()
```
paultranvan, line 73: Could you just set back the `Tracking started` log? It is sometimes useful to know when it was started

paultranvan (approved):

paultranvan:
Thanks for cleaning my mistakes!

PR #1118, opened 2024-01-10
Update album for dedup before upload

zatteo:
Adjust album management with dedup https://github.com/cozy/cozy-flagship-app/pull/1114

### DIFF
--- a/src/app/domain/backup/queries/index.ts
+++ b/src/app/domain/backup/queries/index.ts
@@ -66,11 +66,11 @@ export interface File {
 
 export type FilesQueryAllResult = File[]
 
-export const buildAlbumsQuery = (deviceId: string): QueryDefinition => {
+export const buildAlbumsQuery = (): QueryDefinition => {
   return Q(DOCTYPE_ALBUMS)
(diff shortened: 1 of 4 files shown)


Ldoppea (approved):
My understanding is that you don't use the `backupDeviceIds` value on Albums anymore.
So do we really still need it? (vs a `fromBackup` boolean)

If yes, shouldn't you add the new iPhone ID to the Album's list of `backupDeviceIds` if it contains only the old iPhone ID?

zatteo:
> My understanding is that you don't use the `backupDeviceIds` value on Albums anymore. So do we really still need it? (vs a `fromBackup` boolean)
> 
> If yes, shouldn't you add the new iPhone ID to the Album's list of `backupDeviceIds` if it contains only the old iPhone ID?

I do not use them anymore for albums, but I decided to rely on their existence because they are already here for current backup user. If I started from scratch I would have added a 'fromBackup' boolean like you said.

But I agree that it is confusing to use it without updateing it. I think about it. Maybe I will just write some comment/doc about it.

PR #1123, opened 2024-01-12
Refactoring and improvments of notification service

zatteo:
- Code refactored to better understand differences between server and local notification
- Receive notification and display it for all server and local notification (initial notification, background and foreground)


Ldoppea (approved):
> notifee.getInitialNotification is noted as deprecated on notifee
documentation, but currently it is the easier solution that work both
on iOS and Android.

I just checked the deprecation notice on https://notifee.app/react-native/reference/getinitialnotification :
> Deprecated for iOS in favour of onForegroundEvent - you can still use this method on iOS but you will also receive a onForegroundEvent

This is confirmed by what we saw yesterday, `onForegroundEvent` can receive the initial event if called fast enough. Today we don't receive it because we call it too late.

I'm wondering if is some fast devices we can receive the event in both handlers simultaneously? Should we add defensive code to prevent this to happen?

zatteo:
> > notifee.getInitialNotification is noted as deprecated on notifee
> > documentation, but currently it is the easier solution that work both
> > on iOS and Android.
> 
> I just checked the deprecation notice on https://notifee.app/react-native/reference/getinitialnotification :
> 
> > Deprecated for iOS in favour of onForegroundEvent - you can still use this method on iOS but you will also receive a onForegroundEvent
> 
> This is confirmed by what we saw yesterday, `onForegroundEvent` can receive the initial event if called fast enough. Today we don't receive it because we call it too late.
> 
> I'm wondering if in some fast devices we can receive the event in both handlers simultaneously? Should we add defensive code to prevent this to happen?

I agree it's theoretically possible but : 
- never encountered it for the moment
- notifications handlers are set only when user is logged in (so it always take some times before initialization) so it's not likely to happen
- even if it happens, it will do two times the action (so two times redirection or two time restart) which I think is okay

So I propose to not complicate things here for the moment.

BUT if one day we want, there is an id which is common between getInitialNotification and onForegroundEvent. We could use this id to avoid the second notification.

src/libs/notifications/notifications.ts
```diff
@@ -144,6 +160,14 @@ export const handleLocalNotificationOpening = (
     // by `showLocalNotification` but the user did not clicked on it yet
     if (!event.detail.pressAction) return
 
+    const notificationId = event.detail.notification?.id
+
+    if (notificationId === lastLocalNotificationId) {
+      return
```
Ldoppea, line 166: nit: would be nice to print a `log.error()` here so we can track in Sentry if this hypothetical case is real.

Ldoppea (approved):

PR #1126, opened 2024-01-15
Migrate from SafetyNet to Play Integrity (VO-208)

acezard:
Before merging:

- [x] `cozy-flagship-app` has to be updated to the `cozy-client` version using play integration (not published yet)
- [x] integrity flow has to be tested in real world condition when the stack can verify our token

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,24 +133,6 @@ See https://github.com/cozy/react-native-webview/blob/cozy_main/docs/Debugging.m
 - Then run ./scripts/build-debug-offline-apk.sh
 - The output will give you the path to the apk
 
-## How to enable Flagship certification
-
-Flagship certification is the process of verifying that the current running app is a genuine Cozy application.
-
-This verification requires the `cozy-stack` to be configured with the app information.
-
-To enable Flagship certification:
-
-- Retrieve the project's Safetynet Api Key on the pass manager
-  - Or generate a new one following Google documentation: https://developer.android.com/training/safetynet/attestation#add-api-key
-- Put the token in the `ANDROID_SAFETY_NET_API_KEY` variable in your local `.env` file
-- On `src/libs/client.js` set `shouldRequireFlagshipPermissions` to `true`
-- Read `cozy-client` instruction in [flagship-certification/README.md](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/flagship-certification/README.md)
-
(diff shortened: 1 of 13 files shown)


android/app/src/main/java/io/cozy/flagship/mobile/MainApplication.java
```diff
@@ -4,6 +4,7 @@
 import android.content.Context;
 import com.facebook.react.PackageList;
 import com.facebook.react.ReactApplication;
+import sk.kedros.playintegrity.RNGooglePlayIntegrityPackage;
```
Ldoppea, line 7: Why is this needed?

README.md
```diff
@@ -133,24 +133,6 @@ See https://github.com/cozy/react-native-webview/blob/cozy_main/docs/Debugging.m
 - Then run ./scripts/build-debug-offline-apk.sh
 - The output will give you the path to the apk
 
-## How to enable Flagship certification
```
Ldoppea, line 136: I'm wondering if we should keep this section and talk about the google project json file instead? But I wouldn't know what to say as I don't know which part of the file is useful for the certification and which one is useful for push notifications

src/libs/clientHelpers/createClient.ts
```diff
@@ -29,7 +28,7 @@ export const createClient = async (instance: string): Promise<CozyClient> => {
       clientKind: 'mobile',
       clientName: await getClientName(),
       shouldRequireFlagshipPermissions: true,
-      certificationConfig: { androidSafetyNetApiKey }
+      certificationConfig: {}
```
Ldoppea, line 31: Shouldn't we put the `cloudProjectNumber` here?

acezard: > Shouldn't we put the `cloudProjectNumber` here?

yes, has to be the prod number. for dev/local, it doesn't matter as all certification request will fail, so I'm thinking of always using the prod number regardless of env

README.md
acezard: - [ ] Will update

android/app/src/main/java/io/cozy/flagship/mobile/MainApplication.java
acezard: > Why is this needed?

Somehow the app wouldn't build without it, had to link manually

PR #1131, opened 2024-01-16
feat: Handle `planId` parameter in `startIap` url, pass language code to Cloudery and improve IAP

Ldoppea:
This is the last PR for IAP before merging everything

___
**feat: Handle planId parameter in startIap url**
In previous implementation the cloudery did pass only a `productId`
parameter that would contain a `planId` on Android platform

With the new implementation, the cloudery passes the `planId` in a new
parameter in addition to the `productId` that now always contains the
real `productId` on both platforms

This simplifies the `getSubscriptionOffers` method that does not
require a nested map anymore in order to retrieve the missing
`productId`

___

**feat: Add device's language code to Cloudery url when displaying offers**
The stores require us to display localized prices and descriptions to
the user when displaying IAP options

Our understanding is that descriptions and prices should displayed
using the phone's language (vs the cozy's language)

So we want to provide the phone's language code to the Cloudery using
the `lang` query parameter

### DIFF
--- a/src/app/view/IAP/hooks/useClouderyOffer.ts
+++ b/src/app/view/IAP/hooks/useClouderyOffer.ts
@@ -2,6 +2,7 @@ import { Dispatch, SetStateAction, useEffect, useState } from 'react'
 import { Platform } from 'react-native'
 import { initConnection, useIAP } from 'react-native-iap'
 import { getLocales } from 'react-native-localize'
+import Toast from 'react-native-toast-message'
 import type { WebViewNavigation } from 'react-native-webview/lib/WebViewTypes'
 
 import { useClient, useInstanceInfo } from 'cozy-client'
@@ -14,6 +15,8 @@ import {
   buySubscription,
   formatOffers
 } from '/app/domain/iap/services/clouderyOffer'
+import { getErrorMessage } from '/libs/functions/getErrorMessage'
+import { t } from '/locales/i18n'
 
 const log = Minilog('💳 Cloudery Offer')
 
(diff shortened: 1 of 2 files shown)


src/app/view/IAP/hooks/useClouderyOffer.ts
```diff
@@ -116,9 +117,12 @@ export const useClouderyOffer = (): ClouderyOfferState => {
       setPopupUrl(null)
     } else {
       const offersParam = formatOffers(subscriptions)
+      const lang = getLocales()[0].languageCode
```
zatteo, line 134: We have `getDeviceLanguage` that does the same job in our wrapper `src/locales/i18n.ts`

src/app/view/IAP/hooks/useClouderyOffer.ts
```diff
@@ -94,9 +97,20 @@ export const useClouderyOffer = (): ClouderyOfferState => {
...
       const doAsync = async (): Promise<void> => {
-        await initConnection()
-        await getSubscriptions({ skus: SKUS })
-        setPartialPopupUrl(current => current ?? href)
+        try {
+          await initConnection()
+          await getSubscriptions({ skus: SKUS })
+          setPartialPopupUrl(current => current ?? href)
+        } catch (error: unknown) {
+          const errorMessage = getErrorMessage(error)
```
zatteo, line 105: :o I finally understood the purpose of `getErrorMessage`. Will use it in backup feature...

zatteo (approved):

src/app/view/IAP/hooks/useClouderyOffer.ts
Ldoppea: Yep but it is not exported, so after talking with @acezard we chose to directly call `getLocales()` here instead of exporting the other method.

PR #1133, opened 2024-01-17
[DAT-26] feat: Pass stored webhook to pipeline execution 

paultranvan:
When passing a webhook URL to the pipeline execution request, the server
will request it after its completion.
This webhook is typically the openpath trips fetching. Thanks to this,
we will now be able to retrieve the trips as soon as they are produced.

Related to https://github.com/cozy/coachCO2/pull/370

### DIFF
--- a/src/app/domain/geolocation/hooks/tracking.ts
+++ b/src/app/domain/geolocation/hooks/tracking.ts
@@ -1,6 +1,6 @@
 import { useEffect } from 'react'
 
-import { useClient } from 'cozy-client'
+import { useClient, useQuery } from 'cozy-client'
 import Minilog from 'cozy-minilog'
 
 import {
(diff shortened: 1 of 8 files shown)


src/app/domain/geolocation/hooks/tracking.ts
```diff
@@ -45,4 +59,19 @@ export const useGeolocationTracking = (): void => {
 
     void initializeTracking()
   }, [client])
+
+  const webhookQuery = buildServiceWebhookQuery()
+  const webhookResp = useQuery(webhookQuery.definition, webhookQuery.options)
+
+  if (Array.isArray(webhookResp.data) && webhookResp.data.length > 0) {
```
zatteo, line 66: nit: You can webhookResp?.data?.length > 0

zatteo:
Very nice a huge step in user experience!

src/app/domain/geolocation/hooks/tracking.ts
paultranvan: I actually added the Array.isArray because typescript was complaining with: `Property 'length' does not exist on type 'object | any[]'.`

PR #1135, opened 2024-01-19
fix: don't hide splashcreen by default on sec flow (VO-197)

acezard:
It resulted in splashscreen being hidden before the
home webview finished loading


acezard:
- [ ] Don't forget to update react-native-bootsplash after it is published, or else the bug will persist

Ldoppea (approved):

PR #1136, opened 2024-01-19
[VO-171] fix: Force android tools version on CI when signing APK

Ldoppea:
Since January 8 the  `ubuntu-latest` image excludes Android SDKs prior to version 31

By default the `r0adkll/sign-android-release` package uses Android SDK 29. So the build would fail

In order to fix that, we can set the `BUILD_TOOLS_VERSION` env variable to define which version `r0adkll/sign-android-release` should use

Related doc:
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

Related issue: r0adkll/sign-android-release#84

> [!IMPORTANT]
> This seems to also fix the bug we had on `prod` build that would instantaneously crash when opening them on Android devices


Ldoppea:
Fixes VO-171

zatteo (approved):

acezard (approved):

PR #1137, opened 2024-01-19
[VO-258] feat: Allow to send logs my email after clicking on a deep link

Ldoppea:
In order to allow sending logs whatever if the app is logged or not, or even when the app is stuck on SplashScreen, the easiest way to trigger sending logs is to use deep links

By implementing this, user can now click on `cozy://enablelogs` or `https://links.mycozy.cloud/flagship/enablelogs` links to tell the app to start recording logs into a local file

They can then click on `cozy://sendlogs` or `https://links.mycozy.cloud/flagship/sendlogs` links to trigger the OS send email intent pre-filled with log files and Cozy's  support email

Finally then can click on `cozy://disablelogs` or `https://links.mycozy.cloud/flagship/disablelogs` links to tell the app to stop recording logs

This method is not the most user friendly but it has the advantage to be resilient to UI bugs

We may want to find a more user friendly method in the future and keep this one for exceptional cases (or internal ones)

___

This feature is based on the [react-native-file-logger](https://github.com/BeTomorrow/react-native-file-logger) plugin, that does not need to be configured with cozy-minilog as it captures all console calls and redirect them to a file

By default `react-native-file-logger` will redirect all logs to a file in the app's cache directory

Daily rolling is enabled and the max file size is 5 files of 1MB

We want to decrease the log level to `Info` so all the app's logs are redirected

Finally `${now} [${level}]` is added before all log in the resulting file

___

TODO:

- [x] Test on Android (for now Android's build is broken on master)
- [x] Checks if GPS Memory logs are included
- [x] Prevent `hideSplashscreen()` calls to hide the email pane

Improvement ideas:
- [x] Add another deep link to enable/disable logging to file and disable it by default
- [x] Add useful data in the email body (current OS, current app version etc)

### DIFF
--- a/docs/how-to-retrieve-logs-in-release.md
+++ b/docs/how-to-retrieve-logs-in-release.md
@@ -8,6 +8,17 @@ However there are still some solution to retrieve some logs from the device
 
 This document explains how to retrieve them
 
+# Send logs by email
+
+It is possible to send logs by email using the 3 following links:
+
+- https://links.mycozy.cloud/flagship/enablelogs or [cozy://enablelogs](cozy://enablelogs)
+  - tell the app to start recording logs into a local file
+- https://links.mycozy.cloud/flagship/sendlogs or [cozy://sendlogs](cozy://sendlogs)
+  - trigger the OS send email intent pre-filled with log files and Cozy's support email
+- https://links.mycozy.cloud/flagship/disablelogs or [cozy://disablelogs](cozy://disablelogs)
+  - tell the app to stop recording logs
+
 # Android
 
 To retrieve Android logs, you need to use ADB
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -530,6 +530,9 @@ PODS:
     - Firebase/Messaging (= 8.15.0)
     - React-Core
     - RNFBApp
+  - RNFileLogger (0.4.1):
+    - CocoaLumberjack
+    - React
   - RNFileViewer (2.1.5):
     - React-Core
   - RNFS (2.20.0):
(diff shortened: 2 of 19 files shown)


zatteo:
> Add another deep link to enable/disable logging to file and disable it by default

I think it may be needed before releasing.

>  Checks if GPS Memory logs are included

Currently, GPS Memory messages are logged to a cozy-minilog logger AND to react-native-background-geolocation logger. So it should. In the next weeks, I will try to not send them to Sentry, but I want to keep them in the console for debugging purpose.

zatteo:
Can't wait to see it in production. Will be very useful.

Ldoppea:
I added a new commit to this PR: https://github.com/cozy/cozy-flagship-app/pull/1137/commits/f121a8b5866e284eb206fa997865861d50f16243

This ones makes the feature "opt-in" and the user now needs to click a UL to enable login into files.

Also I found a bug on iOS when the app is connected. If for some reason the splashscreen is hidden while the email pane is opened, then the splashscreen plugin will hide the email pane instead (because it blindly hides the topmost view)

I'll have to find a workaround for this, either by showing an `email_splashscreen` while the email pane is opened, or by finding a way to fix the bootsplash library so it always hide the correct view. I'm not sure yet on to do these

Ldoppea:
This PR is ready for review, I fixed previous issue by patching react-native-file-logger package in https://github.com/cozy/cozy-flagship-app/pull/1137/commits/9157df2f39a93d2450236a4175bc3907e04fdbb7

acezard (approved):

PR #1139, opened 2024-01-22
fix: always hide splashscreen on wakeup (VO-267)

acezard:
On appStart the homeview will hide the splashscreen
On wakeup, it is more ambiguous so we force it

### DIFF
--- a/src/hooks/useGlobalAppState.ts
+++ b/src/hooks/useGlobalAppState.ts
@@ -5,10 +5,14 @@ import CozyClient, { useClient } from 'cozy-client'
 import Minilog from 'cozy-minilog'
 
 import { StorageKeys, removeData, storeData } from '/libs/localStore/storage'
-import { showSplashScreen } from '/app/theme/SplashScreenService'
+import {
+  hideSplashScreen,
+  showSplashScreen
+} from '/app/theme/SplashScreenService'
 import { handleSecurityFlowWakeUp } from '/app/domain/authorization/services/SecurityService'
 import { devlog } from '/core/tools/env'
 import { synchronizeDevice } from '/app/domain/authentication/services/SynchronizeService'
+import { lockScreens } from '/app/view/Lock/useLockScreenWrapper'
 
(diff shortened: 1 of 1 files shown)


src/hooks/useGlobalAppState.ts
```diff
@@ -32,6 +36,10 @@ const handleWakeUp = async (
   }
 
   await handleSecurityFlowWakeUp(client)
+
+  // On appStart the homeview will hide the splashscreen
```
Crash--, line 40: even if this is not the "home cozy app" displayed inside the `homeview`?

src/hooks/useGlobalAppState.ts
```diff
@@ -32,6 +36,10 @@ const handleWakeUp = async (
...
+  // On wakeup, it is more ambiguous so we force it
+  if (!isAppStart) await hideSplashScreen(lockScreens.LOCK_SCREEN)
```
Ldoppea, line 42: nit: If you use `lockScreens.LOCK_SCREEN` enum here, then you should use it also in line 23

Ldoppea, line 42: I'm wondering if this `hideSplashScreen(lockScreens.LOCK_SCREEN)` should be called in `LockScreenWrapper` instead.
We can put it next to `hideSplashScreen()` so we hide both the initial SplashScreen and the LockScreen one.

I think this should work as other "non lock" scenario already hide this splashscreen on `SecurityService:61` and `SecurityService:207`

src/hooks/useGlobalAppState.ts
Ldoppea: @acezard I think that I already answered this question during a visio call with @Crash-- 

For now HomeView only render the cozy-home app. And if, in the future, we plan to replace this in some context, then calling `hideSplashScreen()` should be part of the contract.

However if I remember correctly, @Crash-- was thinking about default cozy-app. Then here we display them in the CozyAppScreen. In that case, the HomeView will be still displayed in the background.

PR #1140, opened 2024-01-22
fix: Do not crash at startup when client not ready when fetching geol…

zatteo:
…ocation webhook

### DIFF
--- a/src/app/domain/geolocation/helpers/webhook.ts
+++ b/src/app/domain/geolocation/helpers/webhook.ts
@@ -1,0 +1,33 @@
+import CozyClient from 'cozy-client'
+
+import { buildServiceWebhookQuery } from '/app/domain/geolocation/helpers/index'
+import { storeFetchServiceWebHook } from '/app/domain/geolocation/tracking'
+import { FETCH_OPENPATH_TRIPS_SERVICE_NAME } from '/app/domain/geolocation/tracking/consts'
+
+interface WebhookTrigger {
+  message: {
+    name: string
+  }
+  links?: {
+    webhook?: string
+  }
+}
+
+export const fetchAndStoreWebhook = async (
+  client: CozyClient
+): Promise<void> => {
+  const webhookQuery = buildServiceWebhookQuery()
+  const webhookResp = await client.fetchQueryAndGetFromState(webhookQuery)
+
+  if (Array.isArray(webhookResp.data) && webhookResp.data.length > 0) {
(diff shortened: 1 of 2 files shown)


src/app/domain/geolocation/helpers/webhook.ts
```diff
@@ -0,0 +1,31 @@
...
+    name: string
+  }
+  links?: {
+    webhook?: string
+  }
+}
+
+export const fetchWebhook = async (client: CozyClient): Promise<void> => {
+  const webhookQuery = buildServiceWebhookQuery()
+  const webhookResp = await client.fetchQueryAndGetFromState(webhookQuery)
```
paultranvan, line 18: Why using `fetchQueryAndGetFromState`  instead of `useQuery`?

src/app/domain/geolocation/hooks/tracking.ts
```diff
@@ -60,18 +48,13 @@ export const useGeolocationTracking = (): void => {
...
+    const initializeWebhook = async (): Promise<void> => {
+      if (!client) return
 
-    const openpathServiceWebHook = data.find(
-      trigger => trigger.message.name === FETCH_OPENPATH_TRIPS_SERVICE_NAME
-    )
-    const webhook = openpathServiceWebHook?.links?.webhook
-    if (webhook) {
-      void storeFetchServiceWebHook(webhook)
+      await fetchWebhook(client)
```
paultranvan, line 55: The naming might be a bit misleading, as the primary goal of the function is to store the webhook value in local storage.
What do you think of `fetchAndStoreWebhook` ?

paultranvan (approved):

src/app/domain/geolocation/hooks/tracking.ts
zatteo: Right!

src/app/domain/geolocation/helpers/webhook.ts
zatteo: We can not use useQuery here because we are not in a hook or a component. And even without the refactoring, we can not use a hook after a condition (basically I do `if client then query`).

Crash--: And you can't use client.query because of https://github.com/cozy/cozy-client/issues/931 & https://github.com/cozy/cozy-client/issues/867

With all the links to the issue, I think we should work on it ^^'

PR #1141, opened 2024-01-23
Polyfill clipboard API [VER-116]

zatteo:
We want to be able to use navigator.clipboard on our cozy apps to copy and paste. But : 
- navigator.clipboard works only on secure context so not on iOS if the cozy app is served by our local http server
- navigator.clipboard does not work on Android inside a react-native-webview (https://github.com/react-native-webview/react-native-webview/issues/2869)

That's why we polyfill here with a react native bridge. We support only readText and writeText.

Tested with a custom mespapiers build (git://github.com/cozy/mespapiers.git#build-test-copy-with-log) showing : 
- a button to use readText 
- a button to use writeText
- a button to use read => throw an error
- a button to use write => throw an error

Tested if the app is served by local http server OR by the stack.

* [x] Tested on iOS
* [x] Tested on Android


Crash-- (approved):
Nice!

PR #1143, opened 2024-01-24
[VO-239 & VO-241] fix: Check if IAP offers exist when checking if IAP is available and improve view transitions

Ldoppea:
This PR improves the IAP experience on phones that are not eligible for IAP and improve screens transition when opening the ClouderyOffer view

This PR includes multiple fixes that are all documented in their own commit, feel free to read their description


src/app/domain/iap/services/availableOffers.ts
```diff
@@ -0,0 +1,8 @@
+import { Platform } from 'react-native'
+
+const IOS_OFFERS = [
```
acezard, line 10: Are those things really "service" level? Looks more like data/configuration

src/app/view/Limits/hooks/useOAuthClientsLimitExceeded.ts
```diff
@@ -38,18 +39,28 @@ export const useOAuthClientsLimitExceeded = (
 
   useEffect(() => {
     const eventCallback = (href: string): void => {
-      if (client === null) {
-        log.error('Client is null, should not happen')
-        return
+      const doAsync = async (): Promise<void> => {
```
acezard, line 42: looks complex for hook logic, I think we could extract that logic outside of react, especially since it doesn't seem closely coupled to react context values

src/libs/intents/localMethods.ts
```diff
@@ -99,7 +100,7 @@ const isAvailable = (featureName: string): Promise<boolean> => {
   } else if (featureName === 'shareFiles') {
     return Promise.resolve(true)
   } else if (featureName === 'iap') {
-    return Promise.resolve(true)
+    return isIapAvailable()
```
acezard, line 103: not in the scope of this feature, but I think we should start thinking about a more elegant way to handle the isAvailable method implementation (more declarative at least)

acezard (approved):
non-blocking comments about coding style/architecture

src/app/view/Limits/hooks/useOAuthClientsLimitExceeded.ts
Ldoppea: I moved it in those commits:
- https://github.com/cozy/cozy-flagship-app/pull/1143/commits/1ef7dcba7fdaa406c53fa6824629c547ab40c72f
- https://github.com/cozy/cozy-flagship-app/pull/1143/commits/790f3a4d4defea638914e782b5905941310a0549

src/app/domain/iap/services/availableOffers.ts
Ldoppea: They are not really "service" level, but they are highly related to availableOffers as their only goal is to get the list of available IAP offers on OS stores.

I'm not a big fan of over splitting files. This service is already less then 30lines so I don't feel the need scatter the info into more files. So I propose to keep it like this. Do you think this is acceptable?

PR #1144, opened 2024-01-26
fix: Do not manage shareFiles errors in flagship app

zatteo:
We need on the front side to know if :
- share succeeded
- share was canceled
- share failed

So let's not catch everything here.


acezard (approved):
is this linked to a ticket?

zatteo:
> is this linked to a ticket?

Indirectly where I need to implement this on mespapiers.

PR #1145, opened 2024-01-26
fix: Avoid infinite restart loop on notification

zatteo:


src/libs/notifications/notifications.ts
```diff
@@ -100,6 +100,11 @@ export const handleInitialServerNotification = async (
 ): Promise<void> => {
   const notification = await messaging().getInitialNotification()
 
+  if (notification?.data?.refresh) {
+    log.info('Ignored refresh on app start to avoid infinite loop')
+    delete notification.data.refresh
```
Ldoppea, line 105: nit: Would it work if we do `delete notification.refresh` inside of `handleNotificationEvent` (l:86) instead?

Ldoppea (approved):

src/libs/notifications/notifications.ts
zatteo: In `handleNotificationEvent` we do not know anymore if we are in a `handleInitialXXXnotification` (so if we come from the 'quit' state) so I would say no?

Ldoppea: Would this have a negative side effect if we do it for non-initial notifications?

zatteo: We would never restart ? Oh I don't know if the object where I `delete` is persisted so I do not know if it would work this way.

Ldoppea: After reading it again I realise my error. I thought the attribute was deleted to prevent the next loop, but is is deleted to prevent any `refresh` notification on app starts. So you are right we cannot mutualize this code in `handleNotificationEvent`

PR #1149, opened 2024-01-29
fix: reset cloud webview on connection retrieval (VO-269)

acezard:
This will avoid a perpetual error state.
We reset the webview state on connection so it doesn't stay stuck

### DIFF
--- a/src/screens/login/components/ClouderyViewSwitch.tsx
+++ b/src/screens/login/components/ClouderyViewSwitch.tsx
@@ -1,4 +1,5 @@
 import Minilog from 'cozy-minilog'
+
 import React, {
   forwardRef,
   useEffect,
@@ -37,6 +38,7 @@ import {
 import { LOGIN_FLAGSHIP_URL } from '/screens/login/components/functions/oidc'
 import { jsPaddingInjection } from '/screens/login/components/functions/webViewPaddingInjection'
 import { APPLICATION_NAME_FOR_USER_AGENT } from '/constants/userAgent'
+import { navigationRef } from '/libs/RootNavigation'
 
 const log = Minilog('ClouderyViewSwitchProps')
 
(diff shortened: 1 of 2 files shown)


src/screens/login/components/ClouderyViewSwitch.tsx
```diff
@@ -210,6 +212,21 @@ const ClouderyWebView = forwardRef(
       onMessage?.(event)
     }
 
+    const handleError = async (webviewErrorEvent: unknown): Promise<void> => {
+      try {
+        const isOffline = await NetService.isOffline()
+        isOffline &&
+          NetService.handleOfflineWithCallback(() => {
+            webviewRef.current?.reload() // Have to reload the webview when the user is back online or it will stay errored
+            navigationRef.navigate(routes.welcome) // Go back to the welcome screen to leave the offline screen (offline screen should be refactored to be a modal)
```
Crash--, line 221: > offline screen should be refactored to be a modal

Can't we take the opportunity to do it? If @Ldoppea is agreed with this take?

Crash-- (approved):

src/screens/login/components/ClouderyViewSwitch.tsx
Ldoppea: Actually, i'm not sure about what is the best option for the offline screen.

My arguments for using modals for lock screen or clouderyOffer screen is that they should not have any impact on the application state when displayed.

Here the offline screen is often displayed when the webviews are broken. So it may be useful to have a navigation that enforce reseting the view.

That being said, by reading your `reload()` call, I understand that state was not reseted with the current way of handling offline?

To sumarize, I see 2 approaches:
- A screen that unmount the current screen while displaying the offline one
  - pro: we ensure the screen is reset and that no "offline" related bug remains
  - cons: we need to track context to reapply it when connexion is restored
- A modal that keep the app's state intact
  - pro: we don't need to track context
  - cons: the screen need to reset itself when connexion is restored (this may be tricky is some screens)

PR #1151, opened 2024-01-30
fix: handle some dimensions race conditions (VO-264)

acezard:
This will use the hook return values if possible, otherwise use 24px as default height on Android.
This will remove cases where the dimensions were wrongfully set at the start of the app to be 0, resulting in overlap in statusbar and application
This PR also ensures all getDimensions() call that were at the module level are now moved into the application lifecycle


zatteo (approved):

zatteo:
nit: First commit start with `temp:`

zatteo:
I do not understand what is the fixed race condition ?

PR #1153, opened 2024-01-30
feat: add a timeout for splashscreen (VO-321)

acezard:
Due to instability we decide to hide the splashcreen after a set value
(here 15s).
A message will also be sent to Sentry should that error arise.
This is mainly to prevent infinitely stuck splashscreen while the app
below is still functioning

### DIFF
--- a/src/app/theme/SplashScreenService.spec.ts
+++ b/src/app/theme/SplashScreenService.spec.ts
@@ -1,0 +1,80 @@
+import RNBootSplash from 'react-native-bootsplash'
+
+import { logToSentry } from '/libs/monitoring/Sentry'
+import config from '/app/theme/config.json'
+import {
+  showSplashScreen,
+  hideSplashScreen,
+  setAutoHideDuration,
+  splashScreens
+} from '/app/theme/SplashScreenService'
+
+jest.mock('react-native-bootsplash', () => ({
+  show: jest.fn(() => Promise.resolve()),
+  hide: jest.fn(() => Promise.resolve()),
+  getVisibilityStatus: jest.fn(() => Promise.resolve('hidden'))
+}))
+
+jest.mock('/app/view/FlagshipUI', () => ({
+  flagshipUIEventHandler: {
+    emit: jest.fn()
+  },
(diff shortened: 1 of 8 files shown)


src/app/theme/SplashScreenService.ts
```diff
@@ -2,19 +2,31 @@ import RNBootSplash, { VisibilityStatus } from 'react-native-bootsplash'
...
 } as const
+
 type SplashScreenEnumKeys = keyof typeof splashScreens
 export type SplashScreenEnum = (typeof splashScreens)[SplashScreenEnumKeys]
 
-export class SplashScreenService {
-  show = showSplashScreen
-  hide = hideSplashScreen
+let autoHideTimer: NodeJS.Timeout | null = null
+let autoHideDuration = config.autoHideDuration // Default timeout duration in milliseconds
```
zatteo, line 17: Why is it outside in a global value ? It is not possible to pass it as a param of `showSplashScreen` with a default value ?

src/app/theme/SplashScreenService.ts
```diff
@@ -29,12 +41,41 @@ export const showSplashScreen = (
     }
   )
 
+  if (autoHideTimer) {
+    clearTimeout(autoHideTimer)
+  }
+
+  autoHideTimer = setTimeout(() => {
```
zatteo, line 48: I am not sure about it, but I think we will autohide the `secure_background`. This one should not use autohide.

src/app/theme/SplashScreenService.spec.ts
zatteo, line 1: Nice tests!

src/app/theme/SplashScreenService.ts
acezard: I preferred to separate what I considered as plain config from the actual code, so the timeout value is implicit/declarative and at the same time clearly visible in the config file

src/app/theme/SplashScreenService.ts
acezard: - [ ] to update

zatteo (approved):

PR #1158, opened 2024-02-02
 fix: Do not log geolocation plugin log in sentry 

zatteo:


Ldoppea (approved):
What do you mean by `They could be added in breadcrumb` ?

I had to check to understand, so it is worth to mention that this works because `devlog()` does nothing in production

zatteo:
> What do you mean by `They could be added in breadcrumb` ?
> 
> I had to check to understand, so it is worth to mention that this works because `devlog()` does nothing in production

Updated with a more complete commit message 👍

PR #1162, opened 2024-02-07
fix: handle timeout issues in splashscreen (VO-231)

acezard:
This fix introduces several changes in splashscreen failsafe management:

- More logging
- Initial timer on "global" to handle scenarios where no showSplashScreen() are called
- Use a map to allow multiple timers coexisting
- On background handling (delete every timer)
- On active handling (reset timers that existed before going to background to the timeout value, this will allow us to ignore all problems related to unreliable timeout in background mode, remember that these timeout should not be needed in normal circumstances)

Examples

Normal flow:
image

Going from active to background to active
image

On this one I had a bug (linked to hot reloading so can't say exactly what happened), but we can still see the timeout managed the issue in the end
image

### DIFF
--- a/src/app/theme/SplashScreenService.ts
+++ b/src/app/theme/SplashScreenService.ts
@@ -19,7 +19,10 @@ type SplashScreenEnumKeys = keyof typeof splashScreens
 export type SplashScreenEnum = (typeof splashScreens)[SplashScreenEnumKeys]
 
 // Using a map to handle multiple timers
-const autoHideTimers: Record<string, NodeJS.Timeout | undefined> = {}
+const autoHideTimers = {} as Record<
+  SplashScreenEnum | string,
+  NodeJS.Timeout | undefined
+>
 
 let autoHideDuration = config.autoHideDuration // Default timeout duration in milliseconds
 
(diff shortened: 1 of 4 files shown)


zatteo (approved):

src/app/theme/SplashScreenService.ts
```diff
@@ -72,28 +78,155 @@ export const showSplashScreen = (
...
+    `App is becoming active. Evaluating timers to reset...`
+  )
+
+  const timersToReset = Object.keys(activeTimersAtBackground)
+
+  if (timersToReset.length === 0) {
+    splashScreenLogger.info(`No splash screen timers to reset.`)
+  } else {
+    timersToReset.forEach(splashScreen => {
+      if (activeTimersAtBackground[splashScreen]) {
```
Ldoppea, line 196: Why do we need this check? Seems like no code can set this value to false, and because `timersToReset` is binded to `activeTimersAtBackground` keys, seems like this condition is always `true`?

src/app/theme/SplashScreenService.ts
```diff
@@ -72,28 +78,155 @@ export const showSplashScreen = (
...
+        )}" for splash screen "${bootsplashName}" after manual hide`
+      )
+    }
+
+    clearTimeout(autoHideTimers[bootsplashName])
+    Reflect.deleteProperty(autoHideTimers, bootsplashName)
+  }
 }
+
+let activeTimersAtBackground: Record<string, boolean> = {}
```
Ldoppea, line 183: nit: would using something like `Record<key of SplashScreenEnum, boolean>` prevent the cast on line 200?

src/app/theme/SplashScreenService.ts
```diff
@@ -72,28 +78,155 @@ export const showSplashScreen = (
...
+  splashScreenLogger.debug(
+    `App is going to background. Clearing active timers...`
+  )
+
+  Object.keys(autoHideTimers).forEach(splashScreen => {
+    if (autoHideTimers[splashScreen]) {
+      splashScreenLogger.info(
+        `Clearing timer for splash screen "${splashScreen}". Will reset on active.`
+      )
+      activeTimersAtBackground[splashScreen] = true // Mark as active
```
Ldoppea, line 219: nit: Seems like values is always `true` some maybe we can use an array instead of a dict? So we would not have to use the `Object.keys` stuff on line 190

src/app/theme/SplashScreenService.ts
```diff
@@ -72,28 +78,155 @@ export const showSplashScreen = (
...
+      clearTimeout(autoHideTimers[splashScreen])
```
Ldoppea, line 220: nit: Should we use `destroyTimer` here?

Ldoppea (approved):

src/app/theme/SplashScreenService.ts
```diff
@@ -180,45 +183,46 @@ const destroyTimer = (
...
   )
 
-  Object.keys(autoHideTimers).forEach(splashScreen => {
-    if (autoHideTimers[splashScreen]) {
+  const activeTimers = Object.keys(autoHideTimers) as SplashScreenEnum[]
+
+  activeTimers.forEach(splashScreen => {
+    const timerId = autoHideTimers[splashScreen]
+
+    if (timerId) {
```
acezard, line 218: technically will always be defined but it's simpler in terms of typing to assume the impossible

PR #1169, opened 2024-02-16
Migrate all AsyncStorage direct calls to our storage wrapper

zatteo:
```
### 🔧 Tech

* Migrate all AsyncStorage direct calls to our storage wrapper
```

Now every react-native-async-storage calls are centralized in our storage wrapper. This is a prerequisite to potentially migrate to react-native-mmkv or another library.

**Some cases I encountered**

- a JSON object is stored stringified with `AsyncStorage.setItem` and is parsed after `AsyncStorage.getItem`
  => because getData and storeData already does the stringify/parse, I removed the remaining `JSON.stringify` and `JSON.parse` calls when migrating to getData and storeData. Ex. : Bundle, Oauth
- a string is stored directly with `AsyncStorage.setItem`
  => getData will try to parse a not stringified string if there is already a value stored directly with `AsyncStorage.setItem`. I added a compatibilty check.  Ex. : DefaultRedirectionUrl

**What did I test**

- Logging in with previous app version (master) ; migrate to this version; use the app
- Logging in with this version; use the app

#### Checklist

* [x] Tested on iOS
* [x] Tested on Android
* [x] Test coverage

### DIFF
--- a/src/libs/intents/localMethods.ts
+++ b/src/libs/intents/localMethods.ts
@@ -18,7 +18,7 @@ import { setHomeThemeIntent } from '/libs/intents/setHomeThemeIntent'
 import strings from '/constants/strings.json'
 import { EnvService } from '/core/tools/env'
 import { clearCookies } from '/libs/httpserver/httpCookieManager'
-import { clearData } from '/libs/localStore/storage'
+import { clearCozyData } from '/libs/localStore/storage'
 import { deleteKeychain } from '/libs/keychain'
 import { hideSplashScreen } from '/app/theme/SplashScreenService'
 import { openApp } from '/libs/functions/openApp'
@@ -66,7 +66,7 @@ export const asyncLogout = async (client?: CozyClient): Promise<null> => {
   await client.logout()
   await deleteKeychain()
   await clearCookies()
-  await clearData()
+  await clearCozyData()
   // eslint-disable-next-line @typescript-eslint/no-unsafe-call
   await stopTrackingAndClearData()
   RootNavigation.reset(routes.welcome, { screen: 'welcome' })
--- a/src/libs/localStore/storage.ts
+++ b/src/libs/localStore/storage.ts
@@ -98,7 +98,7 @@ export const getData = async <T>(name: StorageKey): Promise<T | null> => {
   }
 }
 
-export const clearData = async (): Promise<void> => {
+export const clearCozyData = async (): Promise<void> => {
   try {
     const keys = Object.values(CozyPersistedStorageKeys)
 
--- a/src/app/domain/authentication/services/BiometryService.spec.ts
+++ b/src/app/domain/authentication/services/BiometryService.spec.ts
@@ -2,7 +2,7 @@ import {
   CozyPersistedStorageKeys,
   getData,
   storeData,
-  clearData
+  clearCozyData
 } from '/libs/localStore/storage'
 import {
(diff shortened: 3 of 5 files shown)


src/libs/localStore/storage.ts
```diff
@@ -24,21 +28,35 @@ export enum StorageKeys {
...
 
 export type IconsCache = Record<string, { version: string; xml: string }>
 
 export interface StorageItems {
   biometryActivated: boolean
   biometryType?: BiometryType
-  sessionCreatedFlag: string
   iconCache: IconsCache
   firstTimeserie: FirstTimeserie
+  onboardingPartner: object
```
Ldoppea, line 54: What is the reason for not using the `OnboardingPartner` type here? Is that to prevent cyclic imports?

src/libs/localStore/storage.ts
```diff
@@ -24,21 +28,35 @@ export enum StorageKeys {
...
+  clouderyEnv: string
 }
 
 export const storeData = async (
-  name: StorageKeys,
+  name: StorageKeys | DevicePersistedStorageKeys,
```
Ldoppea, line 59: nit: here I would rename `StorageKeys` as `CozyPersistedStorageKeys` and I would create a new `StorageKeys` that is the union of both `CozyPersistedStorageKeys` and `DevicePersistedStorageKeys`

Ldoppea (approved):
Nice, this simplifies a lot the API and this seems to be retro-compatible 👍 

Note to myself: update my "log to files" PR #1137 to use `storage` API

src/libs/localStore/storage.ts
zatteo: To prevent cyclic imports and because I was a bit lazy too. `import type { ... }` does not show a cyclic import at the start of the app so I updated the PR with it. I also improved other types in the meatime.

src/libs/intents/localMethods.ts
```diff
@@ -66,8 +64,6 @@ export const asyncLogout = async (client?: CozyClient): Promise<null> => {
 
   await sendKonnectorsLogs(client)
   await client.logout()
-  await clearClient()
-  await resetSessionToken()
```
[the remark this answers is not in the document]
zatteo: I removed `resetSessionToken`.

Why do you think `clearClient` in `iconTable.ts` may be a potential bug ? cc @acezard if you have insight here

src/libs/localStore/storage.ts
zatteo: Would you also rename `StorageKeys` as `CozyPersistedStorageKeys` in every file ? So that we can quickly distinguish in our code if it is cozy persisted or device persisted ? Or just here ? 

Asking because it will modify 200+ files.

src/libs/intents/localMethods.ts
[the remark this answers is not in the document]
Ldoppea: I anticipate that clearing the client only without cleaning related CozyPersistedStorageKeys may produce bugs.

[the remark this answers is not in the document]
zatteo: Oh so you mean we could replace this we the `clearData` function.

src/libs/localStore/storage.ts
zatteo: Did it (with `rename symbol` 👍)

PR #1170, opened 2024-02-19
fix: implement alternative tar.gz on old android (VO-310)

acezard:
### 🐛 Bug Fixes

* Handle scenario where app crashed constantly on Android API inferior to 26. This was caused by a dependency of dependency (apache.commons 1.21 https://commons.apache.org/proper/commons-compress/changes-report.html#a1.21) that was using API unavailable in older phones. To fix this issue, we forked react-native-gzip and added ungzip scenarios specific to older phones, while not touching it for recent applications.
Please refer to https://github.com/cozy/react-native-gzip/pull/2 for the actual fix implementation


Ldoppea (commented):
I have too comments about this.

First, why did you create the repository in the `cozy-labs` org instead of instead of the `cozy` one? There are already a few react-native forks in the `cozy` one (cf [here](https://github.com/cozy/?q=react-native&type=all&language=&sort=))

Also, I would suggest to use only PRs in those repositories, this would help to list all Cozy's edits from a single search (the PR page) in the future.
However this is nit as I don't expect the original repo to suddenly get new commits in the future, so we won't need to do a merge upstream. But keep in mind that for other repos using PRs would ease a lot.

Also I added 2 comments to the new repo's commits, I don't know how to keep track of them as they are not in a PR so I put them here:
- https://github.com/cozy-labs/react-native-gzip/commit/e93900e863e206241b78873d436139152decded7#r138793443
- https://github.com/cozy-labs/react-native-gzip/commit/fe7ca0278ceb18dcbc2a6b4be9575a7629e5f50d#r138793025

acezard:
> I have too comments about this.
> 
> First, why did you create the repository in the `cozy-labs` org instead of instead of the `cozy` one? There are already a few react-native forks in the `cozy` one (cf [here](https://github.com/cozy/?q=react-native&type=all&language=&sort=))
> 
> Also, I would suggest to use only PRs in those repositories, this would help to list all Cozy's edits from a single search (the PR page) in the future. However this is nit as I don't expect the original repo to suddenly get new commits in the future, so we won't need to do a merge upstream. But keep in mind that for other repos using PRs would ease a lot.
> 
> Also I added 2 comments to the new repo's commits, I don't know how to keep track of them as they are not in a PR so I put them here:
> 
> * [cozy-labs/react-native-gzip@e93900e#r138793443](https://github.com/cozy-labs/react-native-gzip/commit/e93900e863e206241b78873d436139152decded7#r138793443)
> * [cozy-labs/react-native-gzip@fe7ca02#r138793025](https://github.com/cozy-labs/react-native-gzip/commit/fe7ca0278ceb18dcbc2a6b4be9575a7629e5f50d#r138793025)

you're right it should be in cozy namespace, and use PR rather than commit to main directly. Also keep history of the first change and check why it was made exactly

acezard:
@Ldoppea updated there https://github.com/cozy/react-native-gzip/releases/tag/1.0.0

Ldoppea (approved):
Perfect, thanks for your efforts 🎉

PR #1172, opened 2024-02-19
feat: remove draft status when publishing (VO-345)

acezard:
### 🔧 Tech

* This will automatically publish pushed apps in internal track without
the draft status that was previously applied
on all non-cozy apps (was needed at first because GL wasn't
deployed to production yet)

### DIFF
--- a/.github/workflows/build-android-prod.yml
+++ b/.github/workflows/build-android-prod.yml
@@ -112,4 +112,3 @@ jobs:
           packageName: ${{ inputs.brand != 'cozy' && format('io.cozy.flagship.mobile.{0}', inputs.brand) || 'io.cozy.flagship.mobile' }}
           releaseFiles: android/app/build/outputs/bundle/prodRelease/*.aab
           track: internal
-          status:  ${{ inputs.brand != 'cozy' && 'draft' || 'completed' }}


.github/workflows/build-android-prod.yml
```diff
@@ -112,4 +112,3 @@ jobs:
           packageName: ${{ inputs.brand != 'cozy' && format('io.cozy.flagship.mobile.{0}', inputs.brand) || 'io.cozy.flagship.mobile' }}
           releaseFiles: android/app/build/outputs/bundle/prodRelease/*.aab
           track: internal
-          status:  ${{ inputs.brand != 'cozy' && 'draft' || 'completed' }}
```
Ldoppea, line 115: Based on the [documentation](https://github.com/r0adkll/upload-google-play?tab=readme-ov-file#inputs) this wouldn't work:
> One of completed, inProgress, halted, draft. **Cannot be null.**

Instead you should revert to the previous value. You can see it here: https://github.com/cozy/cozy-flagship-app/commit/178f95498547b61643084b1ce48b15f856b9834e

acezard: @Ldoppea I understood as cannot be null but can be undefined, as the key is said to be non required (required: false)

Ldoppea: ah ok, my bad 😅

Ldoppea (approved):

PR #1174, opened 2024-02-20
feat: bypass 2FA when logging with email (VO-380)

acezard:
### ✨ Features

* Do not use 2FA when already logging from an email link


src/screens/login/LoginScreen.js
```diff
@@ -107,6 +108,17 @@ const LoginSteps = ({
...
     const onboardUrl = consumeRouteParameter('onboardUrl', route, navigation)
+    const emailVerifiedCode = consumeRouteParameter(
+      'emailVerifiedCode',
+      route,
+      navigation
+    )
+
+    if (emailVerifiedCode) {
+      // We will use it at a later stage when starting the OAuth flow
+      emailVerifiedCodeRef.current = emailVerifiedCode
```
Ldoppea, line 119: Here I would suggest to put this data into the `state` object for two reasons.

First because the login state is already complex so I think it is better to keep it centralised

Second because, as for the other state's entries we need to ensure it is correctly reset when login is canceled. This is done when calling `setState({step: CLOUDERY_STEP})` (i.e. in the `handleBackPress` callback or in `cancelLogin` method). With your implementation the email verified code is not cleared and will be send to the next login attempt.

Also the email code can be set only in `setInstanceData` as MagicLink is already an email based mechanism and the OIDC is not an email based one, so none of them would receive an email code.
So I think that instead of using the useRef you can safely put the code in the setState at line 190 and retrieve it in the state destructuring at line 347.

Ldoppea:
Related PRs:
- https://github.com/cozy/cozy-stack/pull/4213
- https://github.com/cozy/cozy-stack/pull/4252
- https://github.com/cozy/cozy-stack/pull/4318
- https://github.com/cozy/cozy-client/pull/1447

src/screens/login/LoginScreen.js
acezard: updated

src/screens/login/LoginScreen.js
```diff
@@ -158,7 +164,13 @@ const LoginSteps = ({
   }
 
   const setInstanceData = useCallback(
-    async ({ instance, fqdn }) => {
+    /**
+     * Sets the instance data.
+     * @param {{ instance: Record<string, unknown>, fqdn: string }} instanceData - The first parameter object with `instance` as a record of string keys to unknown values, and `fqdn` as a string.
```
Ldoppea, line 169: instance is supposed to be a string, did you find a case where it was something else? 🤔

acezard: updated

Ldoppea (approved):

PR #1177, opened 2024-02-26
fix: use default empty object in setInstanceData (VO-380)

acezard:
Without it, we risk destructuring an undefined value


Ldoppea (approved):
Nice catch 👍

PR #1179, opened 2024-02-26
[VER-277] Fix missing tracking events

paultranvan:
```
### ✨ Features

*

### 🐛 Bug Fixes

* Fix missing tracking events resulting in no upload

### 🔧 Tech

*
```

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android
* [ ] Test coverage


src/app/domain/geolocation/hooks/useGeolocationTracking.ts
zatteo, line 1: ❤️❤️❤️

src/app/domain/geolocation/tracking/index.js
```diff
@@ -44,11 +44,26 @@ export {
...
+    BackgroundGeolocation.onActivityChange(async event => {
+      return handleActivityChange(event)
+    })
+
+    // Register on motion change
+    BackgroundGeolocation.onMotionChange(async event => {
+      return handleMotionChange(event)
+    })
+
+    // Register on conectivty change
```
zatteo, line 62: nit: connectivity (in any case these comment are not very useful)

src/app/domain/geolocation/tracking/index.js
```diff
@@ -80,20 +95,41 @@ export const startTracking = async () => {
...
-      await BackgroundGeolocation.start()
-      Log('Tracking started')
-    }
+
+    Log('State after init: ' + JSON.stringify(state))
+
     await storeData(
       CozyPersistedStorageKeys.ShouldBeTrackingFlagStorageAdress,
       true
     )
```
zatteo, line 104: This should not be stored here but in the real `startTracking` function.

src/app/domain/geolocation/hooks/useGeolocationTracking.ts
```diff
@@ -27,7 +27,7 @@ export const useGeolocationTracking = (): void => {
       if (trackingEnabled) {
         await checkGeolocationQuota(client)
       } else {
-        await checkShouldStartTracking()
+        await startTrackingIfAllowed()
```
zatteo, line 30: It is not really if "allowed". It is if tracking has been started before, so the shouldStartTracking value is true in local storage but we see when starting the app that the tracking is not started. So tracking is not started even if it should be started.

src/app/domain/geolocation/tracking/index.js
paultranvan: Absolutely, I realized that in parallel. Good catch!

src/app/domain/geolocation/hooks/useGeolocationTracking.ts
paultranvan: Here "Allowed" is in the meaning that the geolocation was enabled by the user. The `shouldStartTracking` is set to true for the first time when the user enables the toggle, so it is kind of a consent, thus the "allowed". 
So I guess both "allowed" and "shouldStart" are correct here, depending on the perspective.
My concern here was especially on the "check" vs "start", as I do not expect a "check" function to do more than checking something. 
Is `startTrackingIfShouldBe` ok? Sounds a bit weird, but understandable?

src/app/domain/geolocation/hooks/useGeolocationTracking.ts
paultranvan: My greatest contribution

src/app/domain/geolocation/hooks/useGeolocationTracking.ts
zatteo: For me it is not only a consent because is it removed when the user disables tracking.

But totally agree about changing _check_ to _start_. `startTrackingIfShouldBe` is like you said weird but understandable. So clearly better that usual but not understandable.

paultranvan: > For me it is not only a consent because is it removed when the user disables tracking.

Well, when you disable the tracking, you kind of remove your consent to be tracked ;)
Anyway, I renamed in `startTrackingIfShouldBe`!

zatteo (approved):

PR #1181, opened 2024-02-28
[VO-428] fix: Prevent the app to stuck on SplashScreen after a restart notification

Ldoppea:
On Android, when the app request a restart, the current Activity is not
directly destroyed and the app react as if it was sent to `background`

This triggers the `useGlobalAppState` hooks which shows the
`LOCK_SCREEN` splashScreen

The after the app effectively restarted, it fails to hide all
splashScreens because the `LOCK_SCREEN` one is not expected to be
displayed (but it is)

To prevent this kind of side effects, we want to fully unmount the
components tree, so all hooks would unregister themselves

This PR also fix some hideSplashScreen scenario and improve logs

```
### ✨ Features

* Update the existing IAP entry with this PR (was not in production yet)

```

___

TODO:

- [x] Upgrade react-native-bootsplash after cozy/react-native-bootsplash#4 being merged

### DIFF
--- a/src/app/theme/SplashScreenService.ts
+++ b/src/app/theme/SplashScreenService.ts
@@ -1,5 +1,8 @@
 import { AppState } from 'react-native'
-import RNBootSplash, { VisibilityStatus } from 'react-native-bootsplash'
+import RNBootSplash, {
+  ResultStatus,
+  VisibilityStatus
+} from 'react-native-bootsplash'
 
 import Minilog from 'cozy-minilog'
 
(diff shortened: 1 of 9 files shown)


src/app/theme/SplashScreenService.ts
```diff
@@ -65,14 +68,18 @@ export const showSplashScreen = async (
   setTimeoutForSplashScreen(bootsplashName)
 
   try {
-    await RNBootSplash.show({ fade: true, bootsplashName })
-    return splashScreenLogger.info(`Splash screen shown "${bootsplashName}"`)
+    const result = await RNBootSplash.show({ fade: true, bootsplashName })
+    splashScreenLogger.info(
+      `Splash screen shown "${bootsplashName}" (${result.toString()})`
```
acezard, line 73: seems like there are typing issues detected by github ci here

Ldoppea: This is because [of](https://github.com/cozy/react-native-bootsplash/pull/4) is not merged yet

zatteo (approved):

PR #1186, opened 2024-03-08
fix: Fix plural translations by importing missing locale-data

zatteo:
Previously, I did not understand why plural translations did not work on iOS. They were returning the key instead of the pluralized translation. I duplicated the translations to support the JSON v3 fallback but it still did not work properly. They were always returning the singular key.

Here, I found that the root of the issue was the missing locale-data import needed by 'intl' and 'intl-pluralrules' polyfill.

So plural translations works and I removed the duplicate translations.

```
### 🐛 Bug Fixes

* Fix plural translations on backup notification
```

#### Checklist

* [x] Tested on iOS
* [x] Tested on Android
* [ ] Test coverage


zatteo:
Related for information : https://github.com/i18next/i18next/pull/2155

acezard (approved):

PR #1192, opened 2024-03-20
[DAT-41] Fix transitions

paultranvan:
```
### ✨ Features

*

### 🐛 Bug Fixes

* Prevent missing trips because of overlapping transitions

### 🔧 Tech

*
```

#### Checklist

* [ ] Tested on iOS
* [ ] Tested on Android
* [ ] Test coverage

### DIFF
--- a/src/app/domain/geolocation/tracking/index.js
+++ b/src/app/domain/geolocation/tracking/index.js
@@ -274,12 +274,13 @@ export const startOpenPathUploadAndPipeline = async ({
   untilTs = 0,
   force = false
 } = {}) => {
+  let uploadedCount = -1
   try {
     const user = await getOrCreateId()
     Log(`User: ${JSON.stringify(user)}`)
 
     // Upload data to openpath server
-    const uploadedCount = await uploadData(user, { untilTs, force })
+    uploadedCount = await uploadData(user, { untilTs, force })
     if (uploadedCount >= 0) {
       // Start openpath pipeline
       const webhook = (await getFetchServiceWebHook()) || ''
@@ -292,5 +293,5 @@ export const startOpenPathUploadAndPipeline = async ({
     Log('Failed openpath processing: ' + err)
     log.error(err)
   }
-  return
+  return uploadedCount
 }
--- a/src/app/domain/geolocation/tracking/index.js
+++ b/src/app/domain/geolocation/tracking/index.js
@@ -268,7 +268,8 @@ export const handleConnectivityChange = async event => {
  * @property {number} [untilTs] - Until when the locations points should be fetched. Default is 0
  * @property {boolean} [force] - Whether or not the upload should be forced
  *
- * @param {Params} - The method parmas
+ * @param {Params} - The method params
+ * @returns {Promise<number>} The number of uploaded tracking data points
  */
 export const startOpenPathUploadAndPipeline = async ({
   untilTs = 0,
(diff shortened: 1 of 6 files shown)


src/app/domain/geolocation/tracking/index.js
```diff
@@ -292,5 +293,5 @@ export const startOpenPathUploadAndPipeline = async ({
     Log('Failed openpath processing: ' + err)
     log.error(err)
   }
-  return
+  return uploadedCount
```
Crash--, line 296: you should update the jsdoc to reflect that change

src/app/domain/geolocation/tracking/tracking.js
```diff
@@ -252,10 +256,16 @@ export const createNewStartPoint = (previousPoint, nextPoint) => {
...
+  let newTime = date.getTime() - timeToSubtract
+
+  // Check the new time is greater than the previous location point, plus 1s for safety.
+  // Here, we want to prevent overlaping points, that might results in overlapping transitions.
+  if (newTime > new Date(previousPoint.timestamp).getTime() + 1000) {
     // Subtract the time from the date
-    date.setTime(date.getTime() - timeToSubtract)
+    date.setTime(newTime)
+  } else {
+    newTime = date.getTime() - 1000
```
Ldoppea, line 267: I don't get this one. Shouldn't it be `+1000` if we want to be sure `newTime` is superior to previous point?

src/app/domain/geolocation/tracking/tracking.js
```diff
@@ -160,10 +160,10 @@ const prepareMotionData = async ({ locations, lastUploadedPoint }) => {
         // Force a stop/start transition when the temporal delta is too high
         // Note forcing the transition won't automatically create a new trip, typically a long-distance
         // flight of 12h+ should behave as one trip, even with this transition.
-        await addStopTransitions(contentToUpload, getTs(previousPoint) + 1, {
+        await addStopTransitions(contentToUpload, getTs(previousPoint) + 0.01, {
```
Ldoppea, line 163: nit: would be great to extract all those `0.01` values into a named constant. I think this would help undestanding what happens here without having to rely on the commit message

src/app/domain/geolocation/tracking/tracking.js
paultranvan: No, let me explain a bit more: let's say you have P1 (the previous point, i.e. the last point of the previous trip) and P3 (the first point of the new trip). Here we try to compute P2, as the tracking is often a bit late to wake-up (especially on iOS) and P3 might not represent to actual start of the trip.
So, temporally speaking P1 < P2 < P3.

Here, `date` is actually the one from P3. We try to compute the a `newTime`, relying the on P3 speed. We check if `newTime` is greater than P1 + 1s. If it is not the case, we give up relying on the speed, and set:  P2  = P3 - 1s.  

That being said, we might have a problem if P1 +1s > P3. But if those points are temporally very close, P2 is not useful anyway (and will be filtered by the server).

src/app/domain/geolocation/tracking/tracking.js
paultranvan: Ah, I see you point and it makes sense. But, in my modest experience, I'm not a big fan of named constants in such situations, because in my opinion, it's quite less readable. 
I prefer to read `stuff + 0.01` rather than `stuff + TRANSITION_TIME_SHIFT_1_MS` typically. 
But it has the benefit of adding some doc on `TRANSITION_TIME_SHIFT_1_MS` so it might be worth it

src/app/domain/geolocation/tracking/index.js
paultranvan: Right, it's done

src/app/domain/geolocation/tracking/tracking.js
paultranvan: I added the const

PR #1193, opened 2024-03-20
fix: avoid path provider authority conflict (VO-570)

acezard:
fix: avoid path provider authority conflict

These two libraries were accessing the default
appname.provider.
It resulted in file-viewer using the configuration of the logger,
the configuration being not compatible, file-viewer was not operable
We decide to fix this issue by using specific
path providers authorities for 
both these libraries, eliminating
the present conflict and avoiding future ones


patches/react-native-file-logger+0.4.1.patch
```diff
@@ -10,7 +36,7 @@ index 7ee491d..978ce4e 100644
  
  enum LogLevel {
      LOG_LEVEL_DEBUG,
-@@ -19,6 +20,17 @@ @interface FileLogger () <MFMailComposeViewControllerDelegate>
+@@ -19,6 +20,17 @@ static const DDLogLevel ddLogLevel = DDLogLevelDebug;
```
acezard, line 39: I guess this is because I didn't run a pod install? No iOS changes are expected or wanted

Ldoppea (approved):

PR #1208, opened 2024-05-24
Add SharedMemory service

zatteo:
```
### 🔧 Tech

* Add SharedMemory service
```

Related to this PR : https://github.com/cozy/cozy-libs/pull/2540

#### Checklist

* [x] Tested on iOS
* [ ] Tested on Android
* [ ] Test coverage

#### Todo

Merge https://github.com/cozy/cozy-libs/pull/2548 and update cozy-intent

### DIFF
--- a/src/app/domain/scanner/services/scanner.ts
+++ b/src/app/domain/scanner/services/scanner.ts
@@ -3,20 +3,29 @@ import DocumentScanner, {
   ResponseType
 } from 'react-native-document-scanner-plugin'
 
+import {
+  storeSharedMemory,
+  removeSharedMemory
+} from '/libs/localStore/sharedMemory'
+
 type Base64 = string
 
 export const scanDocument = async (): Promise<Base64 | undefined> => {
   try {
+    removeSharedMemory('mespapiers', 'scanDocument')
+
     const { scannedImages } = await DocumentScanner.scanDocument({
       responseType: ResponseType.Base64,
       maxNumDocuments: 1
     })
 
-    if (scannedImages) {
-      return scannedImages[scannedImages.length - 1]
-    } else {
-      return undefined
-    }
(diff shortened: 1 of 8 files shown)


acezard (approved):

src/app/domain/scanner/services/scanner.ts
```diff
@@ -3,20 +3,29 @@ import DocumentScanner, {
...
+import {
+  storeSharedMemory,
+  removeSharedMemory
+} from '/libs/localStore/sharedMemory'
+
 type Base64 = string
 
 export const scanDocument = async (): Promise<Base64 | undefined> => {
   try {
+    removeSharedMemory('mespapiers', 'scanDocument')
```
Ldoppea, line 15: Do we really want to clear this before doing the OS scan (that may produce memory flush too)? Would it be better to do this the later we can?

src/libs/localStore/sharedMemory.ts
```diff
@@ -0,0 +1,51 @@
...
+export const removeSharedMemory = (
+  slug: SharedMemorySlug,
+  key: SharedMemoryKey
+): void => {
+  sharedMemory[slug][key] = undefined
+}
+
+export const sharedMemoryLocalMethods = {
+  getSharedMemory: (
+    slug: SharedMemorySlug,
```
Ldoppea, line 39: @acezard do you know if we can add a mechanism to inject the `slug` automatically from the nativeIntent code? This would prevent a cozy-app to inject the wrong slug and steal data from other cozy-app.

acezard: yes but we would need to update every cozy-app if we add this in cozy-intent. or we could also guess the slug from react-native which requires only updating the mobile app

PR #1209, opened 2024-05-28
feat: Implement offline support through CozyPouchLink

Ldoppea:
We want to make the application work when offline.

To make this happens we want to configure a local PouchDB in order to replicated remote CouchDB data and then we want to serve them to the cozy-client instances in both the react-native side and the cozy-apps side (in WebViews)

This PR implement offline support first for cozy-home and mespapiers, other cozy-apps will come later when the new mechanism is proved to work

Related PR:
- cozy/cozy-client#1507
- cozy/cozy-libs#2562
- cozy/cozy-home#2186
- cozy/mespapiers#660
- cozy/cozy-libs#2585

```
### ✨ Features

* Add support for Offline mode (for now only on cozy-home and mespapiers)

### 🐛 Bug Fixes

*

### 🔧 Tech

*
```

___

TODO:

- [x] Upgrade cozy-client after cozy/cozy-client#1507 being merged
- [x] Upgrade cozy-intent after cozy/cozy-libs#2562 being merged


src/pouchdb/platformReactNative.appState.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const stopListeningAppState = (): void => {
+  appStateHandler?.remove()
+}
+
+const isGoingToSleep = (nextAppState: AppStateStatus): boolean =>
+  Boolean(appState.match(/active/) && nextAppState === 'background')
+
+const isGoingToWakeUp = (nextAppState: AppStateStatus): boolean =>
+  Boolean(appState.match(/background/) && nextAppState === 'active')
```
zatteo, line 34: Is there a reason to use a regex here ?

src/libs/client.js
```diff
@@ -123,14 +128,28 @@ export const fetchCozyDataForSlug = async (slug, client, cookie) => {
...
+    if (err.message === 'Network request failed') {
+      const cachedResult = await getClientCachedData(client, cacheKey)
+
+      if (cachedResult) {
+        return cachedResult
+      }
+    }
+
+    throw err
+  }
```
zatteo, line 167: So this cache has an unlimited lifetime ?

src/pouchdb/sendDbByEmail.ts
```diff
@@ -0,0 +1,146 @@
...
+    for (const dbFile of dbFiles) {
+      const dirs = RNFetchBlob.fs.dirs
+
+      const internalPath = dbFile.path
+
+      if (Platform.OS === 'android') {
+        const date = Number(new Date())
+        const externalPath = `${dirs.DCIMDir}/DbFile_${dbFile.name}${date}.sqlite`
+
+        await RNFS.copyFile(internalPath, externalPath)
```
zatteo, line 55: It does not work with the external path ?

src/pouchdb/sendDbByEmail.ts
```diff
@@ -0,0 +1,146 @@
...
+  const deviceModel = DeviceInfo.getModel()
+  const os = DeviceInfo.getSystemName()
+  const version = DeviceInfo.getSystemVersion()
+
+  const appInfo = `App info: ${appVersion} (${appBuild})`
+  const bundleInfo = `App bundle: ${bundle}`
+  const deviceInfo = `Device info: ${deviceBrand} ${deviceModel} ${os} ${version}`
+
+  return `${appInfo}\n${bundleInfo}\n${deviceInfo}`
+}
```
zatteo, line 137: Same method as [src/app/domain/logger/fileLogger.ts](https://github.com/cozy/cozy-flagship-app/blob/be4e45832f7a83b13ef46d1686e735b0ec9a001b/src/app/domain/logger/fileLogger.ts#L101)

src/components/webviews/jsInteractions/jsCozyInjection.ts
```diff
@@ -17,6 +17,7 @@ const makeMetadata = (routeName?: string): string => {
   return JSON.stringify({
     immersive: routeName ? immersiveRoutes.includes(routeName) : false,
     navbarHeight,
+    offline_available: true,
```
zatteo, line 20: Why is it the best way to announce that offline is available ?

zatteo:
Maybe to do in another PR : replace the `atob` and `btoa` polyfill we have here https://github.com/cozy/cozy-flagship-app/blob/be4e45832f7a83b13ef46d1686e735b0ec9a001b/src/App.js#L70 by `[react-native-quick-base64](https://github.com/craftzdog/react-native-quick-base64)`

src/pouchdb/platformReactNative.storage.ts
```diff
@@ -0,0 +1,14 @@
...
+  getItem: async (key: string): Promise<string | null> => {
+    return AsyncStorage.getItem(key)
+  },
+  setItem: async (key: string, value: string | undefined): Promise<void> => {
+    if (value === undefined) return
+    return AsyncStorage.setItem(key, value)
+  },
+  removeItem: async (key: string): Promise<void> => {
+    return AsyncStorage.removeItem(key)
+  }
```
zatteo, line 13: A prefix is added to the key by PouchDB ?

zatteo:
No real changes requested, just a lot of question.

As always, very easy to read 👍

src/libs/client.js
Ldoppea: For now, yes it is. I imagine that opening the app after a very long time AND while being offline is a rare event so this should be acceptable for now. But for sure we may want to add an invalidation mechanism soon.

src/pouchdb/sendDbByEmail.ts
Ldoppea: do you mean "internal" path?

What I understand is that the Mail app cannot access the FlagshipApp's internal path to load attachments. So I have to copy them in the external path

Note that this behavior is different in the Android emulator and I didn't find a way to make this work in that env (but this is not problematic for now)

src/pouchdb/sendDbByEmail.ts
Ldoppea: We can yes, for now it is not really important, those are debug methods so it won't be problematic if they differ someday

src/components/webviews/jsInteractions/jsCozyInjection.ts
Ldoppea: Not sure if this is the best way, but it is the easiest way I found for now. This is because we should check for offline compatibility during the cozy-app initialization, and in some cozy-apps we don't have access to cozy-intent yet during this phase. So injecting it here is easier to implement than using cozy-intent.

src/pouchdb/platformReactNative.storage.ts
Ldoppea: No, the keys will be:
- `cozy-client-pouch-link-lastreplicateddocid`
- `cozy-client-pouch-link-lastreplicationsequence`
- `cozy-client-pouch-link-synced`

PR #1213, opened 2024-06-05
Sharing files to the app was not working properly on mabulle app on iOS

zatteo:
```
### 🐛 Bug Fixes

* Sharing files to the app was not working properly on mabulle app on iOS
```

#### Checklist

Before merging this PR, the following things must have been done if relevant

* [x] Tested on iOS
* [ ] Tested on Android
* [ ] Test coverage


Ldoppea (approved):
It is a bit difficult to predict how the regex based script will behave so I have one question: did you check that calling multiple times the configure-brand script and switching between cozy and mabulle option will always produce the same result?

If yes everything is ok for me 👍

zatteo:
Tested and validated 👍

PR #1237, opened 2024-08-26
feat: Allow cozy-apps to access files offline

Ldoppea:
When the application is offline, we want the user to be able to access their files that are accessible offline

To make this possible, we want the Flagship app to provide a new `downloadFileAndPreview()` interface that can be called from cozy-apps

Internally this method should allow to download and preview files, but also to store them in the device's storage in order to make them accessible later even when the app is offline

This method will be called from cozy-client through `file` model

For now we make files accessible only if they have already downloaded
once
___
We also want the user's important files to always be accessible offline

To make this possible, we want the app to download them in background on startup

We consider important files as files from `mespapiers` with specific qualification labels like passport, identification papers, vehicle registration, etc.
___
Finally, as we store offline accessible files in the device's storage, we want to prevent them to fill the entire device's storage

We chose to keep only files that were accessed in the previous month. All other files would be deleted

Important files are never deleted as long as they exist on the Cozy instance

___

TODO:

- [x] Upgrade cozy-intent after cozy/cozy-libs#2562 being merged

### DIFF
--- a/src/app/domain/io.cozy.files/importantFiles.ts
+++ b/src/app/domain/io.cozy.files/importantFiles.ts
@@ -1,0 +1,44 @@
+import CozyClient from 'cozy-client'
+import Minilog from 'cozy-minilog'
+
+import { downloadFile } from '/app/domain/io.cozy.files/offlineFiles'
+import {
+  getOfflineFilesConfiguration,
+  removeOfflineFileFromConfiguration
+} from '/app/domain/io.cozy.files/offlineFilesConfiguration'
+import { getImportantFiles } from '/app/domain/io.cozy.files/remoteFiles'
+import { getErrorMessage } from '/libs/functions/getErrorMessage'
+
+const log = Minilog('📁 Offline Files')
+
+const IMPORTANT_FILES_DOWNLOAD_DELAY_IN_MS = 100
+
+export const makeImportantFilesAvailableOfflineInBackground = (
+  client: CozyClient
+): Promise<void> => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      makeImportantFilesAvailableOffline(client)
+        .then(resolve)
+        .catch(error => {
+          const errorMessage = getErrorMessage(error)
+          log.error(
+            `Something went wrong while making important files available offline: ${errorMessage}`,
+            resolve()
+          )
+        })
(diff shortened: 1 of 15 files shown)


src/app/domain/io.cozy.files/offlineFiles.ts
```diff
@@ -0,0 +1,115 @@
...
+    const result = await RNFS.downloadFile({
+      fromUrl: downloadURL,
+      toFile: `${filesFolder}/${filename}`,
+      begin: () => undefined,
+      progress: res => {
+        const progressPercent = res.bytesWritten / res.contentLength
+        setDownloadProgress?.(progressPercent)
+      },
+      progressInterval: 100
+    }).promise
```
zatteo, line 73: Maybe extract the download logic like we did for the upload ? https://github.com/cozy/cozy-flagship-app/tree/master/src/app/domain/upload/services

src/app/domain/io.cozy.files/offlineFilesConfiguration.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const getOfflineFileFromConfiguration = async (
+  fileId: string
+): Promise<OfflineFile | undefined> => {
+  const files = await getOfflineFilesConfiguration()
+
+  if (files.has(fileId)) {
+    return files.get(fileId)
+  }
```
zatteo, line 58: Do we need the `.has()` ? `.get()` returns undefined if it does not exists.

src/app/domain/io.cozy.files/importantFiles.ts
```diff
@@ -37,8 +42,47 @@ const makeImportantFilesAvailableOffline = async (
...
+      const lastOpened = offlineFile.lastOpened
+      if (
+        !importantFilesIds.includes(offlineFile.id) &&
+        (differenceInMonths(lastOpened, now) > NB_OF_MONTH_BEFORE_EXPIRATION ||
+          !lastOpened)
+      ) {
+        log.debug(
+          `Remove old unimportant file ${
+            offlineFile.id
+          } (last opened on ${lastOpened.toString()})`
```
zatteo, line 74: Is there a risk to have an error here if lastOpened is undefined ?

PR #1243, opened 2024-09-13
feat: Prevent local PouchDB modification to be replicated to remote DB

Ldoppea:
For now we don't want to replicate local changes into remote PouchDB as we first want to ensure everything works as expected locally before activating two-way replication in the future


zatteo (approved):

Ldoppea:
The commit has been cherry picked in #1209

PR #1244, opened 2024-09-13
feat: Open cozy-app when offline only if it supports Offline mode

Ldoppea:
When offline, we want to open coz-apps only when they declare to be compatible with Offline mode

Cozy-app can declare this by setting `offline_support: true` in their manifest.webapp file

Non compatible cozy-app won't be opened when the Flagship app is offline, and instead an error message will be displayed

### DIFF
--- a/src/libs/httpserver/httpServerProvider.tsx
+++ b/src/libs/httpserver/httpServerProvider.tsx
@@ -10,6 +10,7 @@ import Config from 'react-native-config'
 import type CozyClient from 'cozy-client'
 import Minilog from 'cozy-minilog'
 
+import { isOfflineCompatible } from '/app/domain/offline/isOfflineCompatible'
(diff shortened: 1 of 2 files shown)


src/libs/httpserver/httpServerProvider.tsx
```diff
@@ -140,12 +141,23 @@ export const HttpServerProvider = (
       )
 
       if (source === 'cache') {
-        if (slug !== 'home' && slug !== 'mespapiers') {
+        log.debug(
+          'App from cache detected, cheking if the app is compatible with offline mode'
+        )
+        const isOffflineCompatitble = await isOfflineCompatible(fqdn, slug)
+
+        if (!isOffflineCompatitble && slug !== 'home') {
```
zatteo, line 149: Why the special case for _home_ ?

PR #1254, opened 2024-10-11
feat: Allow the user to choose to save debug traces on error (SCR-872)

doubleface:
This feature is still behind the `clisk.automatic.html-on-error` flag
because it still needs some UI work.

When an unexpected error is raised from a clisk konnector, the current
page in the worker is displayed to user and a button to send traces is
proposed which does exactly the same as the `clisk.html-on-error` flag.

```
### ✨ Features

* Offers the possibility to save html traces on clisk konnector error
```

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android
* [ ] Test coverage

### DIFF
--- a/src/constants/strings.json
+++ b/src/constants/strings.json
@@ -17,7 +17,9 @@
   "authLogin": "/auth/login?redirect=",
   "konnectors": {
     "worker": {
-      "backButton": "RETOUR À MON COZY"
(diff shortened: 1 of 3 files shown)


KillianCourvoisier (approved):
Thank you, it might be very useful !

PR #1260, opened 2024-11-12
feat: Allow clisk konnectors to run server job (SCR-845)

doubleface:
Clisk konnectors can now run themselves as server jobs. The result of    the execution is given back to the clisk konnector.
    
This is just a first step. Later, it will be possible to let hybrid
konnector be run only in server mode when the context is relevant.

```
### ✨ Features

* Allow clisk konnectors to save data in their own account
* Allow clisk konnectors to run server job
* Allow clisk konnectors to have incognito mode
```

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android
* [ ] Test coverage

### DIFF
--- a/src/libs/Launcher.js
+++ b/src/libs/Launcher.js
@@ -472,6 +472,32 @@ export default class Launcher {
   }
 
   /**
+   * Save data in the current account's data attribute
+   *
+   * @param {Object} data - any object serializable with JSON.stringify
+   * @returns <Promise<import('cozy-client/types/types').IOCozyAccount>>
+   */
+  async saveAccountData(data) {
+    const { launcherClient: client, account } = this.getStartContext() || {}
+
+    if (!account._id) {
+      throw new Error('Launcher: No associated account. Cannot save account data yet')
+    }
+
+    const { data: currentAccount } = await client.query(
+      Q('io.cozy.accounts').getById(account._id)
+    )
+    currentAccount.data = data
+    const { data: newAccount } = await client.save(currentAccount)
+
+    this.setStartContext({
+      ...this.getStartContext(),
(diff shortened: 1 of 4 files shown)


src/libs/ReactNativeLauncher.js
```diff
@@ -583,6 +584,15 @@ class ReactNativeLauncher extends Launcher {
     this.emit('SET_USER_AGENT', userAgent)
   }
 
+  /**
+   * Set incognito mode for worker webview
+   *
+   * @param {Boolean} incognito
+   */
+  async setIncognito(incognito) {
+    this.emit('SET_INCoGNITO', incognito)
```
paultranvan, line 593: I assume the little `o` is a typo?

src/libs/Launcher.js
```diff
@@ -471,6 +471,30 @@ export default class Launcher {
...
+    const jobArgs = {
+      ...args,
+      konnector: konnector.slug,
+      account: account._id,
+      folder_to_save: trigger.message?.folder_to_save
+    }
+    const { data: job } = await client
+      .collection('io.cozy.jobs')
+      .create('konnector', jobArgs, options, true)
+    return await client.collection('io.cozy.jobs').waitFor(job.id)
```
paultranvan, line 495: It implies we need a way to target the server version of the connector, right?

doubleface: Yes, and since the targeted file is not the sames for clisk konnectors (main.js) and server konnectors (index.js), it works. The next goal is to be able define a more explicit names in the manifest of the konnectors.

paultranvan (approved):
CI should be fixed, but looks good otherwise!

PR #1264, opened 2024-12-19
feat: Add performance measurements

Ldoppea:
This PR adds tooling for measuring app's performances.

It is based on [recent CozyClient implementation of Performance API](https://github.com/cozy/cozy-client/pull/1574) and on [react-native-performances plugin](https://github.com/oblador/react-native-performance)

This has been done to mitigate missing RN native Performance APIs that implements `performances.now()` but not `preformances.mark()` nor `preformances.measure()`

The result is that we can now analyze performance through Flipper's plugin `flipper-plugin-performance`

image

Or with a bit of tooling (see commit named `feat: Add conversion script to allow opening logs in Chrome devtools`) it is possible to import those metrics into Google Chrome's devtools

image

TODO:

- [x] Add documentation
- [x] Upgrade cozy-client after cozy/cozy-client#1574 is merged

```
### ✨ Features

*

### 🐛 Bug Fixes

*

### 🔧 Tech

* Add tooling to measure app's performances
```


src/App.js
```diff
@@ -67,8 +71,16 @@ import { useDimensions } from '/libs/dimensions'
...
 import { useColorScheme } from '/app/theme/colorScheme'
 
+if (__DEV__) {
+  require('react-native-performance-flipper-reporter').setupDefaultFlipperReporter()
+}
+
+configurePerformances()
 configureFileLogger()
 
+const markStartName = rnperformance.mark(markStartName)
```
zatteo, line 82: rnperformance.mark arg should be a string ?

Ldoppea: Nice catch, it was a refactoring error. Fixed [here](https://github.com/cozy/cozy-flagship-app/compare/9f81a021acfaa8a142ca2126adf3d84f0cea775e..e76e118d3a4da722258e471225978592b0390bba)

src/App.js
```diff
@@ -181,23 +220,34 @@ const InnerNav = ({ client, setClient }) => {
...
-  </NavigationContainer>
-)
+const Nav = ({ client, setClient }) => {
+  const [markNameNav] = useState(() => rnperformance.mark('Nav'))
+  useEffect(() => {
+    rnperformance.measure({
+      markName: markNameNav,
+      measureName: 'Mount <Nav />'
+    })
+  }, [markNameNav])
```
zatteo, line 230: What do you think of an extracted hook like `usePerformanceMeasure(markName, mesureName)` ?

zatteo (approved):

src/App.js
Ldoppea: Not sure how to implement it right now, so I propose to implement it in the future if we need it

Ldoppea:
I added the documentation in `docs/how-to-debug-performances.md`

PR #1266, opened 2024-12-19
feat: Optimize startup timings by delaying heavy tasts

Ldoppea:
This PR's goal is to improve startup timings by delaying heavy tasks after the startup is complete.

What have been delayed:
- PouchDB replication
- Important files downloading (for offline access)
- Call to Oauth client's `synchronize()` api
- Downloading of cozy-app icons for opening animations

For now those delay have been hardcoded, but in the future we may want to implement some "orchestrator" mechanism that would be responsible to call those tasks when the application is IDLE

```
### ✨ Features

* Improve startup timings

### 🐛 Bug Fixes

*

### 🔧 Tech

*
```


zatteo (approved):
Nice optimization!

PR #1268, opened 2024-12-19
feat: Upgrade pouchdb-adapter-react-native-sqlite to V4 and benefits from OPSqlite

Ldoppea:
This upgrade is made possible due to recent RN73 upgrade and will allow to use `op-sqlite` instead of `react-native-quick-sqlite` so we expect performances improvement

```
### ✨ Features

* Improve app's performances

### 🐛 Bug Fixes

*

### 🔧 Tech

* Migrate PouchDB adapter from `react-native-quick-sqlite` to `op-sqlite`
```


src/pouchdb/shim.js
```diff
@@ -1,6 +1,3 @@
-import { shim } from 'react-native-quick-base64'
+import { install } from 'react-native-quick-crypto'
 
-shim()
-
-// Avoid using node dependent modules
-process.browser = true
+install()
```
zatteo, line 3: :+1: :+1:

android/build.gradle
```diff
@@ -3,7 +3,7 @@
 buildscript {
     ext {
         buildToolsVersion = "34.0.0"
-        minSdkVersion = 21
+        minSdkVersion = 23
```
zatteo, line 6: We'll need to inform the team about this update (which is relevant)

zatteo (approved):

PR #1278, opened 2025-01-13
feat: Trigger replication on appstate changes and realtime events

Ldoppea:
With previous PouchDB/offline implementation, the app was able to trigger a full replication on application's startup, but then nothing would be replicating until next app's restart

This PR is responsible to implement 2 new scenario:
- When the app goes to background and then come back to foreground, then we trigger a replication
- When the app receives a realtime event from one of the CozyPouchLink's configured doctypes, then we trigger a replication

For both scenario, the replication is debounced for 30s with a maximum of 10min

### DIFF
--- a/src/app/domain/offline/hooks/useOfflineReplicationOnRealtime.ts
+++ b/src/app/domain/offline/hooks/useOfflineReplicationOnRealtime.ts
@@ -1,0 +1,54 @@
+import { useEffect } from 'react'
+
+import { useClient } from 'cozy-client'
+import type { CozyClientDocument } from 'cozy-client/types/types'
+import Minilog from 'cozy-minilog'
+
+import { triggerPouchReplication } from '/app/domain/offline/utils'
+import { offlineDoctypes } from '/pouchdb/getLinks'
+
+const log = Minilog('📶 useOfflineReplicationOnRealtime')
+
+export const useOfflineReplicationOnRealtime = (): void => {
+  const client = useClient()
+
+  useEffect(() => {
+    if (!client) return
+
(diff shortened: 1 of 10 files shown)


zatteo (approved):

src/app/domain/offline/hooks/useOfflineReplicationOnRealtime.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+
+    const triggerReplicationCreated = triggerReplication('created')
+    const triggerReplicationUpdated = triggerReplication('updated')
+    const triggerReplicationDeleted = triggerReplication('deleted')
+
+    offlineDoctypes.forEach(doctype => {
+      realtime.subscribe('created', doctype, triggerReplicationCreated)
+      realtime.subscribe('updated', doctype, triggerReplicationUpdated)
+      realtime.subscribe('deleted', doctype, triggerReplicationDeleted)
```
paultranvan, line 36: I realize we trigger a full replication on all doctypes for any realtime change, while we probably should only replicate for the concerned doctype, to save resources. But this is an improvement that can be made for later

paultranvan (approved):
Looks good :+1:

src/app/domain/offline/hooks/useOfflineReplicationOnRealtime.ts
Ldoppea: Yes, we will have to implement something to handle that scenario. For now the replication method always replicate all pouches and no mechanism has been implemented to debounce replications separately.

PR #1291, opened 2025-02-10
POC: Reduce loading of app due to crypto feature

zatteo:
- Use react-native-quick-crypto when possible
- Delay CryptoWebview loading (getClient and useAppBootstrap can start before CryptoWebview loading)

**Todo** 

- [ ] check if it is the delay is not too long
- [ ] check on iOS
- [ ] check on Android


zatteo:
> Since we open signup now to login / register, we can remove this cryptowebview, right?

Interesting, at least we do not need to open it at startup. It still used after but it can helps the loading time 👍

Crash--:
Why do we still need it? I can't remember

zatteo:
> Oh, for Clisk connector maybe?

All usage of Crypto WebView
- the onboarding process (see [this commit](https://github.com/cozy/cozy-flagship-app/pull/88/commits/acadb4179a0097ae78a81b1a7691ee7d29c4a886)) => this may not be necessary anymore as you said 👍
- initialisation of the local httpServer (to limit access to routes by cozy-apps, see [this commit](https://github.com/cozy/cozy-flagship-app/pull/170/commits/0484f72b113abf4dd878c8d1387a45d3fd4eac5e))
- to allow apps served over HTTP (non-secure) to access cryptography methods (see [this PR](https://github.com/cozy/cozy-flagship-app/pull/250))
    - on cozy-settings to enable or disable 2FA
    - on all apps that use cozy-keys-lib (= Harvest)

Crash--:
> on cozy-settings to enable or disable 2FA

We don't allow that anymore for new / migrated account. 

> initialisation of the local httpServer (to limit access to routes by cozy-apps, see [this commit](https://github.com/cozy/cozy-flagship-app/pull/170/commits/0484f72b113abf4dd878c8d1387a45d3fd4eac5e))

We can do without this kind of crypto. 


And once we'll be able to load html file from http://localhost/file.html instead of getting the htmlContent and inject it (and with the dataproxy, we should be able to achieve this, we should take one or two hours together to test)

PR #1298, opened 2025-03-13
feat: Record clisk konnectors execution in io.cozy.clisk.records

doubleface:
This is based on [rrweb](https://www.rrweb.io/)
 
   - only with the flag `clisk.record` enabled
    - each record have a unique sessionId
    - job id in the records when available
    - records saved by batches of 50 and debounced by 1 second, to avoid excessive memory usage

Needs https://github.com/konnectors/libs/pull/1022 in the konnector

The content is a suit of event object in io.cozy.clisk.records doctype. A new implementation in conrad will be needed
to view this content

```
### ✨ Features

* Record clisk konnectors execution in io.cozy.clisk.records

```

#### Checklist

* [ ] Tested on iOS
* [X] Tested on Android
* [X] Test coverage

### DIFF
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -309,7 +309,10 @@ class ReactNativeLauncher extends Launcher {
         throw initKonnectorError
       }
       await this.pilot.call('setContentScriptType', 'pilot')
-      await this.worker.call('setContentScriptType', 'worker')
+      const workerOptions = {
+        shouldRecord: flag('clisk.record')
+      }
+      await this.worker.call('setContentScriptType', 'worker', workerOptions)
       const shouldLogout = await this.cleanCredentialsAccounts(konnector.slug)
       if (shouldLogout) {
         log(
@@ -689,7 +692,10 @@ class ReactNativeLauncher extends Launcher {
         listenedEventsNames: this.workerListenedEventsNames,
         label: 'launcher => worker'
       })
-      await this.worker.call('setContentScriptType', 'worker')
+      const workerOptions = {
+        shouldRecord: flag('clisk.record')
+      }
+      await this.worker.call('setContentScriptType', 'worker', workerOptions)
     } catch (err) {
       throw new Error(`worker bridge restart init error: ${err.message}`)
     }
--- a/src/libs/ReactNativeLauncher.js
+++ b/src/libs/ReactNativeLauncher.js
@@ -25,6 +25,10 @@ import {
 } from '/app/domain/sleep/services/sleep'
 import { hasPermission } from '/app/domain/manifest/permissions'
 
+import { v4 as uuidv4 } from 'uuid'
+
+import CliskRecorder from '/screens/konnectors/core/record'
+
 const log = Minilog('ReactNativeLauncher')
 
 const SET_WORKER_STATE_TIMEOUT_MS = 30 * 1000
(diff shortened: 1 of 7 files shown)


src/libs/ReactNativeLauncher.js
```diff
@@ -55,11 +56,13 @@ class ReactNativeLauncher extends Launcher {
...
       'getCookiesByDomain',
       'getCookieByDomainAndName',
-      'getCookieFromKeychainByName'
+      'getCookieFromKeychainByName',
+      'saveCookieToKeychain'
     ]
     this.workerListenedEventsNames = ['log', 'workerEvent']
 
     this.controller = new AbortController()
+    this.sessionId = Date.now()
```
zatteo, line 65: Is it enough to be unique ?

src/libs/ReactNativeLauncher.js
```diff
@@ -55,11 +56,13 @@ class ReactNativeLauncher extends Launcher {
       'sendToPilot',
       'getCookiesByDomain',
       'getCookieByDomainAndName',
-      'getCookieFromKeychainByName'
+      'getCookieFromKeychainByName',
+      'saveCookieToKeychain'
```
zatteo, line 60: How is it related to this PR?

src/screens/konnectors/core/record.js
```diff
@@ -0,0 +1,108 @@
...
+
+      const dataPayload = { ...JSON.parse(rawData), konnectorSlug, jobId }
+
+      if (!dataPayload.data || dataPayload.messageType !== 'rrweb') return
+
+      delete dataPayload.messageType
+
+      await this.handleRecorderEvent(dataPayload)
+    } catch (e) {
+      console.error('clisk recorder error', e) // eslint-disable-line no-console
```
zatteo, line 103: Why not a real logger here?

zatteo (approved):
Looks good, I have not a lot of background knowledge of clisk operation so it's worth what it's worth

src/libs/ReactNativeLauncher.js
doubleface: More or less, since the user can't launch two clisk connectors at the same time. In reality, he can do it with several phones, but the only risk is to have two recordings mixed up for one user.

src/libs/ReactNativeLauncher.js
doubleface: Ho, this is from a another test. I will remove it, thanks

src/libs/ReactNativeLauncher.js
```diff
@@ -309,7 +329,10 @@ class ReactNativeLauncher extends Launcher {
         throw initKonnectorError
       }
       await this.pilot.call('setContentScriptType', 'pilot')
-      await this.worker.call('setContentScriptType', 'worker')
+      const workerOptions = {
+        shouldRecord: flag('clisk.record')
```
Ldoppea, line 333: `shouldRecord` is defined but never used, is this handled inside of clisk library?

src/libs/ReactNativeLauncher.js
```diff
@@ -121,6 +125,21 @@ class ReactNativeLauncher extends Launcher {
     if (context.job) {
       jobId = context.job.id
     }
+
+    this.cliskRecorder.handleRecorderEvent({
+      type: 5,
```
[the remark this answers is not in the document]
doubleface: This is a rrweb custom event type, Thanks, I added a constant for it which will be more explicit

src/screens/konnectors/core/record.js
doubleface: Right, no I use the launcher logger

src/libs/ReactNativeLauncher.js
doubleface: Actually, this parameter is passed to the konnector itself which will handle the recording. 
This recording is automatically handled by the cozy-clisk lib, imported by all clisk konnectors

PR #1308, opened 2025-05-20
fix: Clear all non-device persisted data on logout

Ldoppea:
In previous implementation we did not clear all the non-device persisted data on Logout

Some of them were created by cozy-pouch-link and were not listed here so they were ignored by the clear process (and so the pouch replication process would be broken after switch cozy's instances)

To prevent this in the future, we change the algorithm to clear ALL the persisted data except the device ones


src/libs/localStore/storage.ts
```diff
@@ -100,11 +100,16 @@ export const getData = async <T>(name: StorageKey): Promise<T | null> => {
 
 export const clearCozyData = async (): Promise<void> => {
   try {
-    const keys = Object.values(CozyPersistedStorageKeys)
+    const keys = storage.getAllKeys()
+    const keysToKeep = Object.values(
+      DevicePersistedStorageKeys
+    ) as unknown as string
```
zatteo, line 106: I think we will also remove UserPersistedStorageKeys here no? Which we should not.

PR #1311, opened 2025-05-20
feat: Allow to use Papillon while offline

Ldoppea:
This PR declares the Papillon related doctypes needed for offline replication and allows Papillon app to use local HTTP server on iOS

### DIFF
--- a/src/pouchdb/getLinks.ts
+++ b/src/pouchdb/getLinks.ts
@@ -31,6 +31,11 @@ export const offlineDoctypes = [
   'io.cozy.apps.suggestions',
   'io.cozy.triggers',
   'io.cozy.apps_registry',
+  'io.cozy.calendar.events',
+  'io.cozy.calendar.todos',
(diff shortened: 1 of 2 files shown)


src/pouchdb/getLinks.ts
```diff
@@ -31,6 +31,11 @@ export const offlineDoctypes = [
   'io.cozy.apps.suggestions',
   'io.cozy.triggers',
   'io.cozy.apps_registry',
+  'io.cozy.calendar.events',
+  'io.cozy.calendar.todos',
+  'io.cozy.calendar.presence',
+  'io.cozy.timeseries.grades',
+  'io.cozy.identities',
```
paultranvan, line 38: It's ok for now, but eventually, we probably should think to a better way to declare doctypes, it is a bit cumbersome for now

paultranvan (approved):

zatteo (approved):

PR #1346, opened 2026-09-16
Avoid unusable app when no home app installed

zatteo:


src/libs/httpserver/httpServerProvider.tsx
```diff
@@ -235,7 +235,7 @@ export const HttpServerProvider = (
       })
       return {
         html: false,
-        source: 'offline'
+        source: 'none'
```
Crash--, line 238: i do not understand the consequence of this choice. Why do you say "offline is disabled"? We don't use it?

src/screens/home/components/HomeView.js
```diff
@@ -79,6 +82,38 @@ const HomeView = ({ route, navigation }) => {
...
+        )
+
+        if (slug === 'home') return
+
+        navigateToApp({ navigation, href: defaultRedirectionUrl, slug })
+      }
+
+      void redirectToDefaultRedirection()
+    }, [client, isHomeUnavailable, navigation])
+  )
```
Crash--, line 115: So we have 2 webview with 2 drive instance right? 

So if we have 10 shared drive, then we'll have 20 websocket?

src/libs/httpserver/httpServerProvider.tsx
zatteo: Yes offline is not enabled in flagship app for now. This was kind of a miss when I disabled it.

src/screens/home/components/HomeView.js
zatteo: `navigateToApp` goes to the 2nd webview dedicated to app and open drive inside so no 2 drive webview.

